PR #83, opened 2012-10-03
First pass of lodash function declarations (no aliases)

junosuarez:
I finished my first pass of the method signatures on the documentation page. There are currently no aliases, and I don't have the chained or curried versions (these are the static signatures only). Also, I think the function signature for `template` is more complicated than the current documentation reflects - I didn't have time right now to dig into the source to model the actual interface.

This compiles fine on the command line, but I don't have access to Visual Studio right now to test out the  IntelliSense.

Hopefully someone can take it a step further from here - I'll work more on it later.

### DIFF
--- a/lodash.d.ts
+++ b/lodash.d.ts
@@ -5,6 +5,107 @@ interface ArrayLike {
     length: number;
 }
 
+interface lodashProto {
+    // aliases
+    drop(array: Array, n?: number): Array;
+    head(array: Array): any;
+    head(array: Array, n: number): Array;
+    methods(object: Object): string[];
+    tail(array: Array, n?: number): Array;
+    take(array: Array): any;
+    take(array: Array, n: number): Array;
+    unique(array: Array, isSorted?: bool, callback?: (any) => any, thisArg?: any): Array;
+    unique(array: Array, callback?: (any) => any, thisArg?: any): Array;
+    // functions
+    after(n: number, func: Function): Function;
+    bind(func: Function, thisArg?: any, ...args: any[]): Function;
+    bindAll(object: Object, ...methodNames: string[]): Object;
+    chain(value: any): Object;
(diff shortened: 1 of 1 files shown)


bcherny:
@jdalton Any interest in adding typings to Lodash core? As is, the typings on DefinitelyTyped are low quality (lots of `any`s, lots of incorrect types, lots of missing signatures), and using Lodash with them can be frustrating. The underlying issues are it's hard to keep typings in sync with external code, and the DT community isn't as motivated to write high quality typings as Lodash contributors are to writing high quality code.

If I put up a PR for TypeScript typings (or, converting existing code to TypeScript), would you take a look?

jdalton:
Hi @bcherny!

I want to avoid adding types to Lodash because it leads to having to support others like Flow. I rather defer to experts on a given type system.

bcherny:
Thanks for the quick response @jdalton!

How about support both? Write Lodash in TS, and use [flowgen](https://github.com/joarwilk/flowgen) or something similar to compile to Flow types (the two are very similar for Lodash's use case).

Without official support, it feels like both TypeScript and Flow users are doomed to a bad experience.

With official support, the experience can be really great - there's currently no fully safe and complete utility lib for TS, and it only seems logical that Lodash should be it.

Would you be willing to give the experiment a shot, assuming that a good TS -> Flow compiler exists?

TrySound:
@bcherny Well, why not rewrite lodash in flow?

bcherny:
@TrySound You mean, as opposed to TypeScript? I don't mind either one, as long as I can compile typings for the other.

FWIW, here's a first pass at converting a handful of functions to TypeScript. As you can see, the changes are quite minimal, and can be made incrementally: https://github.com/bcherny/lodash/commit/f3244aec31434a5f4326e9408545a5e4a226d684.

PR #752, opened 2014-10-20
Merge getter

rambo-panda:
``` js
var hhh = {}, 
    name;

var  r = {
  get name(){
    return name;
  },
  get r(){
    return this.t;
  },
  t : 1
};

_.merge(hhh, r);
name = "rambo";

console.log(hhh.name); //   undefined  but  we want get the 'rambo'
```

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -2173,6 +2173,10 @@
           if (isSrcArr || typeof result != 'undefined') {
             object[key] = result;
           }
+			// maybe the key is a getter
+			if(typeof Object.getOwnPropertyDescriptor(source, key).get === 'function'){
+				Object.defineProperty(object,key, Object.getOwnPropertyDescriptor(source, key));
+			}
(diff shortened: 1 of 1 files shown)


SergeySuslov:
Hello, jdalton!
I am newbie in JS and stuck with extending objects with getters using lodash. Could you explain why you don't add support for getters and setters? Should I avoid using them by some reason?

jdalton:
>  Could you explain why you don't add support for getters and setters?

It's a burden in implementation, support, & performance that's not suited for general purpose method. It's better suited for a specialized mixin to handle. 

> Should I avoid using them by some reason?

Use them if you like them and your environment supports them. The choices of a general purpose utility lib may not align with your own so no worries.

SergeySuslov:
@jdalton Thank you for the explanation and for the awesome library!

PR #850, opened 2015-01-06
Add `_.isInvalidDate`

gdi2290:
I signed the CLA


ruimarinho:
In my opinion, if this type of check was added, it should work as a negation of `_.isDate`. However, since `_.isDate` only tests if a value is an instance of `Date`, parsing the actual value would be a breaking change. You could use a mixin instead: 

```js
function isValidDate() {
  return !isNaN(Date.parse(value))
}

_.mixin({ 'isValidDate': isValidDate });
```

jdalton:
I want to stay out of the validation space for the time being as this can expand to more than just Date. I'd like to track interest in this though to see if it's something to tack on to our roadmap. Until then you can always use `_.runInContext` + `_.mixin`:

```js
// create a pristine lodash instance
var _ = require('lodash').runInContext();
// mixin validation methods
_.mixin({ isValidDate: function() { ... } });
```

PR #874, opened 2015-01-26
handle the sort order for sortByAll

melck:
I created for a complete mixin lodash 2.4.1 to sort on multiple fields in ascending or descending order. I wanted to propose this feature for the 3.0, so I added with tests

Sorting behaves the same way as _.sortBy (). Reverse () and detects how many times the same values have changed their indexes.

To select the sort order, add a '+' or '-' before the name of the field:
```javascript
_.sortByAll(collection, ['-a', '+b']);
```

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -425,33 +425,36 @@
 
   /**
    * Used by `_.sortByAll` to compare multiple properties of each element
-   * in a collection and stable sort them in ascending order.
+   * in a collection and stable sort them in ascending or descending order.
    *
    * @private
    * @param {Object} object The object to compare to `other`.
    * @param {Object} other The object to compare to `object`.
    * @returns {number} Returns the sort order indicator for `object`.
    */
-  function compareMultipleAscending(object, other) {
+  function compareMultiple(object, other) {
     var index = -1,
         objCriteria = object.criteria,
         othCriteria = other.criteria,
-        length = objCriteria.length;
+        reverseSameKeys = object.reverseSameKeys,
+        reverse = object.reverse,
+        length = objCriteria.length,
+        result;
 
     while (++index < length) {
-      var result = baseCompareAscending(objCriteria[index], othCriteria[index]);
+      result = baseCompareAscending(objCriteria[index], othCriteria[index]);
       if (result) {
-        return result;
+        return reverse[index] ? -result : result;
       }
     }
     // Fixes an `Array#sort` bug in the JS engine embedded in Adobe applications
-    // that causes it, under certain circumstances, to provide the same value for
-    // `object` and `other`. See https://github.com/jashkenas/underscore/pull/1247
-    // for more details.
+    // that causes it, under certain circumstances, to provide the same value
+    // for `object` and `other`. See https://github.com/jashkenas/underscore/pull/1247.
     //
     // This also ensures a stable sort in V8 and other engines.
-    // See https://code.google.com/p/v8/issues/detail?id=90 for more details.
-    return object.index - other.index;
+    // See https://code.google.com/p/v8/issues/detail?id=90.
+    result = object.index - other.index;
+    return reverseSameKeys ? -result : result ;
   }
 
   /**
@@ -6528,7 +6531,7 @@
 
     /**
      * This method is like `_.sortBy` except that it sorts by property names
-     * instead of an iteratee function.
+     * instead of an iteratee function and the sort order can be set
      *
      * @static
      * @memberOf _
@@ -6536,6 +6539,7 @@
      * @param {Array|Object|string} collection The collection to iterate over.
      * @param {...(string|string[])} props The property names to sort by,
      *  specified as individual property names or arrays of property names.
+     *  The order can be specified by adding '+' or '-'.
      * @returns {Array} Returns the new sorted array.
      * @example
      *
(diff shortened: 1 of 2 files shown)


jdalton:
Hi @melck! Thanks for the pull request however I don't like using the `-` or `+` prefix as that limits key names.

melck:
I choose a prefix system but it can be done otherwise. It's above all the ability to quickly sort and avoid to reverse the array that is interesting here.

I can change him if you know what you want.

lodash.src.js
```diff
@@ -6555,9 +6563,26 @@
...
+          reverse = isLength(length) ? Array(length) : [];
+
+      while(length--) {
+          var regex = /^(\+|-|)(.*)?$/gi.exec(props[length]),
+          isReversed = regex[1] === '-' ? true : false,
+          hex = isReversed ? "1" : "0",
+          prop = regex[2];
+        props[length] = prop;
+        reverse[length] = isReversed;
+        reverseHex = hex.concat(reverseHex);
```
jdalton: Can you explain this a bit more.

melck:
It's a fast way to determine if we need to inverse the index of same values for have the same results as "sortBy / reverse". I'm not good in english so i will show you by exemple

```javascript
var datas = [
  {a: 1, b:1, c:1, d:1},
  {a: 2, b:1, c:1, d:2},
  {a: 3, b:1, c:1, d:3},
  {a: 2, b:2, c:2, d:4},
  {a: 1, b:1, c:1, d:5},
  {a: 3, b:2, c:2, d:6},
  {a: 3, b:3, c:2, d:6},
];



/*  
    Possible cases | sort by using sortBy and reverse                                              
    +a+b+c         _(datas).sortBy('c').sortBy('b').sortBy('a').value()
    +a+b-c         _(datas).sortBy('c').reverse().sortBy('b').sortBy('a').value()
    +a-b+c         _(datas).sortBy('c').reverse().sortBy('b').reverse().sortBy('a').value()
    +a-b-c         _(datas).sortBy('c').sortBy('b').reverse().sortBy('a').value()
    -a+b+c         _(datas).sortBy('c').sortBy('b').reverse().sortBy('a').reverse().value()
    -a+b-c         _(datas).sortBy('c').sortBy('b').reverse().sortBy('a').reverse().value()
    -a-b+c         _(datas).sortBy('c').reverse().sortBy('b').sortBy('a').reverse().value()
    -a-b-c         _(datas).sortBy('c').sortBy('b').sortBy('a').reverse().value()
*/

/*  
    Possible cases | Hexa  | decimal | times of reverse | reverse same values
    +a+b+c           000     0         0                  false                             
    +a+b-c           001     1         1                  true                   
    +a-b+c           010     2         2                  false                   
    +a-b-c           011     3         1                  true                   
    -a+b+c           100     4         2                  false                   
    -a+b-c           101     5         3                  true                   
    -a-b+c           110     6         2                  false                   
    -a-b-c           111     7         1                  true               


    Every times as the decimal result is odd, we need to reverse the identical results    
*/
````

**Comment by melck** [comment]
At the beginning, i used logic like the following, but i have discovered a repeated pattern who permitted a more quick sort:
```javascript
var index = -1,
  length = props ? props.length : 0,
  props = Array(typeof length == 'number' ? length : 0),
  reverse = Array(typeof length == 'number' ? length : 0),
  reverseCount = 0;

while (++index < length) {
  var
    isLastReversed = reverse[index - 1],
    isFirst = index === 0,
    regex = /^(\+|-|)(.*)?$/gi.exec(callback[index]),
    isReversed = regex[1] === "-";

  if ((isReversed && isFirst) || (_.isBoolean(isLastReversed) && (isLastReversed != isReversed))) {
    reverseCount++;
  }

  props[index] = regex[2];
  reverse[index] = isReversed;
}

reverseCount %= 2;
```

Exemple for 4 properties to sort (work for x properties):
* First column present cases.
* Second column: binary represention of each property, '1' is for property sort by desc and '0' sort by asc
* Third column: binary represention of each property with '1' when we need to reverse after a classic function sortBy
* Fourth column is decimal representation of second column.
* Fifth column represent how many times we reversed with classic way of sortBy / reverse
* Sixth column represent when same values keys need to be reversed.



|Cases     | Case binary | Reverse binary | Case decimal | Times reversed                      | Revert |
|:--------:|:-----------:|:--------------:|:------------:|:-----------------------------------:|--------------------------:|
|`+a+b+c+d`|`0000`       |`0000`          |`00`          |`00`                                 |NO                         |     
|`+a+b+c-d`|`0001`       |`0001`          |`01`          |`01`                                 |YES                        |      
|`+a+b-c+d`|`0010`       |`0011`          |`02`          |`02`                                 |NO                         |     
|`+a+b-c-d`|`0011`       |`0010`          |`03`          |`01`                                 |YES                        |      
|`+a-b+c+d`|`0100`       |`0110`          |`04`          |`02`                                 |NO                         |     
|`+a-b+c-d`|`0101`       |`0111`          |`05`          |`03`                                 |YES                        |      
|`+a-b-c+d`|`0110`       |`0101`          |`06`          |`02`                                 |NO                         |     
|`+a-b-c-d`|`0111`       |`0100`          |`07`          |`01`                                 |YES                        |      
|`-a+b+c+d`|`1000`       |`1100`          |`08`          |`02`                                 |NO                         |     
|`-a+b+c-d`|`1001`       |`1101`          |`09`          |`03`                                 |YES                        |      
|`-a+b-c+d`|`1010`       |`1111`          |`10`          |`04`                                 |NO                         |     
|`-a+b-c-d`|`1011`       |`1110`          |`11`          |`03`                                 |YES                        |      
|`-a-b+c+d`|`1100`       |`1010`          |`12`          |`02`                                 |NO                         |     
|`-a-b+c-d`|`1101`       |`1011`          |`13`          |`03`                                 |YES                        |      
|`-a-b-c+d`|`1110`       |`1001`          |`14`          |`02`                                 |NO                         |     

I'd really like to see this feature in lodash whatever the system to sort.
Don't hesitate to tell me the system as you would and i will modify my pr.

PR #883, opened 2015-01-27
Fix Wikipedia link for kebabCase.

aeons:


jdalton:
Ah the anchor changed. Good catch. Have you signed [the CLA](http://dojofoundation.org/about/claForm)?

aeons:
No, but it seems a bit much to do that for one word. Feel free to just fix it in your own commit and delete this.

jdalton:
Please read the [CONTRIBUTING](https://github.com/lodash/lodash/blob/master/CONTRIBUTING.md). Though not technically a must for something like this I don't dig  devs that have problems with CLAs. It's a sign once and done thing that takes less time than setting up this pull request and opens up contributions to all other projects in the foundation at once.

aeons:
Sure thing.

And done. Same email as here, so that's all that's needed, right?

jdalton:
Yap :)

PR #887, opened 2015-01-27
`find` conflicts with native browser function

mreinstein:


jdalton:
It's a basic usage example for now so I think it's alright.

mreinstein:
The problem is the most obvious variable name choice will actually break browsers. :/ It would be good to just make this clear that it's going to conflict, hence the leading _.

jdalton:
The use case is for Node.js not browsers. Browserifying and other environments are outside the scope of this basic installation example.

mreinstein:
I understand it's outside of scope, and I don't advocate litering lodash
source/examples with lots of side comments,  but it stands to reason that a
lot of people are going to use the find function in a browser context. It's
the only function in lodash I've encountered that collides with the
browser's namespace. Would it be that bad to simply put a note in there and
give people a heads-up for this particular function?

jdalton:
> Would it be that bad to simply put a note in there and give people a heads-up for this particular function?

I'll revisit on the first legit case someone in the real world runs into the issue, but until then it's not worth the effort of tracking and modifying the build for.

StreetStrider:
@mreinstein, hmm, but how can they use `require` in browser context? This example only runs in CommonJS environments. If it is a browserify then user knows what he's doing (and browserify always creates IIFE-wrappers, so there's no conflict at all).

PR #910, opened 2015-02-01
typo in isNaN doc

strathausen:
just a small glitch in the doc


KenanY:
That's not a typo; the example is demonstrating the difference between native `isNaN` and `_.isNaN`.

PR #913, opened 2015-02-02
add support cyrillic (russian) for _.words

KhodeN:
ЁА-Я:  \u0401\u0410-\u042F
ёа-я: \u0451\u0430-\u044F

```
_.words('приветМир')
["привет", "Мир"]
```

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -127,8 +127,8 @@
 
   /** Used to match words to create compound words. */
   var reWords = (function() {
-    var upper = '[A-Z\\xc0-\\xd6\\xd8-\\xde]',
-        lower = '[a-z\\xdf-\\xf6\\xf8-\\xff]+';
+    var upper = '[A-Z\\xc0-\\xd6\\xd8-\\xde\\u0401\\u0410-\\u042F]',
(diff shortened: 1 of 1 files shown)


d10:
This seems like a simple enough change, but it would be unrealistic to support the many different languages out there. Also, other string functions like like `_.xyzCase` still would not written with these languages (I think).

KhodeN:
It works fine for _.xyzCase:

```
console.log(_.words('приветМир'))
//  ["привет", "Мир"]
console.log(_.capitalize('приветМир'))
// ПриветМир
console.log(_.kebabCase('приветМир'))
// привет_мир
console.log(_.snakeCase('приветМир'))
// привет_мир
console.log(_.camelCase('Привет Мир'))
// приветМир
```
because all of this functions use the `reWords` regexp

KhodeN:
Perhaps you are right, these functions will rarely be called with Cyrillic. 
On the other side, `reWords` supports for `latin-1`, not only English. For what?
I don't see much difficulty if you add support for Cyrillic characters also.

jdalton:
We support a section of `latin-1` because we support deburring accents. For your case you can leverage `_.partialRight` or `_.partial` with placeholders and pass a [custom regexp](https://lodash.com/docs/#words):

```
var myWords = _.partial(_.words, _, myRegExp);
```

PR #933, opened 2015-02-05
add badge :)

PeterDaveHello:


d10:
I think we don't need the badges as they are basically adding bloat. The version number is at the top of the header, while CI status might not be an indicator of the state of the stable release, just the latest edge build.

jdalton:
@PeterDaveHello Yap, @d10 is correct. Thanks though.

PeterDaveHello:
ok

PR #942, opened 2015-02-08
Add _.pascalCase.

ruimarinho:
Pascal Case is often called Upper Camel Case or Bumpy Case.

`foo_bar -> FooBar`

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -9591,6 +9591,30 @@
     }
 
     /**
+     * Converts `string` to pascal case.
+     * See [Wikipedia](https://en.wikipedia.org/wiki/CamelCase) for more details.
+     *
+     * @static
+     * @memberOf _
+     * @category String
+     * @param {string} [string=''] The string to convert.
+     * @returns {string} Returns the pascal cased string.
+     * @example
+     *
+     * _.pascalCase('Foo Bar');
+     * // => 'FooBar'
+     *
+     * _.pascalCase('--foo-bar');
+     * // => 'FooBar'
+     *
+     * _.pascalCase('__foo_bar__');
+     * // => 'FooBar'
+     */
+    function pascalCase(string) {
+      return capitalize(camelCase(string));
+    }
+
+    /**
      * Repeats the given string `n` times.
      *
      * @static
@@ -10765,6 +10789,7 @@
     lodash.padLeft = padLeft;
     lodash.padRight = padRight;
     lodash.parseInt = parseInt;
+    lodash.pascalCase = pascalCase;
     lodash.random = random;
     lodash.reduce = reduce;
     lodash.reduceRight = reduceRight;
--- a/test/test.js
+++ b/test/test.js
@@ -1449,7 +1449,7 @@
 
   QUnit.module('case methods');
 
-  _.each(['camel', 'kebab', 'snake', 'start'], function(caseName) {
+  _.each(['camel', 'kebab', 'pascal', 'snake', 'start'], function(caseName) {
     var methodName = caseName + 'Case',
         func = _[methodName];
 
@@ -1462,6 +1462,7 @@
       switch (caseName) {
         case 'camel': return 'fooBar';
         case 'kebab': return 'foo-bar';
+        case 'pascal': return 'FooBar';
         case 'snake': return 'foo_bar';
         case 'start': return 'Foo Bar';
       }
@@ -1488,7 +1489,7 @@
     test('`_.' + methodName + '` should deburr letters', 1, function() {
       var actual = _.map(burredLetters, function(burred, index) {
         var letter = deburredLetters[index];
-        letter = caseName == 'start' ? _.capitalize(letter) : letter.toLowerCase();
+        letter = caseName == 'start' || caseName == 'pascal' ? _.capitalize(letter) : letter.toLowerCase();
         return func(burred) === letter;
       });
 
@@ -14892,7 +14893,7 @@
 
     var acceptFalsey = _.difference(allMethods, rejectFalsey);
 
-    test('should accept falsey arguments', 206, function() {
+    test('should accept falsey arguments', 207, function() {
       var emptyArrays = _.map(falsey, _.constant([])),
           isExposed = '_' in root,
           oldDash = root._;


jdalton:
I'd like to hold off on this since it lends itself so well to composition.

```js
_.mixin({ 'pascalCase': _.flow(_.camelCase, _.capitalize) });
```

ruimarinho:
I just found it interesting because I was using the `pascal-case` package for this and even though composition blends so nicely here, I thought the shorthand would be useful considering the popularity of the [pascal-case node module](https://www.npmjs.com/package/pascal-case): *234,666* downloads in the last month.

jdalton:
For comparison [lodash.noop](https://www.npmjs.com/package/lodash.noop) is downloaded more in the last month.

ruimarinho:
Of the 4 dependents of `lodash.noop`, one is another lodash dependency (`lodash._setbinddata`) which is probably downloaded on several other cli packages, which explains its popularity. Let's wait for demand for `pascalCase` then.

aearly:
I'd be :+1: for PascalCase.  Also for CONSTANT_CASE, just so all the various flavors of variable casing are covered.  Currently, I'm just relying on composition.

g13013:
:+1:

mateusmedeiros:
I think this feature should be implemented exactly because it lends itself so well to composition. 

Because that means it would be very trivial to implement, yet a nice extra commodity for the user. Take `capitalize`, for example. Implementing it would be (and is, if I remember correctly) a simple matter of chaining `toLowerCase` and `upperFirst`, yet it is there on lodash and I think we all agree it's a perfectly fine method.

lodash.src.js
```diff
@@ -9591,6 +9591,30 @@
...
+     * // => 'FooBar'
+     *
+     * _.pascalCase('--foo-bar');
+     * // => 'FooBar'
+     *
+     * _.pascalCase('__foo_bar__');
+     * // => 'FooBar'
+     */
+    function pascalCase(string) {
+      return capitalize(camelCase(string));
```
localshred: With the latest lodash you want to use `upperFirst` here instead of `capitalize`.

ruimarinho: If this gets added, I will update it :) thanks.

ozum:
+1

patrikx3:
if we have camcelCase, kebabPase, why we cannot have PascelCase? Please just add it in.

slavafomin:
I've just spent a lot of time looking for this function in the documentation. It's frustrating why it's not added to the lodash, especially considering there is an existing implementation for this.

patrikx3:
This is weird why kebabCase and camelCase is there, but PascalCase core features needs vote.
Such a small request.

michaeljota:
Please, do. 

Thanks for the workaround in the middle time.

khawajaumarfarooq:
@jdalton Pascal case is not just camel case + upper first, because in case the first word is a 2 letter acronym, both letter will be capitalized, so for example "I/O Status" will be ioStatus in camel case and IOStatus in Pascal case

Please implement this :)

shaunc:
@khawajaumarfarooq -- how does that rule work? Presumably `to me` -> `ToMe`. Does it depend on the capitalization of the original? So... `XXX YYY` -> `XXxYyy` ? (Why just 2 letter acronyms?) Note that in camel case, `i/o status` -> `iOStatus` so the example you gave works with camel case + upper first.

khawajaumarfarooq:
@shaunc This might be C#/.NET specific, but the rule there is that acronyms that are 3 or more characters go from, for example, `xml` to `Xml`, whereas acronyms like `io` go to `IO` when going from camel case to Pascal case.

Some examples:

- ioPort becomes IOPort
- xmlDocument becomes XmlDocument
- htmlClient becomes HtmlClient

An exception is sometimes made for `id`, which is Pascal-ed to `Id`, although more recently, I have been tending to Pascal it to `ID`, since I like to write something like `userIDs` instead of `userIds` for variables holding enumerable types.

khawajaumarfarooq:
`i/o status` should be `ioStatus` in camel case and `IOStatus` in Pascal case.

falsyvalues:
@khawajaumarfarooq You can't really say what is an acronym without making a list it. The real possible implementation is in @jdalton comment.

michaeljota:
@khawajaumarfarooq That's how you use it, what about how I use it? What about how other peoples uses it? For me, the solution that @jdalton provides is not perfect, but is the best, and I think that even if _.pascalCase gets implemented, it would most probably be like this. `_` has no way to know what a person, or a group call an acronym, so there is no way to follow your rule, unless is like @falsyvalues said, and make yourself a list. And you can do it, and should do. Cause it would be perfect for your usage, and if you need another acronym, then you just has to add it to the list.

khawajaumarfarooq:
There's no doubt that proper implementation of Pascal case will need a list of 2 letter acronyms (because 3+ letter acronyms get treated as a word). There aren't that many 2-letter words in the English language, so a list to figure out if a 2-letter combo is a word, and possibly a way to over-ride the list to cater to customizations. This is probably relevant to camel-case as well because I have never seen `iOStatus` as valid camel-case.

Anyway, if the majority here doesn't feel the need, that's fine.

mmkal:
@jdalton composition isn't really an option for typescript users - can this be reopened?

MgSam:
Is 55 votes not enough to prove the importance of adding this?

jdalton:
Hi @MgSam!

There is no threshold for 👍reactions. They are used as an indicator of interest when we're looking to add new methods is all.

kylemh:
HI @jdalton 👋 

I see the following wording in the CONTRIBUTING document that you linked.

> until sufficient interest, e.g. 👍 reactions, has been shown by the community

I was curious if you are at all interested in specifying the specific threshold required to be considered "sufficient interest".

teawithsatyam:
This works for me though
`
`_.capitalize(_.camelCase('foo bar'))`

bertho-zero:
@teawithsatyam This is not PascalCase, you should use `upperFirst` instead of `capitalize` to conserve `C` of `Case`

lucasrcosta:
For anyone bumping here wanting to do this is Typescript:
```ts
interface LoDashMixins extends _.LoDashStatic {
  pascalCase(string?: string | undefined): string
}

_.mixin({ pascalCase: _.flow(_.camelCase, _.upperFirst) })
const mixedLodash = <LoDashMixins>_

export { mixedLodash as _ }
```

193 👍 so far, seems pretty popular...

tpict:
Or alternatively use module augmentation, so you can import lodash as normal:
```tsx
declare module "lodash" {
  export interface LoDashStatic {
    pascalCase(string?: string): string
  }
}
```

kihonq:
I am concern that the votes we have on this feature is more than the usual Likes I got on my Instagram post 😢

klingat:
This has been working for me as a temporary workaround for lack of pascalCase in lodash:
```js
import camelCase from 'lodash/camelCase'
import startCase from 'lodash/startCase'

const pascalCase = (string) => {
  return startCase(camelCase(string)).replace(/ /g, '')
}
```

fcsonline:
Not available yet?

ballonura:
Not available yet?

TheJaredWilcurt:
I needed this again.

MareStare:
Real bruh moment

Issue #955, opened 2015-02-10
lazy slice after filtering may be broken


qsona:
`_([1, 2, 3]).filter().slice(1, 3).value();` // expected [2, 3] but actual [2]

I digged by using `git bisect` and it seems to be occured since commit 
6dfc2aca4beaabd758404b1191f7d8044c0b6c23

Sorry if I'm wrong. Thanks.

AaronO:
Hi @jdalton, this issue was troubling us for [GitbookIO/gitbook](https://github.com/GitbookIO/gitbook). Any idea when this fix will be shipped as part of a new release ?

(We could use a `git+https:// ...` dependency, but that's less than ideal).

**P.S:** Thanks for the great work on `lodash` :+1:

jdalton:
A bump will happen this evening or in the morning. It will be v3.2.0.

Issue #912, opened 2015-02-02
support _.uniq to lazy chaining


qsona:
Hi, congratulations for the release of lodash v3! I appreciate for your work very much.

I want the support of `_.uniq` for lazy chaining, but it seems to be not in your roadmap. Is it possible to implement it?
```js
var execCount = 0;
var ary = [0,1,2,3,4,5,6,7,8,9];
_.chain(ary).uniq(function(val){execCount++; return val;}).slice(0,2).value();
console.log(execCount); // 10, so it’s not lazily evaluated
```

p.s. We have used lodash 3.0.0-pre since about 4 months ago on our server-side project with Node.js, and we get along well with it. All bugs that I found are fixed at the edge so it works perfectly now.

jdalton:
At the moment we aren't planning it. That said if you have an implementation in mind I wouldn't mind looking at it.

qsona:
Thanks. I'm planning to send PR.

qsona:
@jdalton I implemented it roughly. Would you please check it?
https://github.com/qsona/lodash/commit/b28e2d9ede6df3a9ac53a63e195f64b0fda154b3
It worked as far as I tested. If you are positive to take it and #944 is solved, I'll clean up the codes and add some test codes.

jdalton:
Thanks, I'll try to review it today.

PR #956, opened 2015-02-10
Add lazy `uniq`.

qsona:
Hi, this is the implementation of #912 .
One test failed, but it seems to be a bug of slice ( #955 .)

I'll welcome any comments, including some corrections of my poor English (variable name and comment).
Thank you.

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -3626,6 +3626,62 @@
     }
 
     /**
+     * Gets an iteratee for the lazy version of `_.uniq`.
+     * @private
+     * @returns {Function}
+     */
+    function getLazyBaseUniqIteratee(iteratee, isLarge) {
+      var indexOf = getIndexOf(),
+          isCommon = indexOf == baseIndexOf,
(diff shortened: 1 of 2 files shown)


qsona:
I have signed CLA now.

jdalton:
Thank you for the implementation example. We already have optimizations in place for `_.uniq` and the code required to lazyify it is substantial so I think we'll pass on adding it.

qsona:
@jdalton Thanks for checking it. 
My use case is that execute uniq to array which has 100~150 elements and then get first 6 element, so I thought it would be nice to evaluate lazily. But I also worried about the performance problem on this implementation. Anyway, I appreciate you because it was so nice experience to read your codes and implement it. Thank you.

jdalton:
You're feedback on issues found while working on this has been super valuable too. And now that you've signed the CLA you're good for all future contributions.

qsona:
Thanks, I'll try to continue contributing :)

PR #959, opened 2015-02-11
Minor lazy pluck simplication

megawac:


jdalton:
I want to hold off on this until I decide what to do with the pluck shorthand. At the moment I'm in between overloading `_.property` or making a new method. Internally the functionality will be split into 2 functions so I'm leaning towards a new method. This would also mean a variant for `propertyOf` as well. This then gets into naming. Because this is comparing a single value to a single property it can be used internally in `matches` and `isMatch` too.

jdalton:
I have this coded locally as `propertyMatch` as a placeholder name. The variant would be `propertyOfMatch`. Another one could be `compareProperty` and `propertyCompare` or `propertyEquals`.

jdalton:
By the way the `_.find(array, propName, propValue)` pattern is ~50% faster than the equiv `matches` pattern so that's rad.

jdalton:
Maybe `matchProperty` and `matchPropertyOf`

megawac:
> By the way the _.find(array, propName, propValue) pattern is ~50% faster than the equiv matches pattern so that's rad.

Dependent on object size ofc, but thats sweet, didn't even notice you add the sugar for that!

> I want to hold off on this until I decide what to do with the pluck shorthand. At the moment I'm in between overloading _.property or making a new method.

Yeah I was thinking about this as well since you brought it up. I would say make it a new method

jdalton:
> Dependent on object size ofc, but thats sweet, didn't even notice you add the sugar for that!

I haven't committed it but the case it applies against is `matches` with a single property, `{ 'foo': 'bar' }`.

> Yeah I was thinking about this as well since you brought it up. I would say make it a new method

Halp naming.

megawac:
What are your thoughts on strict vs deep equality in this case (I'm leaning towards strict)

> Maybe `matchProperty` and `matchPropertyOf`?

Seems reasonable, but may be a bit confusing depending on what sort of equality you go here

jdalton:
> Seems reasonable, but may be a bit confusing depending on what sort of equality you go here

Locally I have `indexOf` rules (so strict plus `NaN`) but it should probably be `isEqual` rules.

jdalton:
Ok I think `matchesProperty` is a better fit as it combos `matches` and `property` :P

megawac:
Makes ¢

jdalton:
So this covers the common case gap of inherited properties and gives a perf boost too, rock.

megawac:
Another thing you may want to consider adding is a `_.hasProperty` or something of the sort, similar to `propertyOf` which is just

```js
function hasProperty(prop) {
    return function(object) {
         return _.has(object, prop);
    }
}
```

 Which can be extremely useful for filters and picks

jdalton:
Btw perf of "_.matchesProperty" shorthand and method use.
 * [shorthand](http://jsperf.com/matches-vs-property-match#chart=bar)
 * [method use](http://jsperf.com/matches-vs-property-match/2#chart=bar

There's bigger wins for the method use as the shorthand of `_.matches` is already optimized to avoid cloning.

PR #968, opened 2015-02-15
inRange and inStrictRange

mzdv:
Added support for testing if a number belongs to a strict range or to a loose range.

Also, first real pull request so sorry in advance if something is out of order.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 *.map
 lodash.compat.min.js
 node_modules
+
+.idea
--- a/lodash.js
+++ b/lodash.js
@@ -10404,6 +10404,15 @@
       return baseMatches(baseClone(source, true));
     }
 
+    function inRange(lower, upper, value) {
+      parsed = parseInt(value);
+      return(parsed >= lower && parsed <= upper);
+    }
+
+    function inStrictRange(lower, upper, value) {
+      parsed = parseInt(value);
+      return(parsed > lower && parsed < upper);
+    }
     /**
      * Creates a function which compares the property value of `key` on a given
      * object to `value`.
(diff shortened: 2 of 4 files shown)


jdalton:
Can explain the common use case.

mzdv:
Use case:
Provided that you have `sqr()`, determine how many squared numbers from 1 to 100 belong in a range between 1 and 1000.

Application:
1) Apply the `sqr()` function to each number in the range from 1 to 100
2) Pass the number into `inRange(1, 1000, number)`, where `number` is the squared number
3) Count the number of `true` responses 

This is a pretty synthetic use case, but it's the only one I can think of right now, mostly because I used it this way. Analogous for strict comparisons (not including 1 and 1000) using `inStrictRange`

jdalton:
I'm ok with having an `inRange` method however I'd want it to be pretty simple at first. No `inStringRange` and the `inRange` method would need to mimic [_.range](https://lodash.com/docs#range)'s `start` and `end` rules (from `start` up to, but not including, `end`).

jdalton:
Also the arg signature should be `value, start, end`. If `end` isn't provided then `start` is treated as `end` with the real `start` being `0`.

mzdv:
Okay, no problem. What about providing an optional parameter to `inRange` in case someone wants a strict comparison? Or should I remove it completely?

jdalton:
>  What about providing an optional parameter to `inRange` in case someone wants a strict comparison?

We can tackle that at a later time.

jdalton:
Have you signed [our CLA](http://dojofoundation.org/about/claForm)?

mzdv:
I signed it right now. Making changes as we speak to the code.

mzdv:
This seems odd. I don't know where to look at, since test passed on my machine and I haven't done anything more than refactoring and removal of `inStrictRange`

jdalton:
Don't worry about the false fails. Travis and npm have an unstable relationship (it errors a lot).

mzdv:
Glad to hear that. I did suspect npm and Travis were having some issues, since the error log was pointing at npm.

jdalton:
It's a recurring issue (it goes weeks without issues then hits like every run for a while). I'll pass it along to travis/npm.

Issue #969, opened 2015-02-15
Docs highlighting messed up for _.some


austinpray:
https://lodash.com/docs#some

![screenshot 2015-02-15 13 26 19](https://cloud.githubusercontent.com/assets/2192970/6204584/526a55d2-b516-11e4-8ea1-95d254416b00.png)

PR #970, opened 2015-02-15
Fix error in _.some example

austinpray:
fixes https://github.com/lodash/lodash/issues/969

Let me know if this is the correct way to contribute to the project. I know there are build steps that need to happen.


jdalton:
Hi @austinpray! Have you signed [our CLA](http://dojofoundation.org/about/claForm)?

austinpray:
Yes!

Issue #414, opened 2013-12-01
Sort order with multiple keys


luksch:
Hi, I am a in awe of the power of lodash. For my project it has been of invaluable help so far. However, now I am stuck with a problem with which lodash seems to fall a bit short.

Here is the problem:

I have an array of objects with the same properties, say 

```
[{name:"barney", age: 38},
 {name:"fred", age: 40},
 {name:"fred", age: 12}]
```

I want to sort the array by name in descending order, but age should be in ascending order. 

lodash supports multiple sorting keys by the "pluck" syntax, but only can deal with ascending sorting this way. Is there a way that the sortBy function is altered that it also accepts an array of objects that define the property name (key) and the order direction? I would love to do:

```
mySortedArr = _.sortby(
    unsortedArr, [
        {propertyName:"name", order:"desc"}, 
        {propertyName:"age", order:"asc"}
    ]);
```

Any chance something like that enters the library?

darthsteedious:
+1 - would love a _.sortBy(myArray, ['foo', '-bar']) syntax.

ghost1face:
+1

jdalton:
I'd dig a proposed implementation to go along with the +1's.
In 3.0 we're separating `_.sortBy` out into `_.sortByAll` to handle multiple properties. This lets use keep `_.sortBy` simple.

ItsLeeOwen:
+1

CTAPbIuMABP:
You always can add _.sortBySome

wrummler:
I'd love to see `sortByAll` handle optionally-specified per-criterion descending sort order. (Btw, gotta mention it... Like many others, I am in love with the performance and expressiveness of lodash. Thanks for your awesome work.)

Perhaps a signature something like `sortByAll(collection, props, order)` for usage like `_.sortByAll(myArray, ['foo', 'bar'], [false, true])`, to borrow from @octref 's example?

Copying & adapting `compareMultipleAscending` to `compareMultiple` seems straightforward (to me at least, relatively new to JS and possibly unaware of some pitfalls):

    /**
     * @param {Array} order Each truthy element indicates descending sort order for the corresponding criterion.
     */
    function compareMultiple(object, other, order) { // diff
        var index = -1,
            objCriteria = object.criteria,
            othCriteria = other.criteria,
            length = objCriteria.length;
    
        while (++index < length) {
            var result = baseCompareAscending(objCriteria[index], othCriteria[index]);
            if (result) {
[...]

PR #985, opened 2015-02-21
Implement descending order sort in #414.

octref:
Change `compareMultipleAscending(this, other)` to
`compareMultiple(this, other, props)`, which allows sorting in descending
order for a specific property if '-' is prepended to its name.

Change sortByAll so it can sort in descending order for a specific property.
For example,` _.sortByAll(users, ['user', '-age'])` sort in ascending
order for 'user' but in descending order for 'age'.

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -439,14 +439,19 @@
 
   /**
    * Used by `_.sortByAll` to compare multiple properties of each element
-   * in a collection and stable sort them in ascending order.
+   * in a collection and stable sort them in order specified by `props`.
+   *
+   * If a property name starts with "-", sort in descending order for that property.
    *
    * @private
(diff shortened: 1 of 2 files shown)


octref:
@jdalton 
Hi, can you take a look and tell me how you think about it?
I responded to that issue half a year ago and now finally implemented it :smiley: 

I signed the CLA. My dojo username is also octref.

jdalton:
Thanks @octref, but I want to avoid string magic for now.

octref:
@jdalton 
OK next time I'll post an issue first and ask how you think about it.

However, I'm wondering if you have a proposed implementation in mind? 
How do you think about this?

```javascript
_.sortByAll(myArray, [{ foo: true, bar: false })
``` 
for sorting foo ascending and bar descending.

PR #987, opened 2015-02-21
Add math methods

### LINKED ISSUE #687: [Feature Request] _.sum
I know that you can use _.reduce to calculate the sum of an collection, but I think that the sum is a very basic and often needed operation. Thus, a shorthand for it would be appropriate.

Compare the length of the following to lines:

``` javascript
_.reduce(array, function(sum, num) { return sum + num}, 0);
_.sum(array);
```

It could provide a _.pluck style shorthand for summing specific attributes:

``` javascript
_.reduce(array, function(sum, num) { return sum + num.attribute}, 0);
_.sum(array, "attribute");
```

I for myself often find myself trying to use _.sum and then I'm always surprised that it doesn't exist. Do you agree? Or is it inappropriate for lodash?

thejameskyle:
Resolves #687 and roadmap item

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -11159,6 +11159,44 @@
 
     /*------------------------------------------------------------------------*/
 
+    /**
+     * Adds two numbers.
+     *
+     * @static
+     * @memberOf _
+     * @category Math
+     * @param {number} augend The first number to add.
+     * @param {number} addend The second number to add.
+     * @returns {number} Returns the sum.
+     * @example
+     *
+     * _.add(6, 4);
+     * // => 10
+     */
+    function add(augend, addend) {
+      return augend + addend;
+    }
+
+    /**
+     * Calculates the sum of an array of numbers.
+     *
+     * @static
+     * @memberOf _
+     * @category Math
+     * @param {number} array The numbers to add.
+     * @returns {number} Returns the sum.
+     * @example
+     *
+     * _.sum([4, 6, 2]);
+     * // => 12
+     */
+    function sum(array) {
+      if (!isArray(array) || array.length === 0) return NaN;
+      return arrayReduce(array, add, 0);
+    }
+
+    /*------------------------------------------------------------------------*/
+
     // Ensure wrappers are instances of `baseLodash`.
     lodash.prototype = baseLodash.prototype;
 
@@ -11301,6 +11339,7 @@
     /*------------------------------------------------------------------------*/
 
     // Add functions that return unwrapped values when chaining.
+    lodash.add = add;
     lodash.attempt = attempt;
     lodash.camelCase = camelCase;
     lodash.capitalize = capitalize;
@@ -11370,6 +11409,7 @@
     lodash.sortedLastIndex = sortedLastIndex;
     lodash.startCase = startCase;
     lodash.startsWith = startsWith;
+    lodash.sum = sum;
     lodash.template = template;
     lodash.trim = trim;
     lodash.trimLeft = trimLeft;
--- a/test/test.js
+++ b/test/test.js
@@ -15098,6 +15098,28 @@
 
   /*--------------------------------------------------------------------------*/
 
+  QUnit.module('lodash.add');
+
+  (function() {
+    test('should add two numbers together', 1, function() {
+      var actual = _.add(6, 4);
+      equal(actual, 10);
+    });
+  }());
+
+  /*--------------------------------------------------------------------------*/
+
+  QUnit.module('lodash.sum');
+
+  (function() {
+    test('should return the sum of an array of numbers', 1, function() {
+      var actual = _.sum([6, 4, 2]);
+      equal(actual, 12);
+    });
+  }());
+
+  /*--------------------------------------------------------------------------*/
+
   QUnit.module('lodash(...).commit');
 
   (function() {
@@ -15579,6 +15601,7 @@
 
   (function() {
     var funcs = [
+      'add',
       'clone',
       'contains',
       'every',
@@ -15610,6 +15633,7 @@
       'parseInt',
       'pop',
       'shift',
+      'sum',
       'random',
       'reduce',
       'reduceRight',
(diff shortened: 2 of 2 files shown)


lodash.src.js
```diff
@@ -11156,6 +11156,211 @@
...
+     * @param {number} exponent
+     * @returns {number}
+     * @example
+     *
+     * _.power(6, 4);
+     * // => 1296
+     */
+    function power(base, exponent) {
+      return Math.pow(base, exponent);
+    }
```
megawac: what's the point of this one

lodash.src.js
```diff
@@ -11156,6 +11156,211 @@
...
+     * // => 4
+     *
+     * _.round(4.006, 2);
+     * // => 4.01
+     *
+     * _.round(4060, -2);
+     * // => 4100
+     */
+    function round(number, precision) {
+      if (precision) {
```
megawac: just default to 0? save a couple loc

lodash.src.js
```diff
@@ -11156,6 +11156,211 @@
...
+     * @example
+     *
+     * _.median([4, 6, 2]);
+     * // => 4
+     *
+     * _.median([4, 6, 2, 2]);
+     * // => 3
+     */
+    function median(array) {
+      if (!isArray(array)) return;
```
megawac: on all of these I'd say returning NaN would be most appropriate

thejameskyle:
Damn it travis

lodash.src.js
thejameskyle: This was to avoid the extra calculations for 0 and other falsy values.

lodash.src.js
thejameskyle: Completeness I guess, lodash-math had it.

lodash.src.js
thejameskyle: :+1:

thejameskyle: Would that include `_.median(1);`? Also, is it worth supporting `_.median(1, 2, 3);`?

megawac: same for empty arrays

megawac:
For the record I think these belong in separate packages such as mathp and _.math

lodash.src.js
```diff
@@ -11156,6 +11156,212 @@
...
+     * @param {number} array The numbers to add.
+     * @returns {number} Returns the sum.
+     * @example
+     *
+     * _.sum([4, 6, 2]);
+     * // => 12
+     */
+    function sum(array) {
+      if (!isArray(array) || array.length === 0) return NaN;
+      return reduce(array, add);
```
megawac: you'll want to put the initval to 0

phated:
Yay math

jdalton:
The travis fails are being trackered [here](https://github.com/travis-ci/travis-ci/issues/2838#issuecomment-75417443) (kind of) not sure what's causing them.

lodash.src.js
```diff
@@ -11156,6 +11156,212 @@
...
+     * _.median([4, 6, 2]);
+     * // => 4
+     *
+     * _.median([4, 6, 2, 2]);
+     * // => 3
+     */
+    function median(array) {
+      if (!isArray(array) || array.length === 0) return NaN;
+      var middle = (array.length + 1) / 2,
+          sorted = baseCreate(array).sort();
```
thejameskyle: reminder to self: test this `baseCreate`

d10:
I don't think add, subtract, multiply, divide and modulo are required, as they are supported natively, and adding them would just cause decreased performance as compared to the native operators.

jdalton:
Let's go with the less is more approach here since Math methods can snowball and a line to draw is kind of blurry. Let's start with the most requested one `sum` and `add` (as it's the base). Then we can add others as they are requested.

jdalton:
> I don't think add, subtract, multiply, divide and modulo are required, as they are supported natively, and adding them would just cause decreased performance as compared to the native operators

It's about function composition here, since there is no `add` method and the like that are built-in.

thejameskyle:
@d10 You wouldn't use something like `_.add(1, 2)` in place of `1 + 2`, you'd use it together with other methods in a chain or in partials. For example a more functional style version of the sum method included above:

```js
var sum = _.partial(_.reduce, _, _.add);
```

@jdalton Can I keep `_.round` too _pleeeease_? The precision is super useful. We do stuff like that at CloudFlare constantly converting bytes/requests/etc.

jdalton:
>  Can I keep _.round too pleeeease? The precision is super useful. We do stuff like that at CloudFlare constantly converting bytes/requests/etc.

I think we should wait for the demand to be there. I use `ceil` or `floor` more ;)

thejameskyle:
DEMAND

jk, will remove shortly

jdalton:
> DEMAND

On second look demand looks a little tiny :P

jdalton:
Your PR appears to be against 3.2.0 too so that'll need to be fixed too.

jdalton:
PR needs rebase.

lodash.src.js
Droogans: I thought the idea here wasn't to [call other lodash functions](https://github.com/lodash/lodash/issues/687#issuecomment-74453969) directly?

jdalton: > I thought the idea here wasn't to call other lodash functions directly?

It won't be, but I rather not nit the PR to death and get it in, then tweak after the fact.

thejameskyle: I'll fix this right now

thejameskyle: Actually is it even worth having `_.add` with this change? All it does it make the code less concise.

thejameskyle:
Alright, everything should be good. I kept a copy of the other methods if you ever want them in the future.

psirenny:
Hey, I was just wondering why modulo didn't get any love here. Should I create an issue for it and see if it gets some 👍 ?

jdalton:
@psirenny totally send in a PR.

PR #1005, opened 2015-02-27
Implement optional descending order sort in #414

octref:
As per #414 .

The only thing I'm not sure is the jsdoc. I wrote this signature for `_.sortByAll` 's `props`:

```
     * @param {...(string|string[]|Object[])} props
```
But technically, it should be an mixed array, where element can be an Object or string.
Should I change it to this?
```
     * @param {...(string|(string|Object)[])} props
```

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -439,14 +439,19 @@
 
   /**
    * Used by `_.sortByAll` to compare multiple properties of each element
-   * in a collection and stable sort them in ascending order.
+   * in a collection and stable sort them in the following order:
    *
+   * If orderArray is unspecified, sort in ascending order for all properties.
+   * Otherwise, for each property, sort in ascending order if its corresponding value in
+   * orderArray is true, and descending order if false.
+   * 
    * @private
    * @param {Object} object The object to compare to `other`.
    * @param {Object} other The object to compare to `object`.
+   * @param {Object} [orderArray] The order to sort by for each property.
    * @returns {number} Returns the sort order indicator for `object`.
    */
-  function compareMultipleAscending(object, other) {
+  function compareMultiple(object, other, orderArray) {
     var index = -1,
         objCriteria = object.criteria,
         othCriteria = other.criteria,
@@ -455,7 +460,7 @@
     while (++index < length) {
       var result = baseCompareAscending(objCriteria[index], othCriteria[index]);
       if (result) {
-        return result;
+        return (orderArray && orderArray[index]) ? result : -1 * result;
       }
     }
     // Fixes an `Array#sort` bug in the JS engine embedded in Adobe applications
@@ -6930,12 +6935,16 @@
      * This method is like `_.sortBy` except that it sorts by property names
      * instead of an iteratee function.
      *
+     * If using an array of object as `props`, a `desc` property can be specified
+     * to allow sorting in descending order.
+     *
      * @static
      * @memberOf _
      * @category Collection
      * @param {Array|Object|string} collection The collection to iterate over.
-     * @param {...(string|string[])} props The property names to sort by,
-     *  specified as individual property names or arrays of property names.
+     * @param {...(string|(string|Object)[])} props The properties to sort by,
+     *  specified as an individual property name, array of property names, or
+     *  array of objects that have the form { key: 'name', desc: true }.
      * @returns {Array} Returns the new sorted array.
      * @example
      *
@@ -6948,6 +6957,14 @@
      *
      * _.map(_.sortByAll(users, ['user', 'age']), _.values);
      * // => [['barney', 26], ['barney', 36], ['fred', 30], ['fred', 40]]
+     *
+     * // If desc is not provided, default to sort in ascending order
+     * _.map(_.sortByAll(users, [{ key: 'user' }, { key: 'age', desc: true }]), _.values)
+     * // => [['barney', 36], ['barney', 26], ['fred', 40], ['fred', 30]]
+     *
+     * // Using a mix of property name string and object
+     * _.map(_.sortByAll(users, ['user', { key: 'age', desc: true }]), _.values)
+     * // => [['barney', 36], ['barney', 26], ['fred', 40], ['fred', 30]]
      */
     function sortByAll(collection) {
       var args = arguments;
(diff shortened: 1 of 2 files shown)


jdalton:
> But technically, it should be an mixed array, where element can be an Object or string.
Should I change it to this?

Yah.

Can you cleanup the commits a bit. Also have you signed our [CLA](http://dojofoundation.org/about/claForm)?

octref:
I signed the CLA with the username octref.

I'll clean up the jsdoc. Anything else to change?

jdalton:
Ya the commits are a bit messy, it has a merge and an unrelated commit in there.

octref:
Done. I wanted to do a rebase on lodash/master but somehow got it wrong. Now it should be good.

jdalton:
I think this PR needs to more time to bake before including. There are several failing unit tests.
Also I'm still on the fence on the way to specify descending order. I'm thinking something like this is the least invasive.

```js
_.sortByAll(users, ['user', true, 'age', 'active', false]);
```

If the value after a key is a boolean it's used as the `desc` option for the preceding key.

octref:
Sure, take your time. I understand not many (are there any?) methods in lodash take option objects like this & you want to keep it consistent.

However I don't see any failing unit tests when running from cli `node test/test.js`. What tests are you referring to?

I just came up with an idea, how about making a function like this:

```
_.sortByOrder(users, ['user', 'age'], [true, false])
```
so that we could keep both `sortByAll` and `sortByOrder` simple.

jdalton:
> However I don't see any failing unit tests when running from cli node test/test.js. What tests are you referring to?

The travis-ci run indicator says it's failing:
https://travis-ci.org/lodash/lodash/builds/52379253

> I just came up with an idea, how about making a function like this:
```
_.sortByOrder(users, ['user', 'age'], [true, false])
```

So much nicer!

octref:
The problem it's failing, i think, is I renamed `compareMultipleAscending` to `compareMultiple`. I made this change since the new one accept a third argument, 
```
  function compareMultiple(object, other, orderArray) {
```
which can be used to specify sorting order, and default to ascending.

On Travis it's using lodash-cli to build it, and since it has
```
'sortByAll': ['baseEach', 'baseFlatten', 'baseSortBy', 'compareMultipleAscending', 'isIterateeCall', 'isLength'],
```
It builds the wrong functions and fails the tests.

--

Should I leave `compareMultipleAscending` as it is and create a new one, `compareMultiple`, or remove the `compareMultipleAscending` completely? 

- Only `sortByAll` depends on `compareMultipleAscending`
- In my version, `compareMultiple(object, other)` behaves exactly as `compareMultipleAscending(object, other)` used to be.

And for `sortByOrder`, my plan is:

```
// Sort by `user` ascendingly
_.sortByOrder(users, 'user')
_.sortByOrder(users, 'user', true)

// Sort by `user` descendingly
_.sortByOrder(users, 'user', false)
_.sortByOrder(users, ['user'], [false])

// `props` and `orders` not same length
_.sortByOrder(users, ['user', 'age'], [false]) // Sort by `user` descendingly but `age` ascendingly
_.sortByOrder(users, ['user'], [false, false]) // Second `false` in `orders` ignored.
```

What do you think? Anything to be careful about?

jdalton:
> The problem it's failing, i think, is I renamed compareMultipleAscending to compareMultiple. I made this change since the new one accept a third argument,

Naw the build is forgiving enough to leave unknown functions alone. No worries.

> Should I leave compareMultipleAscending as it is and create a new one, compareMultiple, or remove the compareMultipleAscending completely?

Create a new `compareMultiple`. Once we kick the implementation around we can remove or merge functionality as needed.

> And for sortByOrder, my plan is:

Looks good.

d10:
Major :+1:! Thanks for the awesome work, @octref!

octref:
@d10 
Nothing compared to what you guys are doing, but thanks!

The last commit get `sortByAll` and `compareMultipleAscending` back while adding `sortByOrder` and `compareMultiple`. Let me know if I did something wrong.

jdalton:
Merged! Woo!

octref:
Thanks I'll keep lurking around and see if I can help on anything :-)

ItsLeeOwen:
What did you end up implementing for sort order?  When I attempt to use what looks like the latest syntax the sort does not produce expected results.  

_.sortByAll(users, ["user", "age"], [true, true])

If I look at what baseSortByOrder is being passed for "props" and "order" I see:
baseSortByOrder: ["user", "age", true, true] []

ItsLeeOwen:
It's difficult to distinguish the intention of this line in sortByAll "baseSortByOrder(collection, baseFlatten(args, false, false, 1), [])" but changing the last two arguments to arg[1], arg[2] at least has it working on my end.  Although it does seem that desc true is actually asc.

The double array syntax is a bummer, but we have _.pluck

var sortBy = [
{key: 'user', desc: true},
{key: 'age', desc: false}
]
var sorted = _.sortByAll(input, _.pluck(sortBy, 'key'), _.pluck(sortBy, 'desc'))

octref:
@ItsLeeOwen 
We end up doing it this way:
- `sortByAll` unchanged.
- Add `sortByOrder` which behaves this way:
```
// Sort 'user' descendingly but 'age' ascendingly
_.sortByOrder(users, ['user', 'age'], [false, true])
```
The signature is `_.sortByOrder(collection, props, orders)`. Sort for each prop ascendingly by default / when its corresponding value in `orders` is true, otherwise descendingly.

`baseSortByOrder` should be a private method and not accessible from outside.

ItsLeeOwen:
@octref   Thanks, is commit 25b36703d068211302817b780d7119c3aed20de8 passing your tests for a sort of:_.sortByOrder(users, ['user', 'age'], [false, true])?

octref:
@ItsLeeOwen 
Yes. At least Travis says so.
And 25b3670 is another problem about `_.defaults` not working properly when called by `_.reduce` or `_.reduceRight`. You can see the details here.
https://github.com/lodash/lodash/pull/1014

ItsLeeOwen:
Gotcha, sorry I was calling _.sortByAll instead of _.sortByOrder.  _.sortByOrder is working fine.

http://codepen.io/ItsLeeOwen/pen/YPjGOv?editors=101

octref:
No problem :-)

PR #1008, opened 2015-02-27
Escape the backtick in escape description

mohsen1:
Noticed this little odd rendering in API docs
screen shot 2015-02-27 at 10 06 30 am

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -5999,7 +5999,7 @@ _.endsWith('abc', 'b', 2);
 ### <a id="_escapestring"></a>`_.escape([string=''])`
 <a href="#_escapestring">#</a> [&#x24C8;](https://github.com/lodash/lodash/blob/3.3.1/lodash.src.js#L9925 "View in source") [&#x24C9;][1] [&#x24C3;](https://www.npmjs.com/package/lodash.escape "See the npm package")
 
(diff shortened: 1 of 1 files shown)


jdalton:
Thanks, all you have to do is escape the `\`` in '\`'

mohsen1:
I noticed other methods don't wrap characters between backticks, but I'm not sure how it's possible to escape a backtick when it's not a inline code block.

mohsen1:
This is another way of doing it:

Converts the characters "&", "<", ">", '"', "'", and "`` ` ``", in `string` to

jdalton:
> Converts the characters "&", "<", ">", '"', "'", and "`", in string to

Let's go that route. I've fixed this on site.

mohsen1:
Updated the commit with that

jdalton:
Have you signed our [CLA](http://dojofoundation.org/about/claForm)?

jdalton:
Actually on second look simply escaping with \` works. Maybe not in github comments but in markdown in genera). See https://gist.github.com/jdalton/985387ddf9fd4e67cd21

mohsen1:
signed the CLA and updated the commit. I didn't know backslash escaping works in Markdown. Good to know!

Issue #1016, opened 2015-03-04
Add string built-ins to lodash wrapper?


jasonkarns:
It's super handy that a bunch of the built-in `Array` methods (`concat`, `join`, `pop`, `push`, etc) are all added to the wrapper object. Thus when an array is wrapped, one can use any of the built-in array methods directly on the chain as if operating directly on the array.

Now that lodash has a suite of String methods, I was a bit surprised to learn that the `String` built-ins *aren't* added to the wrapper object as well.

Was this intentional or an oversight? I would love to be able to use methods like `split` or `replace` directly on a lodash-wrapped string.

jdalton:
I can dig a couple helpers. Let's open it to PRs :D

megawac:
If these are back on the table it may be a good time to reconsider https://github.com/lodash/lodash/issues/821?

jdalton:
Still not wanting method proxies. Chaining is pretty scenario specific.

octref:
I guess I can help :-)
I'll spend today and tomorrow looking for some possible implementations & post them here & do the PR in this weekend.

jdalton:
Rock, let's reuse the loop that's handling the array methods, adding just `replace`, and `split`, and toggle based on a regexp match of their names similar to how it's handled in other places.

d10:
:+1: `split` and `replace` are a good start!

jdalton:
Taking it to #1016.

PR #1021, opened 2015-03-06
WIP: StringWrapper implementation for #1016

octref:
#1016 

Just an attempt. Still **WIP**.
I'd like to get some feedback before going too deep, in case I did something wrong.
No `split` and `replace` yet.

### What you can do now?
```
_("hiya")
// -> StringWrapper {__wrapped__: "hiya", __actions__: Array[0], __chain__: true, constructor: function, after: function…}
_("hiya").capitalize().value()
// -> "Hiya"
_("hiya").capitalize().pad(6, '_').repeat(5).value()
// -> "_Hiya__Hiya__Hiya__Hiya__Hiya_"
_("hiya").capitalize().pad(6, '_').repeat(5).words().value()
// -> ["Hiya", "Hiya", "Hiya", "Hiya", "Hiya"]
```

One question I have:
#### Any better idea to detect if an `action` is a String method?
  - In `baseStringWrapperValue`, I listed all string methods to check if the actions are string methods.
This is definitely not a good solution as we'd have to change this when we add new string methods.
  - In `baseStringWrapperValue`, I'm using `action.func.name` to check the function name, however, for methods like `camelCase` which uses `createCompounder` to return a anonymous func, it's empty.
I'm allowing anonymous functions to be performed now.

Those kind of tests don't pass for now, as expected:
```
    test('should return an unwrapped value when implicitly chaining', 1, function() {
      if (!isNpm) {
        strictEqual(_('fred').capitalize(), 'Fred');
      }
      else {
        skipTest();
      }
    });
```

I'll fix tests later.  (submitted a CLA on 2015-02-21 03:34:30)

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -858,6 +858,9 @@
      * `concat`, `join`, `pop`, `push`, `reverse`, `shift`, `slice`, `sort`, `splice`,
      * and `unshift`
      *
+     * These `String` methods are also available:
+     * `split` and `replace`
+     *
      * The wrapper methods that support shortcut fusion are:
      * `compact`, `drop`, `dropRight`, `dropRightWhile`, `dropWhile`, `filter`,
      * `first`, `initial`, `last`, `map`, `pluck`, `reject`, `rest`, `reverse`,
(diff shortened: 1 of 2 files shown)


jdalton:
Eeeek totally wrong direction here. We don't want to change the chaining behavior. Devs can use `.chain()` to force chaining of strings.

jdalton:
See [here](https://github.com/lodash/lodash/blob/3c01ecccc4a85fdef406da1f6aa45d72041cec03/lodash.src.js#L11710). The idea would be to just add `'split', 'replace'` to values and a regexp to detect them, `/^(?:replace|split)$/` to determine which built-in prototype to use `arrayProto` or `stringProto`.

octref:
What if the `LodashWrapper` is not a string? I thought we should only allow functions in String section to be chained when the wrapped value is a string.

jdalton:
You can currently pass junk to all our methods, we throw in only a select few cases.

octref:
OK. I was overthinking it. Thanks for your pointer! I'll try to do it again tmr as you suggested.

jdalton:
You don't have to close this, you can rebase it.

octref:
Sure. 
BTW I was thinking maybe `replace` would be a good addition #1022 .

octref:
Just wondering. Do you have a reason to keep these non-capturing groups `?:` ?
```
isStringFunc = /^(?:split|replace)$/.test(methodName),
chainName = /^(?:push|sort|unshift)$/.test(methodName) ? 'tap' : 'thru',
fixObjects = !support.spliceObjects && /^(?:pop|shift|splice)$/.test(methodName),
retUnwrapped = /^(?:join|pop|shift)$/.test(methodName);
```

jdalton:
> Do you have a reason to keep these non-capturing groups ?:

Makes it clear my intent, not concerned over the few characters it costs.

PR #1041, opened 2015-03-11
Micro Optimisation - Ternary Operator

georgePadolsey:
Turn plus ternary operator into bitwise or operator. This is slightly faster;
See http://jsperf.com/parseint-vs-unary-plus


jdalton:
No.

PR #1042, opened 2015-03-12
Add an example of the alternative zipObject syntax

bryce-gibson:
I usually glance at the examples before the actual blurb, and a number of times I've assumed zipObject doesn't do what I need because there's no example.

This is a tiny doc change to add an example of the two dimensional array usage.

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -5636,10 +5636,10 @@
      * @returns {Object} Returns the new object.
      * @example
      *
-     * _.zipObject(['fred', 'barney'], [30, 40]);
+     * _.zipObject([['fred', 30], ['barney', 40]]);
      * // => { 'fred': 30, 'barney': 40 }
-     * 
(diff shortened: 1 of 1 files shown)


octref:
I think you'd better follow the current coding style
So change 
`_.zipObject([ ['fred', 30], ['barney', 40] ]);` 
to 
`_.zipObject([['fred', 30], ['barney', 40]]);`
And in the doc this style `[[key1, value1], [key2, value2]]` comes before the `_.zipObject(props, values)`. So better put this example before the current one.

But let's hear what @jdalton would say. You also need to sign the [CLA](http://dojofoundation.org/about/claForm).

jdalton:
Thank you @bryce-gibson! Very nice addition. :+1: on @octrefs suggestions too.

bryce-gibson:
Cool, done I think? :-)

I've done the CLA online form :-)

jdalton:
Thanks!

PR #1050, opened 2015-03-13
Fix `_.reduce` doc

octref:


d10:
Thanks for the great work you've been putting in here, @octref!

lodash.src.js
```diff
@@ -6599,7 +6599,7 @@
      * each element in `collection` through `iteratee`, where each successive
      * invocation is supplied the return value of the previous. If `accumulator`
      * is not provided the first element of `collection` is used as the initial
-     * value. The `iteratee` is bound to `thisArg`and invoked with four arguments;
+     * value. The `iteratee` is bound to `thisArg` and invoked with four arguments:
```
d10: Actually, I'm not too sure about the semicolon at the end. @jdalton seems to prefer that, but it might be more natural to use colons instead. Opinions?

octref:
:wink:

PR #1056, opened 2015-03-17
Allocate proper size

jridgewell:
Since we're putting `rest` at `start` index, `array.length = {start + 1}`


jdalton:
Thanks!

jdalton:
When you first suggested it, it didn't click. Really digging `_.restParam` woooo!

jridgewell:
All good, I had to test it once or twice to make sure I wasn't going crazy.

PR #1057, opened 2015-03-17
fix _.zipObject docs to more clearly demonstrate it as _.pairs' inverse

funkenstein:
I spent a while today attempting to look for a method that was effectively the inverse of `_.pairs`. Although the docs do point out that `_.zipObject` does this, it's hard to find with Ctrl+F and there are no actual examples of it which is what I usually scan through when looking for things. This just clears that up. :)


jdalton:
Cool. Have you signed our [CLA](http://dojofoundation.org/about/claForm)?

funkenstein:
Wasn't aware you had one, just signed it though.

jdalton:
Thanks!

PR #1062, opened 2015-03-20
exposed createAggregator as _.createAggregator

Sigfried:
I wanted to make my own aggregator. Specifically, a groupBy that would group on array fields and allow records to appear in multiple groups.

```
var multiValuedGroupBy = _.createAggregator(function(result, value, keys) {
    _.each(keys, function(key) {
    if (_.has(result, key)) {
        result[key].push(value);
    } else {
        result[key] = [value];
    }
    });
})
```
So, `multiValuedGroupBy([{A:[1,2],B:0},{A:[1,3],B:0}], 'A')` produces:

    {
         "1":[{"A":[1,2],"B":0},{"A":[1,3],"B":0}],
         "2":[{"A":[1,2],"B":0}],
         "3":[{"A":[1,3],"B":0}]
     }

I couldn't just make a `multiValuedGroupBy` mixin because I needed access to createAggregator.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -11329,6 +11329,9 @@
     lodash.include = includes;
     lodash.inject = reduce;
 
+    // expose createAggregator so people can make their own aggregators
+    lodash.createAggregator = createAggregator;
+
     mixin(lodash, (function() {
       var source = {};
       baseForOwn(lodash, function(func, methodName) {


jdalton:
It looks like `createAggregator` is polished enough to be exposed but I want to hold off and see if there's a stronger demand for it (then we can bikeshed a public facing name for it)

That said if you're in [node](https://github.com/lodash/lodash/tree/3.5.0-npm) or [amd](https://github.com/lodash/lodash/tree/3.5.0-npm) you can have access to `createAggregator`.

In node:
```js
var createAggregator = require('lodash/internal/createAggregator')`
```
or
```js
var createAggregator = require('lodash._createaggregator')`
```
using the amd branch:
```js
require({
  'packages': [
    { 'name': 'lodash', 'location': 'path/to/lodash' }
  ]
}, ['lodash/internal/createAggregator'], function(createAggregator) {
  // use `createAggregator`
});
```

Sigfried:
Thank you!

I was trying to support amd, but then got confused. If my library (
https://github.com/sigfried/supergroup) gets any attention, I'll go back to
supporting it and do this the way you're suggesting. For now I'm just
including a custom copy of lodash.js in my repo.

jdalton:
Could you browserify just the `createAggregator` module instead of creating an lodash.js that way it can be companioned easier.

bkoplin:
+1 demand for _.supergroup! Use it all the time and wold love to see it native in Lodash. It's _incredibly_ useful.

Sigfried:
Wow. That's so nice to hear, @bkoplin. I had no idea anyone but me was using it. I don't think it's ready for lodash at this point -- and even though I've written it as a lodash mixin, I wonder if belongs more with d3 than lodash. Anyway, contact me offline and maybe we can figure out how best to support its use.

@jdalton -- sorry, I don't think I ever followed your suggestion -- because I thought I was the only person using supergroup. I'll talk with @bkoplin and see if it seems worth doing this.

Thanks!

Issue #700, opened 2014-09-02
Get a value inside an object at a provided path.


kaewka:
Hi there,

I am trying to use lodash to get a value inside an object at a provided path.
Similar to `Y.Object.getValue` in YUI3:
http://yuilibrary.com/yui/docs/api/files/yui_js_yui-object.js.html#l305

For example:

``` js
Y.Object.getValue({a:{b:{c:"1234"}}}, ["a","b", "c"])
// => "1234"
```

PR #1081, opened 2015-03-26
Adding _.resultDeep() to resolve #1060, #700, et al.

### LINKED ISSUE #1060: Add function to get nested property value
The only thing I really miss from lodash is the `get` function that MOUT implements to get nested property value, returning `undefined` if it doesn't exist.

MOUT API Documentation: http://moutjs.com/docs/latest/object.html#get

Would it be possible to implement a similar function for lodash?

jpiccari:
I saw that #1060 had lost momentum so thought I'd revive it with a PR.

I choose to create a new `_.resultDeep()` since the functionality seemed to me like an extension of what `_.result()` does. Not sure if anyone has strong feelings about the naming or arguments but at least now we have some code in front of us to hash this out.

I'm also hoping for suggestions to tidy up the `dropRight()` + `last()` combo. I had originally used a for-loop but ended up trading the performance gains for fewer bytes and better maintainability. Thoughts?

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -11224,6 +11224,31 @@
     }
 
     /**
+     * Creates a function which returns the property value of `key` on a given object.
+     *
+     * @static
+     * @memberOf _
+     * @category Utility
+     * @param {array} keyPath The key path of the property to get.
+     * @returns {Function} Returns the new function.
+     * @example
+     *
+     * var highScores = [
+     *   { user: 'fred', score: 850 },
+     *   { user: 'barney', score: 630 }
+     * ];
+     *
+     * var getHighestScore = _.propertyDeep([0, 'score']);
+     * getHighestScore(highScores);
+     * // => 850
+     */
+    function propertyDeep(keyPath) {
+      return function(object) {
+        return resultDeep(object, keyPath);
+      };
+    }
+
+    /**
      * The opposite of `_.property`; this method creates a function which returns
      * the property value of a given key on `object`.
      *
@@ -11245,6 +11270,36 @@
     function propertyOf(object) {
       return function(key) {
         return object == null ? undefined : object[key];
+      };
+    }
+
+    /**
+     * The opposite of `_.propertyDeep`; this method creates a function which returns
+     * the property value of a given key path on `object`.
+     *
+     * @static
+     * @memberOf _
+     * @category Utility
+     * @param {Object} object The object to inspect.
+     * @returns {Function} Returns the new function.
+     * @example
+     *
+     * var users = [
+     *   { user: 'fredrick', nickname: 'fred' },
+     *   { user: 'bernard', nickname: 'barney' }
+     * ];
+     *
+     * var propertyOfUsers = _.propertyDeepOf(users);
+     *
+     * propertyOfusers([0, 'nickname']);
+     * // => 'fred'
+     *
+     * propertyOfusers([1, 'name']);
+     * // => 'bernard'
+     */
+    function propertyDeepOf(object) {
+      return function(keyPath) {
+        return resultDeep(object, keyPath);
       };
     }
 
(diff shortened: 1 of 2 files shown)


lodash.src.js
```diff
@@ -9777,6 +9777,44 @@
...
+     *
+     * _.result(object, ['user', 'dates', 'anniversary']);
+     * // => '07/29/2012'
+     *
+     * _.result(object, ['status', 'timestamp'], _.constant('busy'));
+     * // => 'busy'
+     */
+    function resultDeep(object, keyPath, defaultValue) {
+      arrayEach(dropRight(keyPath), function(key) {
+        return object = result(object, key);
```
d10: By returning the result you could stop iteration early if one of the values is `false`.

jdalton: @d10 I think that's the idea though it would have to be explicit `false`. So it would be better to do:
```js
object = result(object, key);
return object != null;
```

jpiccari: Correct, this is intentional to avoid excessive looping.

How about taking it a step further and simply casting the result to a boolean? I can't think of any falsey values that have enumerable properties, so it seems safe to me (albeit a bit cryptic for passersby).
```javascript
return !!(object = result(object, key));
```

jdalton: Naw, nullish is the way to go, 0..toFixed()

jpiccari: Touchè, updated!

octref:
Have been busy with exams and GSOC. Lost some momentum but not enthusiasm :wink: 

Looks good to me. But while looking at the source I'm wondering if we could base `_.result` on `_.baseProperty`.
```
    function baseProperty(key) {
      return function(object) {
        return object == null ? undefined : object[key];
      };
    }

    function result(object, key, defaultValue) {
      var value = object == null ? undefined : object[key];
      if (typeof value == 'undefined') {
        value = defaultValue;
      }
      return isFunction(value) ? value.call(object) : value;
    }
```

jpiccari:
I was also looking around the code for something that could be useful, but came up empty-handed. Using `baseProperty` looks like it would be possible but seems very gc intensive (creating and destroying functions in a loop).

jdalton:
Can you add a `_.propertyDeep` in this PR too.

octref:
@jdalton 
I implemented `propertyDeep` yesterday, but still thinking about how should I update the doc.

Do I just add a paragraph to this verbatim piece?

> If a property name is provided for predicate the created _.property style callback returns the property value of the given element. 
If a value is also provided for thisArg the created _.matchesProperty style callback returns true for elements that have a matching property value, else false. 
If an object is provided for predicate the created _.matches style callback returns true for elements that match the properties of the given object, else false.

I think it's time to extract information for `predicate`, `iteratee`, etc and put them into a "Introduction to Lodash" piece.

jdalton:
> I think it's time to extract information for predicate, iteratee, etc and put them into a "Introduction to Lodash" piece.

Let's not get side tracked. The docs for `_.propertyDeep` would be similar to that of [_.property](https://lodash.com/docs#property), probably beginning like "This method is like `_.property`except..."

octref:
But I also made it work with `baseCallback`, so now you could do
```
var foo = {
  bar: {
    baz: 'qux'
  }
};
_.propertyDeep(['bar', 'baz'])(foo);
// => 'qux'
_map([foo], ['bar', 'baz'])
// => ['qux']
```

I was thinking about adding documentation to all `iteratee` since now they can take an array and use `_.propertyDeep` to form a callback now.

I guess I'll first make a PR for `propertyDeep`.

jdalton:
> But I also made it work with `baseCallback`

Let's hold off on that.

lodash.src.js
```diff
@@ -11210,6 +11274,36 @@
...
+     * @example
+     *
+     * var users = [
+     *   { user: 'fredrick', nickname: 'fred' },
+     *   { user: 'bernard', nickname: 'barney' }
+     * ];
+     *
+     * var propertyOfUsers = _.propertyDeepOf(users);
+     *
+     * propertyOfusers([0, 'nickname']);
```
d10: This should be `propertyOfUsers`.

jdalton:
Closed via https://github.com/lodash/lodash/commit/3d0beb1a2e85271f97c8800e60f1977ab8ac8a16.

PR #1082, opened 2015-03-26
Don't mutate array during #remove

### LINKED ISSUE #1079: Question about _.remove index arg behavior
```javascript
var x = [0, 1, 2, 3, 4, 5];
_.remove(x, function (e, i, a) {
    console.log(e, i, a); return i % 2 === 0;
}); //=> [0, 1, 2, 3, 4, 5]
x//=> 0
```

Logs:
```
0 0 [0, 1, 2, 3, 4, 5]
1 0 [1, 2, 3, 4, 5]
2 0 [2, 3, 4, 5]
3 0 [3, 4, 5]
4 0 [4, 5]
5 0 [5]
```

Looking at the source, one sees that `_.remove` splices the array/decrements index as it removes elements, so that in the above case index will always be 0. Was there a reason this behavior was chosen (I don't see a test case which is immediately relevant, so perhaps it's an edge case)? This behavior can be changed by iterating backwards instead of forwards, I believe. Either way, perhaps could use some documentation - would be happy to submit a PR.

jridgewell:
Closes #1079.

I’m not sure how this’ll align with Lodash’s current coding conventions. Please review.

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -5310,16 +5310,31 @@
     function remove(array, predicate, thisArg) {
       var index = -1,
           length = array ? array.length : 0,
-          result = [];
+          result = [],
+          removes = 0,
+          indexes = [],
+          last;
 
       predicate = getCallback(predicate, thisArg, 3);
       while (++index < length) {
         var value = array[index];
         if (predicate(value, index, array)) {
+          if (last && last[0] + last[1] === index) {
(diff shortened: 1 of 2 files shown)


jdalton:
Cool. I was just finishing a patch. I'll pull this in for the unit test though.

PR #1087, opened 2015-03-27
Add `_.propertyDeep` for #1060

octref:
Lemme know if I did anything wrong!

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -2599,6 +2599,28 @@
     }
 
     /**
+     * The base implementation of `_.propertyDeep` which does not coerce `keys` to strings.
+     *
+     * @private
+     * @param {string[]} keys Keys that specify the property to get.
+     * @returns {Function} Returns the new function.
+     */
+    function basePropertyDeep(keys) {
+      return function(object) {
+        var index = -1,
+            length = keys.length;
+
+        while (++index < length) {
+          object = baseProperty(keys[index])(object);
+          if (object == null) {
+            return undefined;
+          }
+        }
+        return object;
+      };
+    }
+
+    /**
      * The base implementation of `_.pullAt` without support for individual
      * index arguments and capturing the removed elements.
      *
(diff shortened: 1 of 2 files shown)


octref:
Reference #1060 , so people can find this PR from there.

lodash.src.js
```diff
@@ -11204,6 +11226,34 @@
...
+     * var getAge  = _.propertyDeep(['user', 'age']);
+     *
+     * _.map(users, getName);
+     * // => ['fred', 'barney']
+     *
+     * _.map(users, getAge);
+     * // => [undefined, 36]
+     */
+    function propertyDeep(keys) {
+      keys = baseMap(keys, function(key) { return key + '' });
```
megawac: you can do baseMap(String) here

lodash.src.js
```diff
@@ -2599,6 +2599,28 @@
...
+     * @param {string[]} keys Keys that specify the property to get.
+     * @returns {Function} Returns the new function.
+     */
+    function basePropertyDeep(keys) {
+      return function(object) {
+        var index = -1,
+            length = keys.length;
+
+        while (++index < length) {
+          object = baseProperty(keys[index])(object);
```
megawac: what's the reason to use baseProp here instead of a simple accessor

lodash.src.js
octref: @megawac Nice, thanks!

octref: But I guess maybe `baseToString` is a better option?

megawac: Yeah, probs

jdalton: `arrayMap`.

octref: `arrayMap` doesn't allow falsey first argument though. Should I make `propertyDeep` rejectFalsey?

lodash.src.js
octref: Right, my bad.

jdalton:
Can you add `propertyDeepOf` too.

octref:
Sure thing, give me an hour.

PR #1089, opened 2015-03-28
Inconsistent results from _.propertyDeep and _.propertyDeepOf

jpiccari:
I noticed some inconsistencies in methods newly added to the API. `_.propertyDeepOf()` internally uses `resultDeep()` while `_.propertyDeep()` uses `basePropertyDeep()`. One implementation calls function encountered in the key path, passing the result as the object of subsequent key lookups. The other returns the reference to functions it encounters.

```javascript
var object = Object,
  keys = ['prototype', 'toString'];
_.propertyDeep(keys)(object);
// => function toString() { [native code] }
_.propertyDeepOf(object)(keys);
// => "[object Object]"
```

After looking at `_.property()`, I'm guessing the intended behavior is to **not** invoke encountered functions. To reconcile the results of these methods I generalized `basePropertyDeep()` to suit both methods by returning the property at the end of the key path rather than a function. This allows the method to be used by both `_.propertyDeep` & `_.propertyDeepOf` which each create their own functions.

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -2599,28 +2599,28 @@
     }
 
     /**
-     * The base implementation of `_.propertyDeep` which does not coerce `keys` to strings.
+     * The base implementation of `_.propertyDeep` and `_.propertyDeepOf` which
+     * does not coerce `keys` to strings and takes `object` as a parameter.
      *
      * @private
+     * @param {object} object Object to look for keys.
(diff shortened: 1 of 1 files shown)


jpiccari:
Looks like I was an hour too slow on this change. Closing this pull request.

Issue #1086, opened 2015-03-27
Method helper


gampleman:
A handly lodash extension we made is this one:

```javascript
  _.method = function() {
      var args, meth;
      meth = arguments[0], args = 2 <= arguments.length ? [].slice.call(arguments, 1) : [];
      return function(object) {
        return object[meth].apply(object, args);
      };
    }
```

This allows to create a function from a method:

```javascript
var toStr = _.method('toString');
toStr([3]);  // => '3'
toStr(3);    // => '3'
toStr({a: 3}) // => '[object Object]'

var matchesTag = _.method('match', /^(\w+):(\w+)=(.+)$/)
matchesTag('ns:key=value'); // => ['ns:key=value', 'ns', 'key', 'value']
```

Also very handy for iteration:

```javascript
var waysToSumToEleven = _.map(_.zip(_.range(1, 10), _.range(10, 1, -1)), _.method('join', ' + '))
// => ["1 + 10", "2 + 9", "3 + 8", "4 + 7", "5 + 6", "6 + 5", "7 + 4", "8 + 3", "9 + 2"]
```

I think it would make a nice addition.

jdalton:
Something like:
```js
var matchesTag = _.bind(_.call, ''.match, _, /^(\w+):(\w+)=(.+)$/);
matchesTag('ns:key=value'); // => ['ns:key=value', 'ns', 'key', 'value']
```
The name `method` is pretty generic. We have a `bindKey` which is close but not right. This is like a `partial`+`invoke`.

I think I've seen this request before so I'll note a :+1: for it. Let's keep an eye on it and see if demand picks up.

jdalton:
Something like:
```js
var matchesTag = _.bind(_.call, ''.match, _, /^(\w+):(\w+)=(.+)$/);
matchesTag('ns:key=value'); // => ['ns:key=value', 'ns', 'key', 'value']
```
The name `method` is pretty generic. We have a `bindKey` which is close but not right. This is like a `partial`+`invoke`.

I think I've seen this request before so I'll note a :+1: for it. Let's keep an eye on it and see if demand picks up.

jridgewell:
I had a need for something like this [yesterday](https://github.com/jashkenas/underscore/pull/2124#issuecomment-86682834)...

jdalton:
Like an `_.invokeOf` or something.

jridgewell:
[...]

PR #1091, opened 2015-03-29
_.method and _.methodOf

jridgewell:
Here's a basic implementation of `_.method` and `_.methodOf` from #1086.

I'm noticing a bit of duplication wherever `getPath` is used for deep properties. What are you thoughts on returning a tuple `[object, value]`?

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -10197,6 +10197,125 @@
 
   /*--------------------------------------------------------------------------*/
 
+  QUnit.module('lodash.method');
+
+  (function() {
+    test('should create a function that calls a property function of a given object', 3, function() {
+      var object = { 'a': _.constant(1), 'b': _.constant(2) },
+          method = _.method('a');
+
+      strictEqual(method.length, 1);
+      strictEqual(method(object), 1);
+
+      method = _.method('b');
+      strictEqual(method(object), 2);
+    });
+
+    test('should work with non-string `prop` arguments', 1, function() {
+      var array = _.times(3, _.constant),
+          prop = _.method(1);
+      strictEqual(prop(array), 1);
+    });
+
+    test('should coerce key to a string', 1, function() {
+      function fn() {}
+      fn.toString = _.constant('fn');
+
+      var values = [null, undefined, fn, {}];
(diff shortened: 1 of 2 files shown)


jdalton:
> I'm noticing a bit of duplication wherever `getPath` is used for deep properties. What are you thoughts on returning a tuple `[object, value]`?

Things are still in flux. I haven't documented or added tests yet so a bit early for cleanup.

test/test.js
```diff
@@ -135,7 +135,7 @@
   var isStrict = ui.isStrict;
 
   /** Used to test Web Workers. */
-  var Worker = !(ui.isForeign || ui.isSauceLabs || isModularize) && document && root.Worker;
+  var Worker = !(ui.isForeign || ui.isSauceLabs || isModularize) && document && document.origin != 'null' && root.Worker;
```
jdalton: What's the `document.origin` check about?

jridgewell:
Oh, Chome won't let you start a Worker if you're using a filepath as the URL. I can amend it out if you want.

test/test.js
jridgewell: Oh, Chome won't let you start a Worker if you're using a filepath as the URL. I can amend it out if you want.

jdalton: Can you open a separate PR for it.

jdalton:
I have time to review this a bit more by mid week once I've added unit tests for the deep property enhancements.

jridgewell:
Take you time. I'll keep it following the changes you make elsewhere.

test/test.js
```diff
@@ -10197,6 +10197,125 @@
...
+    test('should work when `object` is nullish', 1, function() {
+      var values = [, null, undefined],
+          expected = _.map(values, _.constant(undefined));
+
+      var actual = _.map(values, function(value, index) {
+        var method = _.method('a');
+        return index ? method(value) : method();
+      });
+
+      deepEqual(actual, expected);
```
gampleman: I'd expect this to throw to be honest. You're calling an undefined method of an object - that should be an error, not a silent failure IMO.

PR #1109, opened 2015-04-05
Modify add() to parseFloat() variables passed

James-Dunn:
Now `_.add('6', '4');` will output 10 instead of 64.

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -11559,7 +11559,7 @@
      * // => 10
      */
     function add(augend, addend) {
-      return parseFloat(augend) + parseFloat(addend);
+      return Number(augend) + Number(addend);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


StreetStrider:
This case is just simple cast from string to number .This is a job for `Number` constructor, not `parseFloat`.

James-Dunn:
Just depends on what behaviour is desired.

Number('') = 0;
Number(false) = 0;
Number('1foo') = NaN;

parseFloat('') = NaN;
parseFloat(false) = NaN;
parseFloat('1foo') = 1;

StreetStrider:
@James-Dunn, yes, the difference is in cast — parse. This is cast.

jdalton:
The `_.sum` behavior is to simply coerce values to a number `+value || 0`.

jdalton:
@James-Dunn I'd be for coercing to a number but not `parseFloat`. Have you signed our [CLA](http://dojofoundation.org/about/claForm)?

James-Dunn:
Changed it to Number() and signed CLA

PR #1110, opened 2015-04-05
Add .npmignore to exclude unnecessary development files from npm package

panuhorsmalahti:
I copied .gitignore as .npmignore and added a few development files and folders to be excluded from the npm package as they're unnecessary. Also removed node_modules from .npmignore as it's unnecessary as per npm's documentation.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,12 @@
+.DS_Store
+*.custom.*
+*.log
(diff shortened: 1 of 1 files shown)


jdalton:
Wrong branch for npm. In our npm branch we already have it taken care of.

Issue #1084, opened 2015-03-26
Extra String utilities?


jh3y:
Hey.

It may not be necessary but I recently read through the API and noticed there were no String utilities for things like replacing all instances of a string within a string. Is this because the native way of doing this is good enough or these type of functions don't fit in with lodashs design?

For example; `_.replaceAll('abcdefg, 'abc', 'x')` would cover something like

```javascript
var string = 'abcdefg';
string = string.replace(new RegExp('abc', 'g'), 'x');
```

It was just a `question` as I'd imagine this has been considered before but I couldn't find the issue in the history. It'd be nice to chain something like that in some instances.

@jh3y

arthurvr:
I think the native methods cover this quite well.

```js
'abcdefg'.replace('abc', 'x');
//=> "xdefg"
```

StreetStrider:
@arthurvr, he speaks about «**all** instances of a string within a string».
Your code will replace only first entry:
```javascript
'abcdefgabc'.replace('abc', 'x');
// 'xdefgabc'
```
The proper way to do it is to use RegExp:
```
'abcdefgabc'.replace(/abc/g, 'x');
// 'xdefgx'
```

I think the reason why @jh3y asks it is because RegExp literal has no simple way to interact with strings. So, if you wanna to replace all occurences of string, but this string is unknown (i.e. we're using variable), this would be ugly.

StreetStrider:
```javascript
function replaceAll (str, from, to)
{
  from = new RegExp(from, 'g');
  return str.replace(from, to);
}

console.log(replaceAll('abcdefgabc', 'abc', 'x'));
```

arthurvr:
@StreetStrider Yup. Sorry!

jh3y:
Yeah @StreetStrider , that's what I was getting at. It abstracts away from the user having to write this `replaceAll` utility function within their code when doing replacements.

'''javascript
function replaceAll(str, from, to) {
  return str.replace(new RegExp(from, 'g'), to);
}
'''

It was just to see whether this had been brought up before.

@jh3y

jdalton:
If you go that route don't forget to use [_.escapeRegExp](https://lodash.com/docs#escapeRegExp) on the `from` param. I think the normal `String#replace` is fine, but if there's a demand I'll reevaluate.

katherine-barn2door:
Probably can't vote anymore, but :+1: for replaceAll(), especially since it isn't in native.

jdalton:
[...]

PR #1123, opened 2015-04-11
Add `_.replaceAll`

mcharytoniuk:
This pull request contains `_.replaceAll` static function. Default behavior of JavaScript's `String.prototype.replace` is to replace only first occurence of a string (unless `RegExp` is used) and `_.replaceAll` replaces all occurences. 

```JavaScript
_.replaceAll('b-c-d', '-', '.');
// => 'b.c.d'
```

It could be eventually further improved in the future to work in a chain like this:

```JavaScript
_(['a.b.c', 'def']).replaceAll('.', '..').value(); 
// => ['a..b..c', 'def']
```


StreetStrider:
@mcharytoniuk, how to use RegExp `i` flag in this function?

mcharytoniuk:
The goal of this function is not to use RegExp to replace strings. `RegExp ` it's an implementation detail here.

jdalton:
Related to #1084.

jdalton:
Hi @mcharytoniuk. Do you have a use case in mind?

mcharytoniuk:
I just use this pretty often and usually end up with something like this:
https://github.com/goreutils/gore-gulp/blob/69fa3d57b7a35beb8936aaec5ec559b2f64570b9/gore-gulp/webpack.js#L68

I find this kludgy, unreliable and repetetive. I would love to see such function in a popular library so it would be much more tested and reliable. Possibly some other parts of lodash could benefit from it in the future and stay more maintainable. It allows to do this simple operation in one function call instead of creating loops or regular expressions. I use lodash in most of my projects so I would not have to repeat myself.

jdalton:
@mcharytoniuk can you create a [jsperf](https://jsperf.com) of that and this implementation.

mcharytoniuk:
Yes, I can but loop + indexOf implementation is not generic because it would fail in the following case:

```JavaScript
replace('a/b/c', '/', '//');
```

It would end in an endless loop producing something like this:

```JavaScript
'a/////////////////////////////(...)'
```

I would have to store the index of last replacement somewhere and look for occurence after given replacement and do some other string manipulations which would be compilcated for such thing. I have also this implementation in mind:

```JavaScript
function replaceAll(string, pattern, replacement) {
  string = baseToString(string);
  pattern = baseToString(pattern);

  if (!pattern) {
    return string;
  }

  replacement = baseToString(replacement);
  
  string = string.split(pattern);

  return string.join(replacement);
}
```

I'll creat jsperf soon.

mcharytoniuk:
jsperf: http://jsperf.com/lodash-string-replace-all

jdalton:
@mcharytoniuk Thanks! This won't make it into 3.7.0 but may make it into the version after. I'll kick it around a bit more.

mcharytoniuk:
No problem. :)

lodash.src.js
```diff
@@ -10610,6 +10610,40 @@
...
+     * // => 'a//b//c'
+     *
+     * _.replaceAll('aabbaabbaacc', 'aa', 'd');
+     * // => 'dbbdbbdcc'
+     */
+    function replaceAll(string, pattern, replacement) {
+      string = baseToString(string);
+      pattern = baseToString(pattern);
+
+      if (!pattern) {
```
megawac: this doesn't seem strictly necessary, replace all(str, '', seperator) seems valid

jdalton: Yap, good point.

test/test.js
```diff
@@ -13064,6 +13064,29 @@
...
+      strictEqual(_.replaceAll('a/b/c', '/', '//'), 'a//b//c');
+    });
+
+    test('should coerce `pattern` and `replacement` to a string', 1, function() {
+      strictEqual(_.replaceAll('123456', 3, '4'), '124456');
+    });
+
+    test('should coerce `replacement` to a string', 1, function() {
+      strictEqual(_.replaceAll('123', '2', { 'toString': _.constant('4') }), '143');
+    });
```
megawac: could you add a test for a non global regexp? afterwards I'm +1

megawac: also I suppose that case should be discussed

jdalton: The param specifies it as a string, I'd assume if someone wanted regexes they'd use `String#replace`.

megawac: I was mentioning it for chain cases

jdalton: Still not sure I get what regexes have to do with it.

lodash.src.js
jdalton: ping @mcharytoniuk.

mcharytoniuk: Right, I'll make a commit in a second.

mcharytoniuk: Sorry for a delay.

Pure JavaScript gives output like that:
```JavaScript
"test".replace("", "/");
// => "/test"
```

So to be consistent `replaceAll` should work like this, which is strange IMO:
```JavaScript
_.replaceAll("test", "", "/");
// => "/t/e/s/t"
```

jdalton:
Thanks for this @mcharytoniuk! I've been keeping this PR around for a bit to see if I'd dig it anymore but haven't really gotten excited about it. I'm going to close it until demand gets a little stronger and then cherry-pick this in (so you get credit) at that time.

sbryfcz:
Maybe bring this back since more and more string functions are getting added? Would love this in lodash. Seems like I'm always having to add this function to my project's utility libraries.

catalint:
+1

ellis:
I would love to see this in lodash too!  I regularly need a simple replaceAll() function like this.

river-jade:
I'd love to see this. replaceAll seems to be far more useful in my experience than replace (first). The built-in version works for most cases, but it's clunky imo.

nevir:
This would be particularly helpful.  It's something that comes up relatively often, and is likely to reduce errors (such as people forgetting to `_.escapeRegExp`)

skfd:
That would be a tasty feature. People periodically come to [ask for it](https://github.com/lodash/lodash/issues?utf8=%E2%9C%93&q=replaceall) now and then.

PR #1126, opened 2015-04-12
Sort by all iteratees

megawac:
(I know this is the wrong file but I'm on a different computer :p)

Anyway @jdalton how do you feel about `sortByAll` and `sortByOrder` accepting more than just properties. I have a couple use cases for this namely multi column sorts with different importance factors.

As a bonus this should have deep support yeah?!!!

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -2842,19 +2842,19 @@
      * @param {boolean[]} orders The sort orders of `props`.
      * @returns {Array} Returns the new sorted array.
      */
-    function baseSortByOrder(collection, comparitors, orders) {
+    function baseSortByOrder(collection, comparators, orders) {
       var callback = getCallback();
 
-      comparitors = arrayMap(comparitors, function(comparitor) {
+      comparators = arrayMap(comparators, function(comparitor) {
         return callback(comparitor);
       });
 
       var result = baseMap(collection, function(value, index) {
-        var length = comparitors.length,
+        var length = comparators.length,
             criteria = Array(length);
 
         while (length--) {
-          criteria[length] = comparitors[length](value);
+          criteria[length] = comparators[length](value);
         }
         return { 'criteria': criteria, 'index': index, 'value': value };
       });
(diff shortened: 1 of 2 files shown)


jdalton:
Can you rebase this.

megawac:
Yeah was more looking for thoughts before I fixed this up tmw morn

jdalton:
I've known iteratees where coming :)

lodash.js
```diff
@@ -2568,17 +2568,21 @@
...
      * @returns {Array} Returns the new sorted array.
      */
-    function baseSortByOrder(collection, props, orders) {
+    function baseSortByOrder(collection, comparitors, orders) {
       var index = -1,
           length = collection.length,
           result = isLength(length) ? Array(length) : [];
 
+      comparitors = arrayMap(comparitors, function(comparitor) {
+        return getCallback(comparitor, null, 1);
```
jdalton: `getCallback` can be hoisted out of `arrayMap` as `callback = getCallback()`.

jdalton: Can you add a unit test for these.

lodash.js
```diff
@@ -2568,17 +2568,21 @@
...
+      });
+
       baseEach(collection, function(value) {
         var length = props.length,
             criteria = Array(length);
 
         while (length--) {
-          criteria[length] = value == null ? undefined : value[props[length]];
+          criteria[length] = comparitors[length](value);
         }
```
jdalton: The one thing that bothers me is there's no `this` binding. It wouldn't be possible. This makes it a bit awkward compared to `_.sortBy`.

megawac: Yep, that was the other thing - should context param be added to
sorybyorder and sortByAll?

jdalton: > Yep, that was the other thing - should context param be added

The thing is both methods can accept multiple prop names (potential iteratees) so the binding would apply to all of them? That gets messy. I think that using `_.bind` in this case would be better.

megawac: Works for me

lodash.js
```diff
@@ -2568,17 +2568,21 @@
...
+    function baseSortByOrder(collection, comparitors, orders) {
       var index = -1,
           length = collection.length,
           result = isLength(length) ? Array(length) : [];
 
+      comparitors = arrayMap(comparitors, function(comparitor) {
+        return getCallback(comparitor, null, 1);
+      });
+
       baseEach(collection, function(value) {
```
megawac: @jdalton any reason you use `baseEach` instead of `baseMap` here?

lodash.js
megawac: Will do

megawac: I guess the `1` doesn't matter here as theres no context

lodash.js
jdalton: Nothing that can't be tweaked. `baseMap` would work.

megawac:
Going to write up some tests and docs now - this implementation look good?

jdalton:
Yap!

PR #1130, opened 2015-04-14
Remove length variable and additional if-else statement in `_.last`

arthurkhu:
We don't need additional if-else statement and variable.
Here is a benchmark, this way is good too: http://jsperf.com/lodash-f-last

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -5309,8 +5309,7 @@
      * // => 3
      */
     function last(array) {
-      var length = array ? array.length : 0;
-      return length ? array[length - 1] : undefined;
+      return array ? array[array.length - 1] : undefined;
     }
 
     /**


d10:
The current behaviour is intentional because lodash tries to avoid throwing errors. By using the current behaviour, errors will not be thrown if say `array` is `undefined`.

arthurkhu:
@d10 current version and my version return the same result if array is undefined. 
In current version `var length = array ? array.length : 0;` — If array is false length is `0`. 
Next line: `length ? array[length - 1] : undefined;` If length is false (0 is false) then return `undefined`. 
My version: `return array ? array[array.length - 1] : undefined;`. If array is false then return `undefined`. Can I have an example, what you mean?

d10:
Try running `_.last(undefined)`.

jridgewell:
See http://jsperf.com/lodash-f-last/2

There's a deopt when grabbing the last item in an empty array with your implementation.

arthurkhu:
@d10 result is `undefined` in current version and my version. https://jsfiddle.net/z4vuj2a9/

d10:
Facepalm! @arthurkhu you're right, my brain is dead already.

arthurkhu:
@d10 Don't worry about it. If it's possible, please, reopen pull request, we can see, what other contributors think

jridgewell:
> If it's possible, please, reopen pull request

[See above](https://github.com/lodash/lodash/pull/1130#issuecomment-92879221). You're prioritizing performance over correctness, and it kills a common case.

jdalton:
Hi @arthurkhu! The others have expressed the gist of why I wouldn't do this. The method should accept falsey values and we want to ensure if an array is passed and is empty that it doesn't access the `-1` property. The performance gain is insignificant (millions compared to more millions of operations a second) so there's no clear win to changing the behavior.

arthurkhu:
Hi @jridgewell @jdalton. My bad with read out of bounds. I can fix it with `return (array && array.length > 0) ? array[array.length - 1] : undefined;` and it's also faster, but of course, it's microoptimization.

jdalton:
Naw, it's ok the way it is.

PR #1143, opened 2015-04-19
Added `_.cut` and appropriate tests

rheh:
Useful wrapper function to cycle the arrays order.  Use-case; for use in a simple caesar cipher.


jdalton:
Hi @rheh! Thanks but I think this is more appropriate as a mixin.

PR #1144, opened 2015-04-19
Added two missing semicolons

rheh:
Trivial code change based on jshint feedback.

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -3464,7 +3464,7 @@
           return index > -1 ? collection[index] : undefined;
         }
         return baseFind(collection, predicate, eachFunc);
-      }
+      };
     }
 
     /**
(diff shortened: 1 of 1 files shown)


jdalton:
I know it may seem silly for two semicolons but I figure while you're contributing getting [our CLA](http://dojofoundation.org/about/claForm) signed would clear the way for future contributions.

jdalton:
Can you also share the linting config you used. If it's not to chatty I may be interested in incorporating it into the repo.

rheh:
I used the default https://github.com/jshint/jshint/blob/master/examples/.jshintrc - I think you'll find it too verbose to start with.  The default config above generate a list of 492 'errors'.  This includes things like 'defined but never used.', "Expected '===' and instead saw '=='." and 'xxx' used out of scope.

I can submit a patch to fix some of these; but, as with everything some are subjective and you may want to tweak the setting jshint uses to negate some of the others.

PR #1150, opened 2015-04-21
Add .npmignore

### LINKED ISSUE #1149: Lots of unused files
It looks like lodash includes lots of files in the npm distro that don't appear to be used.

```sh
$ npm install lodash
lodash@3.7.0 node_modules/lodash
$ find node_modules/lodash | wc -l
     413
```

It doesn't look like these files are required by `index.js` so I was just wondering why they are distributed?

AaronO:
Greatly reduces npm tarball size from `3.6mb` to `450kb` and `134` files to `7` (Without losing any functionality). It simply blacklists dev related files when publishing to `npm`.

Fixes #1149

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,8 @@
+/doc/
+/perf/
+/vendor/
+/test/
+.*
+lodash.js
+lodash.min.js
+


jdalton:
Thanks, but npm package isn't based on the master branch but the [npm branch](https://github.com/lodash/lodash/tree/npm).

Issue #1072, opened 2015-03-24
_.and, _.or, and _.not for combining predicates?


aearly:
Apologies if this has been suggested before, it's practically impossible to search for.

What would you think about adding `_.and`, `_.or`, and `_.not` combinators that allow you to combine predicates?  With these functions, you could do things like:

```js
var onlyNonEmptyArrays = _.filter(coll, _.and(Array.isArray, _.not(_.isEmpty));

var isCollection = _.or(_.isArray, _.isPlainObject, _.isArguments);
```

jdalton:
The names `_.and`, `_.or`, and `_.not` are used for lots of things. To create these 3 methods we would use  a helper function. I think the utility is in that function. Would you like to brain storm usage and argument signature of that?

aearly:
A single helper method might not be possible because the "and" and "or" versions are *n*-ary, while "not" is unary.

I see that lodash-contrib adds a `_.not` function that takes in a boolean, not a predicate. (`_.not(true) === false`)  The same pattern could be expanded to `_.and` and `_.or` functions that take in booleans.  (`_.or(true, false) === true`) Maybe the helper function would be a function that combines predicates in a certain way?

```js
// either this is what _.or is, or you have to define it yourself
function or(a, b) { return a || b; }

// _.combinePredicates could take in an array or be variadic
var isCollection = _.combinePredicates(or, [_.isArray, _.isPlainObject, _.isArguments]);
```

The implementation of this would just be a fairly basic `reduce` operation.

This doesn't work for the "not" combinator.  Perhaps it could be called something like `_.inverse`?

jdalton:
Hmm, let's have this simmer a bit. If you want, create a mixin lib that tackled this to proof out various APIs.

Delapouite:
Nice proposal @aearly 

In Ramda, these functions are part of the logic category : http://ramdajs.com/docs/#not

jdalton:
@Delapouite You're mistaken. The they have the same name, like those [in lodash-contrib](https://github.com/TheNodeILs/lodash-contrib/blob/v241.4.15/_.util.operators.js#L114-L116) but they are not what the original methods proposed.

StreetStrider:
The proposed `not` I call  [`negate`](https://github.com/StreetStrider/aux.js/blob/3912116fa71d016c16db56304fc06e7733d36990/fn/negate.js). `and`, `or` are like compose but for predicates. Might be interesting. But this names are become ambiguous outside of the «functools context».

[...]

PR #1152, opened 2015-04-21
Add `_.isAll` and `_.isAny`

jlmorgan:
* Added private `createConditional` which creates the `_.isAll` and `_.isAny` functions.
* Added associated tests.

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -3378,6 +3378,45 @@
     }
 
     /**
+     * Creates a `_.isAll` or `_.isAny` function.
+     *
+     * @private
+     * @param {boolean} isAll Specify all or any must be true.
+     * @returns {Function} Returns the new conditional function.
+     */
+    function createConditional(isAll) {
+      return function () {
+        var length = arguments.length;
+        if (!length) {
+          return function() { return arguments[0]; };
+        }
+        var wrapper,
+            index = length,
+            leftIndex = 0,
+            funcs = Array(length);
+
+        while (index--) {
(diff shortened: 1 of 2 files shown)


jdalton:
Hi @jlmorgan! Thanks for this but I'm not wanting to add these at the moment. I wonder if there's a lower level utility underneath these that could be used to compose these.

jlmorgan:
Forgive my ignorance, but I am unable to see which of the existing methods would provide `&&`ing or `||`ing the results of each method call together for a singular value.

StreetStrider:
@jlmorgan, please, tell, which functions you want:
1. functions act like wrappers around `||`, `&&` operators
2. functions act like conjunction and disjunction for predicates
?

StreetStrider:
May be related to #1072.

jlmorgan:
@StreetStrider : actually, either of those would be fine. Just something to take multiple boolean functions to make a new semantic one that represents those things.

StreetStrider:
You should take a look at discussion above. Also, maybe you'll find interesting functions I've implemented:
`negate` (as in LoDash), `conj` and `disj` (as `isAll` & `isAny`): https://github.com/StreetStrider/aux.js/commit/02285f9095f07aaa113e1bd6319c5745b55342c8
This functions implement case (2).

As for case (1), take a look at [Ramda](http://ramdajs.com/docs/)'s logic group of functions. Ramda also implements (2) too. There're cases when you can simply reduce this stuff to Array's `some` & `any` (but in loss of laziness).

PR #1155, opened 2015-04-21
Adding meteor support

smmoosavi:
I create files Are needed for meteor package.
Now you must register in [meteor](https://www.meteor.com/account-settings) site with `lodash` username.
Then run following command in lodash directory
```bash
meteor publish --create
```
[more information](https://atmospherejs.com/i/publishing)

@dandv Can you see this pull request as Meteor expert?

### DIFF
--- a/package.js
+++ b/package.js
@@ -1,0 +1,17 @@
+Package.describe({
+  name: 'lodash:lodash',
+  version: '3.7.0',
+  summary: 'A JavaScript utility library delivering consistency, modularity, performance, & extras. https://lodash.com',
+  git: 'https://github.com/lodash/lodash.git',
+  documentation: 'README.md'
+});
+
+Package.onUse(function(api) {
+  api.versionsFrom('METEOR@1.0');
+  api.addFiles('lodash.js');
+  api.addFiles('export.js');
+
+  api.export('lodash');
+  api.export('_');
+});
+
--- a/export.js
+++ b/export.js
@@ -1,0 +1,3 @@
+// For meteor package
+lodash = this._;
+_ = lodash;


package.js
```diff
@@ -0,0 +1,17 @@
+Package.describe({
+  name: 'lodash:lodash',
+  version: '3.7.0',
```
megawac: Can this be `require('./package.json').version`?

smmoosavi: No as I know.

jdalton: @JamesMGreene any tips?

jdalton: I noticed the `underscore` meteor package is maintained by the [mdg](https://atmospherejs.com/mdg) org and so isn't prefixed by a username unlike the [five lodash meteor packages](https://atmospherejs.com/mdg?q=lodash). Is there another way? \@glasser

I wonder if there's a way to get the official package supported in much the same way as `underscore` or

JamesMGreene: Pretty confident this should work, though I've never tried it before:

```js
Package.describe({
  name: 'lodash:lodash',
  version: Npm.require('./package.json').version,
  /* ... */
});
```

jdalton:
I think we should hold off on handling this ourselves because the install experience would look like:
```bash
meteor add lodash:lodash
```
vs
```bash
meteor add underscore
```

I don't think the lodash experience should be second rate compared to the Underscore install experience. Also mdg packages seem to get better exposure. Here's the [underscore package](https://github.com/meteor/meteor/tree/devel/packages/underscore)  for reference (it can't use npm because the meteor version is modified).

smmoosavi:
@JamesMGreene Your suggestion is not working:

```js
Package.describe({
  name: 'lodash:lodash',
  version: Npm.require('./package.json').version,
/* ... */
});
```

```
=> Started proxy.                             
=> Errors prevented startup:                  
   
   While reading package from `/path/to/project/packages/lodash:lodash`:
   package.js:3:16: can't find npm module './package.json'. In package.js, Npm.require can only find built-in modules.
   package.js:3:41: Cannot read property 'version' of undefined
   
=> Your application has errors. Waiting for file change.
```

smmoosavi:
@jdalton [bootstrap](github.com/twbs/bootstrap/) publish meteor package with [twbs](https://atmospherejs.com/twbs/bootstrap) prefix.

JamesMGreene:
> ```
=> Started proxy.                             
=> Errors prevented startup:                  
>
>   While reading package from `/path/to/project/packages/lodash:lodash`:
   package.js:3:16: can't find npm module './package.json'. In package.js, Npm.require can only find built-in modules.
   package.js:3:41: Cannot read property 'version' of undefined
>
> => Your application has errors. Waiting for file change.
```

_\*sigh\*_ Gotta love the fun and completely unnecessary constraints that the Meteor Platform "provides" on top of Node.js. :disappointed:

PR #1157, opened 2015-04-21
Added missing constructor invocation as idendified by JSHint

rheh:
Explanation of JSHint reasoning here: https://jslinterrors.com/missing-invoking-a-constructor

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -843,7 +843,7 @@
     var MAX_SAFE_INTEGER = Math.pow(2, 53) - 1;
 
     /** Used to store function metadata. */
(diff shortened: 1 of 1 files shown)


lodash.src.js
```diff
@@ -1971,7 +1971,7 @@
       return function(prototype) {
         if (isObject(prototype)) {
           Object.prototype = prototype;
-          var result = new Object;
+          var result = new Object();
```
sankargorthi: silly question - isn't it simpler to just `var result = {};`?

jdalton:
Thanks, but no thanks on this one.

PR #1167, opened 2015-04-27
Implement sample and shuffle via baseSample

megawac:
Related https://github.com/jashkenas/underscore/pull/2158

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -7190,6 +7190,33 @@
     }
 
     /**
+     * Create a random sample of an array of up to `n` elements 
+     *
+     * @private
+     * @param {Array} array The array to create a sample from.
+     * @param {Integer} n The number of elements to include.
+     * @returns {Array} Returns the sample of list.
+     */
+    function baseSample(array, n) {
+      var index = -1,
+          length = array.length,
+          result = Array(n);
+
+      while (++index < n) {
+        var rand = baseRandom(0, index);
+        if (index != rand) {
+          result[index] = result[rand];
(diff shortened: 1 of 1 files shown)


jdalton:
Failing tests.

lodash.src.js
```diff
@@ -7190,6 +7190,31 @@
...
+     */
+    function baseSample(array, n) {
+      var index = -1,
+          length = array.length,
+          result = Array(n);
+
+      while (++index < length) {
+        var rand = baseRandom(0, index);
+        if (rand < n) {
+          if (index != rand) {
```
jridgewell: This no longer holds, since `index` can be greater than `n` here. I think this works better (and runs faster) as two separate loops.

jdalton:
I'm not a big fan of this change. The existing implementation is fine and performance is not a concern. If you'd like to abstract the existing `sample` into ` baseSample` I'm cool with that but anything is too much for this kind of method (y'all are spending more time then it's worth on this thing).

jridgewell:
> If you'd like to abstract the existing sample into baseSample I'm cool with that, but anything else is too much for this kind of method

Why's that? It's a faster algorithm with ~5 more lines, and doesn't affect base performance.

jdalton:
> Why's that? It's a faster algorithm with ~5 more lines, and doesn't affect base performance.

Fisher-Yates shuffle is plenty fast and I know it won't behavior wonky (it shuffles well). It doesn't need to be any more clever than it already is.

PR #1177, opened 2015-04-30
Fixes #1046, Restore _.where([], function)

### LINKED ISSUE #1046: Unexpected result returned by _.where
Hey! 

Apologies if something changed in `_ 3.5` that i'm not aware of, but today i found a test failing in my project because _.where returned an unexpected result.

Please take a look at this plunkr: http://plnkr.co/edit/3vkqWrC1T1Lh9XinjdEk?p=preview

robrich:


jdalton:
Thanks @robrich! But the existing behavior is correct. If you want a `predicate` please use `_.filter`.

PR #1186, opened 2015-05-06
sortByOrder can sort properties with null values

ignu:
Without this fix, sortByOrder will just silently leave collections will null sorted properties partially sorted. 

See http://codepen.io/ignu/pen/RPPOwO?editors=001

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -301,6 +301,9 @@
       var valIsReflexive = value === value,
           othIsReflexive = other === other;
 
+      if (value === null) return 1;
+      if (other === null) return -1;
+
       if (value > other || !valIsReflexive || (value === undefined && othIsReflexive)) {
         return 1;
       }
--- a/test/test.js
+++ b/test/test.js
@@ -14415,9 +14415,9 @@
       deepEqual(actual, [3, 1, 2]);
     });
 
-    test('should move `undefined` and `NaN` values to the end', 1, function() {
-      var array = [NaN, undefined, 4, 1, undefined, 3, NaN, 2];
-      deepEqual(_.sortBy(array), [1, 2, 3, 4, undefined, undefined, NaN, NaN]);
+    test('should move `undefined`, `null` and `NaN` values to the end', 1, function() {
+      var array = [NaN, undefined, 4, 1, undefined, 3, null, NaN, null,2];
+      deepEqual(_.sortBy(array), [1, 2, 3, 4, undefined, undefined, NaN, NaN, null, null]);
     });
 
     test('should treat number values for `collection` as empty', 1, function() {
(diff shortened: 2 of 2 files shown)


jdalton:
This change affects more than `_.sortByOrder`. It touches `_.sortBy`, `_.sortByAll`, `_.sortedIndex`, & `_.lastSortedIndex` too. I'll need to kick this around a bit. In the meantime have you signed [our CLA](http://dojofoundation.org/about/claForm)?

ignu:
Cool I just signed it.

jdalton:
By moving the null checks a bit lower
```js
if (value > other || !valIsReflexive || (value === undefined && othIsReflexive)) {
  return 1;
}
if (value < other || !othIsReflexive || (other === undefined && valIsReflexive)) {
  return -1;
}
if (value === null) {
  return -1;
}
if (other === null) {
  return 1;
}
```

It would keep the nulls at the beginning and still allow sorting:
```js
_.sortBy(['b', null, NaN, 'c', undefined, null, undefined, NaN, 'a'])
// => [null, null, "a", "b", "c", undefined, undefined, NaN, NaN]
```

This aligns more closely to the sorting when comparing numbers as `null` is seen as `0`.
Regardless you'll want to run a `_.compact` on your results to remove the empties.

Could you tweak the PR to align with this and add/augment a test for the other sortBy-like methods.

ignu:
Well, I wouldn't expect `undefined` and `null` to be on different ends of the array...

In fact native sort seems to put `null` at the end too.

    [null, 3, 9].sort() // [3, 9, null]

despite `3 < null // false`

(In my case, I'm ordering users by location, from which the API occasionally returns `null`, so `_.compact` isn't a solution.)

gibson042:
[Default native comparison](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-sortcompare) string-casts every value except `undefined` (which follows everything), so `13` precedes `7` and `null` ends up between `"hard place"` and `"rock"`.

jdalton:
Ok so I can move `null` to the end say behind `undefined` and `NaN`. I have it working locally. That said it's a pain because `null` compares as `0` so where as `undefined` compares as `NaN` the `null` throws a wrench in otherwise simple comparisons. It basically means adding `null` checks in like 3 places.

jdalton:
I'll merge this and then pave parts with the fix. This way you get the cred for initiating it.

mathieumg:
Just to make sure: after this change and when sorting an array of objects using their `myProperty` property  which is a `Number`, should I expect an object that doesn't have `myProperty` (i.e. it is `undefined` for all intents and purposes) to come before `0` or after ̀`Math.Infinity`? 

I think my actual use case is not uncommon, I sort in descending order by a number property and objects who don't have the property in question (most of them) are considered to have a value of 0 for that property, implicitly. They need to go at the end (since it's descending) of everything, and not before the high values. I got it working using a custom iteratee but if this PR means I won't need to do that anymore that's even better! Thank you.

jdalton:
This will sort the following values at the end and in the order of `null`, `undefined`, `NaN`.
```js
_.sortBy([Infinity, null, -1, NaN, 2, undefined, 3]);
// => [-1, 1, 2, 3, Infinity, null, undefined, NaN]
```

mathieumg:
So `_.sortBy([{a: 1}, {a: 0}, {}]);` would give `[{a: 0}, {a: 1}, {}]` and not `[{}, {a: 0}, {a: 1}]`?

jdalton:
If you mean `_.sortBy([{a: 1}, {a: 0}, {}], 'a');` then yes.

mathieumg:
Yes I posted a bit too quickly and didn't edit quickly enough it seems. Thanks for the clarification, I guess I'll stick with my custom iteratee then!

jdalton:
> Thanks for the clarification, I guess I'll stick with my custom iteratee then!

No problem. The idea is that it's easier to clip an array's `length` than it is to `shift` them off the front.

PR #1195, opened 2015-05-09
Addition of _.next, _.previous and _.placeInRange

mrjamesriley:
While working on a recent project, I found the addition of the following three simple methods could prove both helpful to a large number of lodash users, and that they feel inline with the existing methods available:

`_.next` - Given an array and one of it's elements, return the next element.
`_.previous` - Similar, but returning the previous element of the array.

`_.placeInRange` - Ensures that a value falls within a range, returning the start or end value if the value provided happens to exceed or fall short of the range.

Have submitted this as a pull request, but it's more an example implementation to trigger discussion. My most recent use case is building a dynamic Grid composed of resizable columns, where I'm wanting to enforce min/max resize limits, and easily obtain the preceding or subsequent column.

(random note: have been inspired to contribute towards lodash after listening to the awesome JSJabber episode with John-David Dalton - love the unrelenting push for improvement).

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -928,8 +928,8 @@
      * `isFinite` `isFunction`, `isMatch`, `isNative`, `isNaN`, `isNull`, `isNumber`,
      * `isObject`, `isPlainObject`, `isRegExp`, `isString`, `isUndefined`,
      * `isTypedArray`, `join`, `kebabCase`, `last`, `lastIndexOf`, `lt`, `lte`,
-     * `max`, `min`, `noConflict`, `noop`, `now`, `pad`, `padLeft`, `padRight`,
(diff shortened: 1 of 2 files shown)


jdalton:
Hi @mrjamesriley! Thanks for these. I've seen something along the lines of `_.next` and `_.previous` around. The names could be tweaked but I think these would work better as a group of mixins. I'll keep an eye on the popularity of the request though and reevaluate as others show interest.

jdalton:
If you haven't signed [our CLA](http://dojofoundation.org/about/claForm) you could still do that for future contributions.

mrjamesriley:
Sure thing - and thanks for the prompt response!
Have signed the CLA `2015-05-09 08:11:27`. 
Goes without saying, but keep up the awesome work.

PR #1209, opened 2015-05-15
Improve remove performance

CodingFabian:
When profiling our application, I saw a significant amount of time spent in our calls to _.remove().
When looking at the current code, I saw that it builds an array of indices to remove from the source array, and then uses basePullAt to remove the elements.

basePullAt for some reason allows to use anything as index, which is why it uses parseFloat().

So I decided to inline a specialized version which knows that the passed arguments are valid integer indices.

PS: I filled out the CLA (submitted a CLA on 2015-05-15 12:50:41)
PPS: I could not figure out if I need to change the es branch as well.

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -5457,7 +5457,12 @@
           indexes.push(index);
         }
       }
-      basePullAt(array, indexes);
+
+      length = indexes.length;
+      while (length--) {
+        index = indexes[length];
+        splice.call(array, index, 1);
+      }
       return result;
     }
 


CodingFabian:
If you are interested in some rough estimate:
http://jsperf.com/parsefloatornot
it shows that the unnecessary parseFloat is about 150x slower.

davidkpiano:
The reason that `parseFloat()` is in `basePullAt()` is because it is necessary for `pullAt()`:

```javascript
_.pullAt([1,2,3], '1', '2'); // uses parseFloat() to parse string indexes
// => [2, 3]
```

CodingFabian:
ok, thats what I assumed, but at leas in remove it is known that they are of correct type and the parsing should be skipped.

jdalton:
> The reason that `parseFloat()` is in `basePullAt()` is because it is necessary for `pullAt()`:

It's not really about numeric string values its more for catching things like `false` which when coerced to a number are `0` but when passed through `parseFloat` are `NaN` and so skipped.

jdalton:
I think this is better tackled in the `isIndex` function. I can add a bit heavier check there for non-number values. This would keep the common case speedy and remove the `parseFloat` use.

jridgewell:
The jsperf shows `parseFloat()` is slow in isolation, but how does it compare to the rest of the method? I'd figure the `splice`ing would be the slowdown, not parsing.

jdalton:
I get what @jridgewell is driving at and I agree. That said moving `parseFloat` into `isIndex` makes it a stronger function so it's a win regardless of the `basePullAt` thing.

CodingFabian:
well yeah of course is the array splice slow, i just thought that the parsefloat is way too unnecessary for the case we on our own build the index array which we know is int.
the change by @jdalton addresses this but keeps the reuse. 
thanks for doing that.

Issue #1229, opened 2015-05-25
set() fails to set keys/value with certain inputs


chrishiestand:
Take for example:

    var hash = {};
    _Lo.set(hash, ['a','1string'], 'myvalue');

I expect this to yield: `{ "a" : { "1string" : "myvalue" } }` but instead it yields: `{ "a" : [] }`

I narrowed the issue down to isIndex() returning true when "1string" is the input. I am filing a pull request which contains failing tests but will pass when this problem is fixed. Apologies I do not know javascript's weird edge cases well enough to quickly fix the problem; I'm hoping that someone else sees a simple solution.

jdalton:
Moved to #1230.

PR #1230, opened 2015-05-25
add tests for isIndex and set()

chrishiestand:
for the case of an input string like "1string"
  currently these two tests fail, and they should not

See #1229

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -829,8 +829,9 @@
       }
     });
 
-    test('should return `false` for non-indexes', 4, function() {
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @chrishiestand! Have you signed [our CLA](http://dojofoundation.org/about/claForm)?

chrishiestand:
Yup, I did just before sending this PR

jdalton:
Thanks!

chrishiestand:
Thanks for the quick fix!

PR #1242, opened 2015-05-28
feat(useWith): add useWith function

steelsojka:


jdalton:
Have you signed [our CLA](http://dojofoundation.org/about/claForm) yet?

steelsojka:
Yep

steelsojka:
I also used the same name as Ramda uses for this functionality, so if you would prefer a different name let me know and I'll update the request.

benjycui:
@steelsojka Could you provide some usage scenarios? thanks!

lodash.src.js
```diff
@@ -7337,6 +7337,46 @@
...
+     * fn(1, 2); // => [1, 4]
+     * fn(5, 10); // => [25, 20]
+     */
+    var useWith = restParam(function(callback, transforms) {
+      return function() {
+        var length = transforms.length,
+            index = -1;
+
+        while (++index < length) {
+          arguments[index] = transforms[index].call(this, arguments[index]);
```
d10: Not too sure here, but I think overwriting `arguments` could cause a deopt. \@jdalton?

jdalton: That should be OK as there aren't any formal params. The issue would come from passing `arguments` to `.apply` below. But I won't nit pick that for the PR.

steelsojka: @jdalton I don't think we can get around the `apply` since the there can be `n` number of arguments.

@d10 I thought about that. The other solution was convert arguments to an array and modify the new list of arguments, but I don't know which one would cause more overhead.

steelsojka: @benjycui I added a more advanced scenario that might spark more use cases.

benjycui: @steelsojka Thank you!
I come up with another solution to convert `arguments` and test them on http://jsperf.com . And I thought those test case can help you to make a decision. http://jsperf.com/the-best-solution-to-map-the-arguments-of-function

benjycui:
@steelsojka I am not sure whether `mapWith` is better than `useWith`.

lodash.src.js
steelsojka: @benjycui looks like creating an array with the length of the transforms seems to be the fastest. I'll update the request. Nice job!

benjycui: How about this? It just a pseudocode, I have not run this code.
```js
var useWith = restParam(function(callback, transforms) {
  return function() {
    var al = arguments.length,
        tl = transforms.length,
        l = Math.min(al, tl),
        res = Array(l),
        index = -1;

    while (++index < l) {
      res[index] = transforms[index].call(this, arguments[index]);
    }
    res.concat([].splice.call(arguments, l));

    return callback.apply(this, arguments);
  }
});
```

P.S. I think you had better add a test case for which `transfroms.length > arguments.length`.

jdalton: I'd prefer you focus on basic implementation. We can tweak perf stuff at a later time.

steelsojka: @benjycui I think that implementation would cause some more overhead than the original implementation. I think the base implementation (with a minor tweak) will be fine performance wise for now. All other scenarios I have tried cause more overhead than the original.

steelsojka:
@benjycui I like `mapWith` better. :+1:

StreetStrider:
@benjycui, @steelsojka, `mapWith` can create unwanted associations with collections.

steelsojka:
@StreetStrider good point. I'll change it back. Does anyone have any other suggestions? `useWith`, `mapArgs`, `composeArgs`, `composeWith`?

StreetStrider:
@steelsojka, I was thinking about it with `proj` and `transform` in first place. But your names is good too. The case of this function is to transform or project or constraint arguments. It also directly related to functions and arguments. I think the name should refer in some form to such things.

steelsojka:
@StreetStrider What about `transformArgs`? I think that is the most concise and easiest to understand what it is suppose to do.

StreetStrider:
@steelsojka, sounds reasonable.

jdalton:
We have `rearg` what about `modArgs`?

steelsojka:
@jdalton Sounds good to me.

steelsojka:
The build seemed to error out for some reason. Can it be restarted?

d10:
I’ve restarted one of the builds, AFAIK it seems to be an nvm issue.

jdalton:
Ya looks like a travis-ci hiccup, no worries.

jdalton:
Ok, so @phated suggested `xargs`. Bikeshedding continued...

benjycui:
We can hold a poll. :-)

PR #1251, opened 2015-05-30
_.ary can create wrappers with exact arity.

wbyoung:
This pull request allows `_.ary` to take an extra argument, which when true, creates a wrapper function that has the exact right arity.

```js
_.ary(fn, 5).length === 5
```

There's are a couple of JS libraries where function arity is important (Express, for instance).

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -7404,19 +7404,34 @@
      * @category Function
      * @param {Function} func The function to cap arguments for.
      * @param {number} [n=func.length] The arity cap.
+     * @param {boolean} [exact] Create a function with matching arity.
      * @param- {Object} [guard] Enables use as a callback for functions like `_.map`.
      * @returns {Function} Returns the new function.
      * @example
      *
      * _.map(['6', '8', '10'], _.ary(parseInt, 1));
      * // => [6, 8, 10]
+     *
+     * _.ary(parseInt, 1, true).length;
+     * // => 1
      */
     function ary(func, n, guard) {
+      var exact = false;
       if (guard && isIterateeCall(func, n, guard)) {
         n = undefined;
       }
+      else {
+        exact = guard;
+      }
       n = (func && n == null) ? func.length : nativeMax(+n || 0, 0);
-      return createWrapper(func, ARY_FLAG, undefined, undefined, undefined, undefined, n);
+      var wrapper = createWrapper(func, ARY_FLAG, undefined, undefined, undefined, undefined, n);
+      if (exact) {
+        var plist = [];
+        for (var i = 0; i < n; i++) { plist.push('p' + i); }
+        var params = plist.join(',');
+        wrapper = Function('fn', 'return function(' + params + ') { return fn.apply(this, arguments); }')(wrapper);
+      }
+      return wrapper;
     }
 
     /**
--- a/test/test.js
+++ b/test/test.js
@@ -1016,6 +1016,10 @@
         skipTest();
       }
     });
+
+    test('should produce a function with matching arity', 1, function() {
+      strictEqual(_.ary(fn, 3, true).length, 3);
+    });
   }());
 
   /*--------------------------------------------------------------------------*/


megawac:
I Personally I think it'd be better to expose the `metaMap` and doing this in an extension

lodash.src.js
```diff
@@ -3592,6 +3594,15 @@
         }
         return fn.apply(thisBinding, args);
       }
+
+      if (isAryExact) {
+        var plist = [];
+        for (var i = 0; i < ary; i++) { plist.push('p' + i); }
+        var params = plist.join(',');
+        var fn = wrapper;
+        eval('wrapper = function(' + params + ') { return fn.apply(this, arguments); }');
```
d10: Using `eval` won’t work in CSP enabled environments, like Chrome extensions. We won’t be able to consistently set the function arity because of this.

wbyoung: @d10 true. I was not considering such environments. As it's an opt-in feature that only occurs when passing the third argument as true, `_.ary(fn, n, true)`, it seems that the impact to users in this environment would be small. I don't see much chance of them being confused if attempting to use this feature & having it fail either.

I believe this is the only way to handle arbitrary airity. To handle arity up to a fixed amount, one could create pre-defined wrappers for argument counts of 0-20, for instance, and fail for functions with more than 20 arguments. Only handling functions up to a certain length seems limiting, dangerous, and cumbersome.

Anyway, thanks for the observation. I still do believe this feature to be useful, but ultimately I will leave it up to whatever standard process you usually follow when considering adding features to the project.

jdalton: Because this is opt-in I don't mind the CSP gotcha. It's similar to `_.template` use. However I would move this into the `_.ary` function directly and not try to incorporate it into the wrapper helpers. I'd also switch `eval` to `Function(..)` use.

wbyoung: @jdalton moved it into `_.ary` in 30b3083. I should have done that to begin with. I was trying to get a grasp on the code base and chose the hard route. ;)

As for changing to `Function(..)` I'm not sure I understand what you mean. The function body needs to reference values in local scope, so if you're suggesting something like `new Function(param1, param2, body)`, I don't understand how to make that work.

doc/README.md
```diff
@@ -3335,6 +3335,7 @@ additional arguments.
 #### Arguments
 1. `func` *(Function)*: The function to cap arguments for.
 2. `[n=func.length]` *(number)*: The arity cap.
+3. `[exact=false]` *(boolean)*: Create a wrapper with the exact arity.
```
jdalton: You don't have to update the doc/README.md, it's generated from the lodash.src.js at a later time.

doc/README.md
```diff
@@ -3343,6 +3344,9 @@ additional arguments.
 ```js
 _.map(['6', '8', '10'], _.ary(parseInt, 1));
 // => [6, 8, 10]
+
+_.ary(parseInt, 1, true).length;
+// => 1
 ```
```
jdalton: You can move the doc additions to the lodash.src.js though.

lodash.src.js
```diff
@@ -3592,6 +3592,7 @@
         }
         return fn.apply(thisBinding, args);
       }
+
```
jdalton: Rogue new line.

lodash.src.js
jdalton: I mean you can do:
```js
var wrapper = Function('func','return function(' + params + '){return func.call(this,' + params + ')}')(func);
```

wbyoung: @jdalton ok, I understand. I did this and opted to simply re-wrap the wrapper generated by `createWrapper` and use `apply` rather than `call` with `params`. I felt this simplified the code, but am happy to change it.

I also addressed the other issues & rebased the changes into a single commit.

tejasmanohar:
+1

jdalton:
@wbyoung Can you rebase this PR to clean it up. Also, have you signed [our CLA](http://dojofoundation.org/about/claForm)?

wbyoung:
@jdalton rebased & signed.

lodash.src.js
```diff
@@ -7404,19 +7404,34 @@
...
+        exact = guard;
+      }
       n = (func && n == null) ? func.length : nativeMax(+n || 0, 0);
-      return createWrapper(func, ARY_FLAG, undefined, undefined, undefined, undefined, n);
+      var wrapper = createWrapper(func, ARY_FLAG, undefined, undefined, undefined, undefined, n);
+      if (exact) {
+        var plist = [];
+        for (var i = 0; i < n; i++) { plist.push('p' + i); }
+        var params = plist.join(',');
+        wrapper = Function('fn', 'return function(' + params + ') { return fn.apply(this, arguments); }')(wrapper);
```
megawac: Its probably worth caching these strings, I imagine they are expensive to create

jdalton:
After some time to think, we've made it this long in node land without this I don' think we need to start attempting this support now. Relying on `.length` of functions will only get trickier with ES6 because it's configurable and because things like default params affect them.

megawac:
How about exposing the metaMap so people can mix this in @jdd?

On Fri, Jun 19, 2015 at 9:38 PM, John-David Dalton <notifications@github.com

jdalton:
> How about exposing the metaMap so people can mix this in

I don't want to expose `metaMap`. It's a nice thing that can melt away if needed.

wbyoung:
@jdalton thanks for considering it.

PR #1254, opened 2015-06-01
More characters recognized by _.deburr

szarsti:
Add latin extended-a unicode characters block. This include characters used in eastern European languages.


jdalton:
Hi @szarsti! Thanks for the PR. The `_.deburr` method certainly has room to expand. That said, this is a substantial expansion. I'd like to hold off until there's a greater demand for it. This looks like something that could grow into it's own specialized library. In the meantime feel free to expand `_.deburr` through a mixin.

PR #1259, opened 2015-06-02
Adding lodash.interleave

Yomguithereal:
The aim of this PR is to add the `interleave` function to lodash. The aim of this method, whose name comes directly from Clojure, is to "interleave" the given arrays' items likewise:

```js
_.interleave([1, 2, 3], [4, 5, 6]);
// => [1, 4, 2, 5, 3, 6]

// using more than two arrays
_.interleave([1, 2], [3, 4], [5, 6]);
// => [1, 3, 5, 2, 4, 6]

// the shortest array will stop interleave
_.interleave([1, 2], [3, 4, 5]);
// => [1, 3, 2, 4]
```

**Couple of notes:**

* If this method is already existing in lodash but under another name which I wouldn't have found in the docs, then sorry for the useless PR :-).
* This PR is merely a discussion opener about `interleave`. I don't know whether lodash's community would need this method.
* A similar result to this function might be obtained through `_.flatten(_.zip(array1, array2))` but I guess it would be less performant.
* A similar result to `_.zip(array1, array2)` might be obtained through `_.chunk(_.interleave(array1, array2), 2)`.
* I use `compact` to filter the given arrays but I am not sure this is the way it is done in the library's internal. What are usually the means of making the array methods accept falsey and invalid values?

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -5067,6 +5067,54 @@
     }
 
     /**
+     * Creates an array from given arrays and consisting of all the first elements from
+     * said arrays then the second elements etc.
+     *
+     * @static
+     * @memberOf _
+     * @category Array
+     * @param {...Array} [arrays] The arrays to interleave.
+     * @returns {Array} Returns the interleaved array.
+     * @example
+     *
+     * _.interleave([1, 2, 3], [4, 5, 6]);
+     * // => [1, 4, 2, 5, 3, 6]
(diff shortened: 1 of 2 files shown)


jdalton:
Hi @Yomguithereal! I'm not sure I see the broad appeal of a method like this so I think for now this would be great left as a mixin.

Issue #815, opened 2014-12-04
Move website to gh-pages


knownasilya:
Moving this suggestion from #791. 

Seems like something for @mathiasbynens (according to @jdalton).

PR #1265, opened 2015-06-07
Move lodash.com to gh-pages

InstyleVII:
I've made the changes necessary to setup gh-pages. Will need to do some work for the domain to point towards it but other than that the leg work is done. Also moved to remote-hosting github buttons and cleaned up a little bit of the code.


jdalton:
Related to #815.

jdalton:
Hi @InstyleVII! Could you squash these down to a single commit.

phated:
Instead of having this in the main repo, I think we should create a repo named lodash.github.io as github suggests at https://pages.github.com

jdalton:
I can dig that. Sorry @InstyleVII, lets move it [over there](https://github.com/lodash/lodash.github.io).

PR #1270, opened 2015-06-09
Ensure `_.get` and `_.result` return the same given object for empty paths.

Dashed:
This PR changes behaviour to consider `[]` and `''` to be empty paths, and thus `_.get` would return the same given object:

```js
const NOT_SET = {};
const obj = {foo: 'bar'};

assert(_.get(obj, [], NOT_SET) === obj); // => previously returned defaultValue for v3.9.3
assert(_.get(obj, '', NOT_SET) === obj); // => previously returned defaultValue for v3.9.3
```

This affects `_.result`:

```js
const obj = {foo: 'bar'};
assert(_.result(obj, [], NOT_SET) === obj); // => previously returned defaultValue for v3.9.3
assert(_.result(obj, '', NOT_SET) === obj); // => previously returned defaultValue for v3.9.3

// as a consequence, something like this would now work

function foo() {
    return this.bar;
}
foo.bar = 'bar';

assert(_.result(foo, '', NOT_SET) === 'bar');
assert(_.result(foo, [], NOT_SET) === 'bar');
```

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -2215,7 +2215,7 @@
       while (object != null && index < length) {
         object = toObject(object)[path[index++]];
       }
-      return (index && index == length) ? object : undefined;
+      return (index == length) ? object : undefined;
     }
 
     /**
@@ -10058,12 +10058,12 @@
      * // => 'default'
      */
     function result(object, path, defaultValue) {
-      var result = object == null ? undefined : toObject(object)[path];
+      var result = object == null ? defaultValue : toObject(object)[path];
       if (result === undefined) {
-        if (object != null && !isKey(path, object)) {
+        if (object != null && (!isKey(path, object) || !hasOwnProperty.call(object, path))) {
           path = toPath(path);
-          object = path.length == 1 ? object : baseGet(object, baseSlice(path, 0, -1));
-          result = object == null ? undefined : toObject(object)[last(path)];
+          object = path.length <= 1 ? object : baseGet(object, baseSlice(path, 0, -1));
+          result = object == null ? defaultValue : baseGet(object, path.length ? [last(path)] : []);
         }
         result = result === undefined ? defaultValue : result;
       }
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -9604,9 +9604,20 @@
      * // => 'default'
      */
     function get(object, path, defaultValue) {
-      var pathKey = isArray(path) && path.length == 0 ? undefined : path + '';
-      var result = object == null ? defaultValue : baseGet(object, toPath(path), pathKey);
-      return result === undefined ? defaultValue : result;
+      var result = (object == null || isArray(path) && path.length == 0) ? undefined : toObject(object)[path];
+      if (result === undefined) {
+        var resolved = object != null && hasOwnProperty.call(object, path);
+        if (object != null && (!isKey(path, object) || !resolved)) {
+          path = toPath(path);
+          object = path.length <= 1 ? object : baseGet(object, baseSlice(path, 0, -1));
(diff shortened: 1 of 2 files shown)


jdalton:
Hi @Dashed! Have you signed our [CLA](http://dojofoundation.org/about/claForm)?

test/test.js
```diff
@@ -13444,7 +13444,7 @@
 
     test('`_.' + methodName + '` should handle empty paths', 4, function() {
       _.each([['', ''], [[], ['']]], function(pair) {
-        strictEqual(func({}, pair[0]), undefined);
+        deepEqual(func({}, pair[0]), {});
```
jdalton: You should use `var object = {}` and do a `strictEqual` that the result is `object`.

lodash.src.js
```diff
@@ -10060,10 +10060,10 @@
     function result(object, path, defaultValue) {
       var result = object == null ? undefined : toObject(object)[path];
       if (result === undefined) {
+        path = toPath(path);
```
jdalton: Why move the `path` assignment up?

Dashed: To make this work:
```js
const obj = {};
assert(_.result(obj, '', {}) === obj);
```

Otherwise, `defaultValue` would be returned.

Dashed: Hmmm is there a reason to check `!isKey(path, object)`? It's implicitly done via `var result = object == null ? undefined : toObject(object)[path];`.

Dashed:
Signed the CLA.

lodash.src.js
```diff
@@ -10060,10 +10060,10 @@
     function result(object, path, defaultValue) {
       var result = object == null ? undefined : toObject(object)[path];
       if (result === undefined) {
-        if (object != null && !isKey(path, object)) {
+        if (object != null) {
```
jdalton: The `isKey` check is there to avoid going down the deep property path if it's a plain property access like `foo` or `bar`.

Dashed:
Interpreting empty strings is tricky:

```js
const obj = {};
_.get(obj, ''); // => returns obj

_.get({'':3}, ''); // => returns 3

// this is expected behavior
_.get(obj, '.'); // => returns obj
_.get(obj, '.....'); // => returns obj
```

@jdalton How do you want to handle empty strings?

jdalton:
An empty string is equiv to a path of `['']`

Dashed:
I refactored `_.result` a bit more. 

I really don't like the fact that `[]` and `['']` are treated to be the same; but I can work around it for my usecase.

jdalton:
After sleeping on it that is odd. Let's have `''` equal `[]` instead.

jridgewell:
Should have posted this here instead of gitter. While it's easy to see `[]` as being empty and returning the root object, `''` is a proper (uncommon) key.

Dashed:
I think we should treat `[]` as a special case; because it's being coerced to `''` and thus treated as the same case as `['']` which is totally different.

Dashed:
I noticed from docs that:

> If the resolved value is `undefined` the `defaultValue` is used in its place.

Wouldn't it make more sense to return any value at the path if the path exists? 

```js
_.get({ 'a': undefined }, 'a', 42); // return undefined instead of 42
```

I made this change in this PR; it may go into another PR.

jdalton:
Because this is a dual purpose method (meaning it gets property values or nested property values) we won't try to get too fancy. The idea is to get a property value. I wouldn't expect that to ever return the object I'm querying on though.

Dashed:
Ah. That's unfortunate.

SimonCheung1989:
I think we really need the feature that get method could return the obj itself.

PR #1276, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "lodash",
-  "version": "3.9.3",
   "main": "lodash.js",
   "ignore": [
     ".*",


jdalton:
Hey Kevin! For a trivial PR like this you don't have to but have you considered signing [our CLA](http://dojofoundation.org/about/claForm)? It would open up further contributions in the future.

kkirsche:
Signed

jdalton:
Do you know if this affects [Duo](http://duojs.org/) support?

kkirsche:
I've never heard of Duo before. My understanding though from the teams I've discussed this stuff with is that bower will be dying a slow and painful death honestly. So anything relying on Bower for this info is going to be screwed long term. Could be wrong, but that's just how I've understood the discussions

jdalton:
From my quick glance Duo seems to gleam info from several places, bower being just 1 so it should be fine.

PR #1294, opened 2015-06-20
Debounce and throttle cancel should reset lastCall timestamp

### LINKED ISSUE #1293: `throttled.cancel()` doesn't clear the timeout and cached value
Is it somehow possible to fully reset the state of the throttled function?
Maybe to add some parameter to the `cancel` method to handle this case:

```javascript
var throttled = _.throttle(function (param) {
    return param + 1;
}, 1000);

var value = throttled(0); // 1
document.body.textContent = value + '***';

// want to reset the cached value and timeout of `throttled`
// so that it would be possible to immediately call it again 
throttled.cancel();

value = throttled(1); // 1, but want to have 2
document.body.textContent += value;
```

http://jsfiddle.net/ccqgdeaa/1/

megawac:
Fixes #1293

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -15889,6 +15889,29 @@
         QUnit.start();
       }
     });
+
+    asyncTest('_.' + methodName + ' should completely reset after cancel', 3, function() {
+      if (!(isRhino && isModularize)) {
+        var callCount = 0;
+
+        var funced = func(function() {
+          return ++callCount;
+        }, 32, { 'leading': true });
+
+        strictEqual(funced(), 1);
+        funced.cancel();
+        strictEqual(funced(), 2);
+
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -15889,6 +15889,29 @@
...
+        funced.cancel();
+        strictEqual(funced(), 2);
+
+        setTimeout(function() {
+          strictEqual(callCount, 2);
+          QUnit.start();
+        }, 64);
+      }
+      else {
+        skipTest();
```
jdalton: `skipTest(3)`.

megawac: Ah, thought `skipTest` skipped the entire function

jdalton:
Thank you Graeme!

Issue #866, opened 2015-01-22
Access to the LazyWrapper


gampleman:
One of the joys of using Lo-Dash is it's [extensibility](http://eng.rightscale.com/2015/01/22/lodash-extensions.html). With Lo-Dash 3, is it possible to extend the Lazy evaluation mechanism? If not, can that be considered as a feature request?

jdalton:
The `LazyWrapper` is available as [a module](https://github.com/lodash/lodash/blob/npm/internal/LazyWrapper.js) for the npm build and indirectly available via the produced wrapper though it's not so pretty:

```js
_([]).map().__wrapped__.constructor
```

Generally I'd avoid accessing properties with double underscores as they are not intended as public API.
So at the moment we don't have a great way. We also don't directly expose `LodashWrapper`. I'm open to exposing it in a future release. If you have some extensions in mind can you kick around some ideas in a gist?

gampleman:
Sure, I'll put something up tomorrow.

gampleman:
One use case is support for other data structures that we don't want to first convert to an array:

https://gist.github.com/gampleman/bce574d3e2b79926e55a

Presumably I could do
     
     var tree = new Tree('hello', [
       new Tree('Really long leaf'),
       new Tree('Brnch', [
         ...
       ])
     ]);
     _(tree).mapTree(function(label) { return label.length;})
       .filter(function(n) { return n > 5;}).first();

calebboyd:
Is exposing `LodashWrapper` still a possibility? It might be particularly convenient if you wanted to do something akin to the following:

```javascript
import { chain } from 'lodash';
const LodashWrapper = chain(null).constructor;

export class MyLodashWrapper extends LodashWrapper {
   myLodashWrapperMethod() {

   }
}
```

PR #1295, opened 2015-06-21
feat: expose LodashWrapper

calebboyd:
This PR exposes `LodashWrapper` on the `lodash` function. 
It also alters how methods return intermediate LodashWrappers that allows chaining with a user defined subclass of `LodashWrapper`

related to #866


jdalton:
Thanks for the PR @calebboyd, but I don't want to expose the `LodashWrapper` at this time.

PR #1296, opened 2015-06-23
Clarify short cut fusion documentation [ci-skip]

megawac:
I think this reads more clearly :)


d10:
LGTM. (BTW it should be `[ci skip]` not `[ci-skip]`.)

PR #1311, opened 2015-07-01
Add _.lines method

megawac:
`_.lines` seems like a natural complement to `words` and is a pretty common need for strings.

Note: the implementation is identical to words and the two functions can be easily share a creator function (I can do that in a subsequent commit if you wish).

The selected line feed characters are based on https://en.wikipedia.org/wiki/Newline#Unicode

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -138,6 +138,8 @@
     return RegExp(upper + '+(?=' + upper + lower + ')|' + upper + '?' + lower + '|' + upper + '+|[0-9]+', 'g');
   }());
 
+  var reLines = /[^\r?\n|\u2028|\u2029|\u000a|\u000b|\u000c]+/g;
+
   /** Used to assign default `context` object properties. */
   var contextProps = [
     'Array', 'ArrayBuffer', 'Date', 'Error', 'Float32Array', 'Float64Array',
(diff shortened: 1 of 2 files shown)


lodash.src.js
```diff
@@ -138,6 +138,8 @@
     return RegExp(upper + '+(?=' + upper + lower + ')|' + upper + '?' + lower + '|' + upper + '+|[0-9]+', 'g');
   }());
 
+  var reLines = /[^\r?\n|\u2028|\u2029|\u000a|\u000b|\u000c]+/g;
```
jdalton: `\u000a` is `\r`
`\u000b` is `\v`
`\u000c` is `\f`

Line terminators are specified [here](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-line-terminators):

jdalton:
As with other additions I think we should hold off until there's a bit more demand.

megawac:
I think this one should be added as its extremely common (reading a file/multi-line text fields/etc) and is part of nearly every popular programming languages string API. Also if its created with a `createFeedSpliter` (or whatever) function to generate `words` and `lines` it ends up only requiring 3 LOC (regexp definition + creator func). Seems like a low hanging fruit to me

lodash.src.js
megawac: Ah thanks, I was looking for those

jdalton:
> I think this one should be added as its extremely common

Then the :+1:'s will come rolling in and it'll be added for sure.

mathieumg:
:+1: :P

PR #1326, opened 2015-07-08
Minor fix in documentation of `includes`: `value` -> `target`

lslah:


jdalton:
Thanks @lslah! While not necessary for this kind of contribution have you [signed our CLA](http://dojofoundation.org/about/claForm)? It'll streamline future contributions.

PR #1334, opened 2015-07-12
Save 1 millisecond!

James-Dunn:
The default action of setTimeout will move the function to the end of the call stack. No need to wait one millisecond.

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -7203,7 +7203,7 @@
      * // logs 'deferred' after one or more milliseconds
      */
     var defer = restParam(function(func, args) {
-      return baseDelay(func, 1, args);
+      return baseDelay(func, 0, args);
     });
 
     /**


oriSomething:
it won't help, `setTimeout` at the minimum delay in **4** seconds. the only option to save time is using `postMessage`

James-Dunn:
It doesn't look like current browsers observe that spec. Testing it on Chrome/Firefox seems to show no minimum for me. http://jsbin.com/foloponuli/edit?html,js,output The word "second" appears first for me if the first timer is set at 2 ms or higher on Chrome, and 1 ms on Firefox.

oriSomething:
you don't test it right. the browser will execute `setTimeout(..., 1)` before  `setTimeout(..., 2)`, no matter what the order, but still the first that will run is after 4ms.

oriSomething:
read [this](http://www.w3.org/TR/html5/webappapis.html#dom-windowtimers-settimeout)

megawac:
@jdalton to derail this a bit, want to consider using `setImmediate` or `process.nextTick` where available?

oriSomething:
also using `MessageChannel` / `postMessage` as a polyfil might be helpful

jdalton:
We avoid a `setTimeout` of `0` because I believe Chrome could end up blocking the UI thread with them.

I avoid `setImmediate` because there's no clean way to cancel them with `clearTimeout`. 
We have `runInContext` which allows you to customize if you want.
```
var defer = require('lodash').runInContext({ 'setTimeout': setImmediate }).defer;
```

PR #1337, opened 2015-07-12
Enable shortcut fusion for find and findLast

megawac:


jdalton:
I'm cool with this because it can be supported with minimal effort. I had a local commit that will cause issues so you'll need to rebase after I commit (in a min).

test/test.js
```diff
@@ -3994,6 +3994,40 @@
 
         deepEqual(actual, expecting);
       });
+
+      if (!/(Index|Key)/.test(methodName)) {
+        test('should enable shortcut fusion in chaining', 2, function() {
+          if (!isNpm) {
+            var callCount = 0;
+            var items = _.range(0, 251);
```
jdalton: Look at some of the other lazy sequence tests for their use of `LARGE_ARRAY_SIZE`.

lodash.src.js
```diff
@@ -11692,9 +11702,10 @@
           if (onlyLazy) {
             value = value.clone();
             value.__actions__.push(action);
-            return func.call(value);
+            return func.apply(value, args);
           }
-          return lodashFunc.call(undefined, this.value())[0];
+          var result = this.thru(interceptor).value();
+          return isTaker ? result[0] : result;
```
megawac: Thoughts on the approach in https://github.com/lodash/lodash/commit/5e9ca3fe9a3946b0858b0481e94a65e2011c66f4 vs this one?

jdalton: Neat approach. It's a bit more work for it to do but cleans things up.

megawac: Yeah, figured it was a bit more expensive - which should I keep post-rebase?

lodash.src.js
```diff
@@ -11680,7 +11690,7 @@
           isLazy = useLazy = false;
         }
         var interceptor = function(value) {
-          return (retUnwrapped && chainAll)
+          return (isTaker && chainAll)
             ? lodashFunc(value, 1)[0]
```
megawac: This should be faster no?

megawac:
Also happy birthday! The :gift: has been rebased

jdalton:
Closed by https://github.com/lodash/lodash/commit/729a557d8ac412e408ddb672702faca02499d562.

PR #1338, opened 2015-07-12
Add test coverage for NodeList as function

megawac:


jdalton:
Thanks, I haven't added tests for function iteration. I'll add them when I address the issue.

PR #1340, opened 2015-07-13
Make sure root is the global object. Fix lodash in Firefox extensions

iamolivinius:
Lodash does not run within the content script of a Firefox Bootstraped extensions. Firefox add-on SDK extensions are not affected.

The basic problem is to get the correct `global` object.

Within the content script sandbox properties like `window` and `self` are available and refer to the current window/tab. Unfortunately `window` and `self` are not the global object inside the sandbox.

For example: 

```js
window.Math === Math // false
self.Math === Math   // false
```

https://github.com/lodash/lodash/blob/3.10.0/lodash.js#L274

At this point `root` has a reference to the wrong object. A working solution can be found here: https://github.com/zloirock/core-js/blob/master/modules/%24.js#L2

For more information and discussion, see: 

https://github.com/zloirock/core-js/pull/91
https://github.com/zloirock/core-js/issues/86

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -279,9 +279,7 @@
    * The `this` value is used if it's the global object to avoid Greasemonkey's
    * restricted `window` object, otherwise the `window` object is used.
    */
-  var freeRoot = freeGlobal || ((freeWindow !== (thisGlobal && thisGlobal.window)) && freeWindow) || freeSelf || thisGlobal;
-  
-  var root = freeRoot.Math === Math ? freeRoot : Function('return this')();
+  var root = freeGlobal || ((freeWindow !== (thisGlobal && thisGlobal.window)) && freeWindow) || freeSelf || thisGlobal || Function('return this')();
 
   /*--------------------------------------------------------------------------*/
 
@@ -431,7 +429,7 @@
    * @returns {null|Object} Returns `value` if it's a global object, else `null`.
    */
   function checkGlobal(value) {
-    return (value && value.Object) ? value : null;
+    return (value && value.Object === Object) ? value : null;
   }
 
   /**
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -279,7 +279,9 @@
    * The `this` value is used if it's the global object to avoid Greasemonkey's
    * restricted `window` object, otherwise the `window` object is used.
    */
-  var root = freeGlobal || ((freeWindow !== (thisGlobal && thisGlobal.window)) && freeWindow) || freeSelf || thisGlobal;
+  var freeRoot = freeGlobal || ((freeWindow !== (thisGlobal && thisGlobal.window)) && freeWindow) || freeSelf || thisGlobal;
+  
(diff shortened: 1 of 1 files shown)


jdalton:
Thanks @iamolivinius! I believe this is resolved in edge.

iamolivinius:
By edge you mean the current master?

jdalton:
Yep.

iamolivinius:
Tested this file: https://github.com/lodash/lodash/blob/master/lodash.src.js

Bug is not resolved.

jdalton:
Yay! Can you dig a bit to see which branch it ends up taking in the bug case.

megawac:
Want to reopen? @jdalton you can consider using underscores approach which works in FF sandboxs last I checked

iamolivinius:
@megawac Can confirm this. Checked it a week ago.

iamolivinius:
@jdalton I'd love to help out but I'm not sure if I get it right what you were asking for. (non-native speaker)

You want me to find out in which branch the bug is present, right?!

megawac:
@iamolivinius can you verify https://github.com/megawac/lodash/commit/46baf7d3848c10ced2f17d8d741fda662a77054c solves your issue?

iamolivinius:
No, it doesn't fix the problem.

iamolivinius:
`window` as well as `self` will return a function in `checkGlobal(value){ value.Object }`

jdalton:
How about simply changing `checkGlobal` to:
```js
function checkGlobal(value) {
  return (value && value.Object === Object) ? value : null;
}
```

iamolivinius:
In this case `freeGlobal || freeWindow || freeSelf || thisGlobal` would be `null` if you use things like `browserify --bare` cause of the encapsulating self invoking function.

How about your suggested `checkGlobal` and this: 

```js
var root = freeGlobal || freeWindow || freeSelf || thisGlobal || Function('return this')()
```

jdalton:
I think the Hail Mary at the end is fine. It won't likely hit so the issues with CSP enviros should be minimum.

jdalton:
Feel free to make the modifications. Have you signed [our CLA](http://dojofoundation.org/about/claForm)?

iamolivinius:
I signed the CLA a few minutes ago.

jridgewell:
CSP isn't done by static analysis?

iamolivinius:
Are you referring to `Function('return this')()`?

jdalton:
@jridgewell
> CSP isn't done by static analysis?

Nope.

@iamolivinius
> I signed the CLA a few minutes ago.

Rock!

indieisaconcept:
Has this fix made it to npm yet? This solves an issue I have been having using a browserified version of lodash within node-vm.

The latest version "3.10.1" has this change omitted.

jdalton:
This will make it in to v4.0.

indieisaconcept:
@jdalton thanks - I'm pulling from master in the interim.

PR #1341, opened 2015-07-13
Fix root detection of self and window in FF sandboxes

megawac:


jdalton:
I want to be careful here as there's also an issue with greasemonkey. I'm half thinking to just punt because these issues are so edge case on such wonky enviros.

megawac:
Works for me, and it only gets wonkier when you consider some of the
wonkieness which may be introduced via bundler tools wrapping lodash

jdalton:
Favoring fresh face contrib of #1340.

PR #1348, opened 2015-07-15
Added example of_.isString()

lekoaf:
I thought it would be appropriate to add another example of the _.isString() function. An empty string is usually considered a falsy value but, of course, not when it comes to _.isString().

Just stumbled upon this in the codebase I was working on as it caused a bug.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -8972,6 +8972,9 @@
      * _.isString('abc');
      * // => true
      *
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @lekoaf! Thanks for taking the time to create a PR. In this case though I think the string doc examples are sufficient. As you say it's all in the name.

PR #1350, opened 2015-07-15
Re-word _.merge docs for clarity

LeekAnarchism:
Previously, when scanning the docs, it could easily be accidentally read as:

'...that don't resolve to 'undefined' **in** the destination object.'

(i.e. "in" instead of "into")

Both myself and a collegue read it this way, until we looked at it more closely.

I've attempted re-pharsing it here to remove any ambiguity, although it's a bit more verbose.

Alternatively, we could just put a comma after 'undefined', which would help to separate it from "into the destination object".

### DIFF
--- a/lodash.src.js
+++ b/lodash.src.js
@@ -9127,9 +9127,9 @@
     }
 
     /**
-     * Recursively merges own enumerable properties of the source object(s), that
-     * don't resolve to `undefined` into the destination object. Subsequent sources
-     * overwrite property assignments of previous sources.
+     * Recursively merges own enumerable properties of the source object(s) into the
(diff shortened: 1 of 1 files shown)


jdalton:
Have you signed [our CLA](http://dojofoundation.org/about/claForm). While not necessary for this kind of PR it will help with future PRs.

LeekAnarchism:
Hi John -- I didn't sign it as it seemed like a lot of personal information to give away for a small documentation change. If I contribute more significant PRs I will be sure to sign it, but honestly I couldn't be bothered at 10pm last night :)

Thanks.

jdalton:
No worries.

PR #1357, opened 2015-07-20
Add String to valid includes parameters [ci skip]

megawac:
There is still an example for includes with a string, as discussed in gitter this should be here if the behaviour is sticking around

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -5990,7 +5990,7 @@
      * @static
      * @memberOf _
      * @category Collection
-     * @param {Array|Object} collection The collection to search.
+     * @param {Array|Object|string} collection The collection to search.
      * @param {*} target The value to search for.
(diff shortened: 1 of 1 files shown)


lodash.js
```diff
@@ -5990,7 +5990,7 @@
      * @static
      * @memberOf _
      * @category Collection
-     * @param {Array|Object} collection The collection to search.
+     * @param {Array|Object|String} collection The collection to search.
```
jdalton: I have this in my local lol. Make the `String` -> `string` and it's good to go.

megawac: Whoopsies, always forget those jsdoc semantics

On Mon, Jul 20, 2015 at 7:43 PM, John-David Dalton <notifications@github.com

jdalton: Toasty!

PR #1360, opened 2015-07-22
Added npm and dependency badges to readme

manikandants:


jdalton:
Thanks @manikandants, but I avoid badges in the readme because they tend to represent the edge state and not the published package.

manikandants:
Ok no probs @jdalton I saw that lodash was the most dependent module in npm registry :+1:  Thought would be good mentioning here :)

Issue #1252, opened 2015-05-31
_.delete/_.remove for objects using a path.


andrewhathaway:
It would be great to be able to remove a value from an object, including it's key using a path just like `_.get` and `_.set`.

```javascript
_.remove(object, 'a[0].b.c');
```

jdalton:
Hi @andrewhathaway! For now see #962.

arxpoetica:
Considering this is a fairly common operation in JS, :+1: :+1: 

(Examples: `delete someVar;`, `localStorage.removeItem('keyPath')`, etc.

megawac:
I agree, `_.remove(object, 'a.b.c')` could be very useful (I'll PR it if you want)

jdalton:
`_.remove` is for arrays.  See #962.

megawac:
I suppose we're still trying to avoid overloading @jdd?

jdalton:
Moving a method to the Collection category is something I want to drag my feet on because it has a lot of baggage with it.

megawac:
How about unset then?

arxpoetica:
Agreed, refactoring `_.remove` as collection === :sob:, but `_.delete` or `_.unset`.

jdalton:
I can dig `unset`.

travisdahl:
I love it when I go searching for something and people are currently talking about it. You guys are awesome.

+1

jdalton:
The next step is for someone to create a PR for `_.unset`.

Braincompiler:
Why not extending `omit`?

arxpoetica:
:zap: :zap: :zap: !!!

andrewhathaway:
:beers:

nabati:
Out of curiosity, why does `unset` return a boolean and not the object itself as does `set`? This behaviour was unexpected for me.

jdalton:
HI @nabati!

It's modeled after `delete x` or `Reflect.deleteProperty` which return boolean values from their operations.

PR #1386, opened 2015-08-07
Remove keys when use `_.set` with `undefined` values

ricardohbin:
The `_.set` and `_.get` (or `_.result`) works very well to "eval" some expressions. With this PR, it's possible to delete some keys: just set them to `undefined`


jdalton:
I want to avoid deleting unless it's explicit. You could retool this for as an [unset method](https://github.com/lodash/lodash/issues/1252#issuecomment-115898801).

Issue #1370, opened 2015-07-27
[feature] get penultimate element


jasonLaster:
Maybe this is the rubyist in me showing through, but I'd love to be able to use negative indices get elements from right to left.

Here's what I'm proposing:

```js
var element = _.at(['a','b','c'], -2); // => 'b'
```

jdalton:
Hi @jasonLaster! In the master branch I've made `_.at` into an "Object" method that accepts paths (shallow or deep). So specializing for arrays seems out of place for it now.

jasonLaster:
Oh crazy, I'll have to take a look. Do you think i could find another more appropriate method for array indexing? Someone's gotta standup for the little square guy.

jdalton:
I've considered negative indexes for `get`, `result` and friends but ran into similar issues.

StreetStrider:
@jasonLaster, do you really need to get `-2`, `-3`… in production code? Or you're wanting this for REPL/playground tasks?
I'm asking because (for me) the most common negative index is `-1` which is covered by `_.last` (which is also much more readable than any negative numbers).

chrisbolin:
I think `_.nth` should work for you
```javascript
_.nth(['a','b','c'],-2)
> 'b'
```

Issue #1315, opened 2015-07-02
Support array push notation in _.set


DylanPiercey:
Currently when using set it will support the computed property notation so that I can go

```javascript
_.set({}, "a[b][c]", 1)
// -> { a: { b: { c: 1 } } }
```

This works great for turning a urlencoded form's input name into an object however it would be nice if the "push" notation was also valid (currently it does nothing).

By that I mean:

```javascript
_.set({ a: [1] }, "a[]", 2)
// -> { a: [1, 2] }

// currently -> { a: 2 }
```

Is this possible? It would be pretty handy.

jdalton:
I can dig it, also negative index to get the last element. Do you want to try your hand at a pull request.

DylanPiercey:
I can attempt it later today - not super familiar with the lodash source but i'll post back here if I can't get it working properly.

jdalton:
Things to consider which may put a damper on this is what happens when the value you're applying this to is not an array?

In [previous issues](https://github.com/lodash/lodash/issues/1145#issuecomment-94345635) I've punted for just doing:
```js
_.get(object, path).push(value)
```

DylanPiercey:
@jdalton - Not sure if it's what we want, but when QS parses forms with something like

```javascript
a[] = 1
a[test] = 2
a[test2] = 3
a[] = 4
```

It simply add's incremental indexes to the object so that that psuedo form becomes:

```javascript
{
  "a": {
    "0": "1",
    "1": "4",
    "test": "2",
    "test2": "3"
  }
}
```

IMO: This is what I would expect.

jdalton:
Ah nice. Ok.

DylanPiercey:
@jdalton is it okay if array paths with empty strings also act as a push operation?

```javascript
_.set({}, ["a", ""], 1)
// -> { a: [1] }
```

It would be consistent with the idea that [] is just an empty path.

jdalton:
At the moment empty strings are considered a key of an empty string which is valid. Let's keep the push syntax to the `[]`.

jdalton:
And with that the problem of consistent behavior across path types rears up. I think we'll punt on this for now.

DylanPiercey:
Alrighty. Don't have time to finish my PR atm anyway.

Thinking it over again though it makes the most sense to simple have [] act as push/pop. I think merging this with negative indexes and empty strings makes it more than it is.

jdalton:
[...]

PR #1396, opened 2015-08-13
WIP - #1370 Add support for negative indices

jasonLaster:
This is a POC implementation for supporting negative indices. There are several performance regressions and shortcuts.

Negative indices are a common pattern that make it easy for developers to get the last, 2nd-last, value in an array.

The way the implementation was added, it'd have far reaching implications.

```js
_.result(['a','b','c'], -3) //=> 'a'
_.get(['a','b','c'], -2) //=> 'b'
_.at(['a','b','c'], -1) //=> ['c']
```

It's also possible to support a path push notation where `a[-0]` could indicate the array location after the last location. This was brought up by #1315.

```js
_.set({ a: [1] }, 'a[-0]', 2) //=> { a: [1, 2] }
```

@jridgewell and I paired on this.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -1978,10 +1978,16 @@
       path = isKey(path, object) ? [path + ''] : toPath(path);
 
       var index = 0,
+          part,
           length = path.length;
 
       while (object != null && index < length) {
-        object = object[path[index++]];
+        part = path[index++];
+        if (isArrayLike(object) && /^-\d+$/.test(part)) {
+          part = object.length + (+part);
+        }
+
+        object = object[part];
       }
       return (index && index == length) ? object : undefined;
     }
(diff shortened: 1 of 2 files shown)


lodash.js
```diff
@@ -9381,13 +9387,13 @@
      */
     function result(object, path, defaultValue) {
       var isPath = !isKey(path, object),
-          result = (isPath || object == null) ? undefined : object[path];
+          result = (isPath || object == null) ? undefined : baseGet(object, [path]);
```
jasonLaster: This removes a common case optimization and should be better thought through.

lodash.js
```diff
@@ -1978,10 +1978,16 @@
       path = isKey(path, object) ? [path + ''] : toPath(path);
 
       var index = 0,
+          part,
           length = path.length;
 
       while (object != null && index < length) {
-        object = object[path[index++]];
+        part = path[index++];
+        if (isArrayLike(object) && /^-\d+$/.test(part)) {
```
jasonLaster: The regex tests for both whether the string is number like and negatives. I'm sure there's a better way to do this

jdalton:
Ahh @jasonLaster! I don't wanna be the bad guy here but I've opted for simplicity for this.
The `at` method is now an "Object" method so a negative index could be a property.

:sob:

jridgewell:
> The `at` method is now an "Object" method

Isn't that just `_.pick`?

jdalton:
Close, `pick` returns an object (key + values), `at` returns an array of values.

jridgewell:
Ahh.

PR #1401, opened 2015-08-20
Add _.isNil function

xixilive:
```
_.isNil(null); // true
_.isNil(); // true
_.isNil(undefined); // true
_.isNil(NaN); // false
```


lodash.js
```diff
@@ -8244,6 +8244,35 @@
...
+     * // => true
+     *
+     * _.isNil(NaN);
+     * // => false
+     *
+     * _.isNil(void 0);
+     * // => false
+     */
+    function isNil(value) {
+      return undefined === value || null === value;
```
jdalton: The yoda conditions aren't really our style. Could you make this `return value == null;`.

jdalton:
You don't have to close the PR. Just simply amend the commit and force push.

xixilive:
There appears a merge log, so ....

PR #1405, opened 2015-08-21
Added additional item to _.fill example array

MaxPRafferty:
While not the minimum example required to show the fill functionality, having the fill write more than one value better fits the purpose of the function in practice, thereby making the example clearer and more useful.


jdalton:
Thanks!

PR #1407, opened 2015-08-24
The underscore(1.8.3) has method _.all which is  _.every 's alias,but not in lodash

178220709:
i try to use lodash.js to replace underscore.js
but throw a error , _.all is undefined
http://devdocs.io/lodash/index#every
The doc of lodash said lodash has all too ,but i cant find in code
so i add the _.all method  .
just direct to every

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -11508,6 +11508,7 @@
     lodash.escape = escape;
     lodash.escapeRegExp = escapeRegExp;
     lodash.every = every;
+    lodash.all = every;
     lodash.find = find;
     lodash.findIndex = findIndex;
     lodash.findKey = findKey;


StreetStrider:
@178220709, which version?
```js
js > lodash.VERSION
'3.10.1'
js > [ lodash.all, lodash.any ]
[ [Function: every], [Function: some] ]
```

jdalton:
As part of our 4.0 work lodash is dropping most of its aliases.

StreetStrider:
@jdalton, that was one of my favorites :C

jdalton:
Aliases are trivial to add, a cleaner, clearer API is a bigger win.

StreetStrider:
@jdalton, ye, agreed, I'm not complaining. It would always be possible to alias it by wish:
```js
var all = require('lodash').every
```

178220709:
ok , it is not a serious problem.
i will  add a mask to the _ to adapter my old code .

PR #1416, opened 2015-08-29
Insert sorted

onumossn:
It seems inserting into a sorted index would be a pretty common case. Though it isn't difficult to use _.sortedIndex and Array splice, I thought it would be nice to have it included.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -5000,19 +5000,80 @@
     }
 
 
+    /**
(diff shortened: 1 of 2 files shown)


jdalton:
Thanks for taking the time to create a PR but we'll wait and see if there's demand for something like this before considering a PR.

PR #1417, opened 2015-08-29
Constant property

onumossn:
Allows _.constant to create an un-editable property on a given object.


jdalton:
Thanks @onumossn but there's no way to shim this functionality so we won't be adopting it.

PR #1420, opened 2015-08-31
Add ES2015 compliant lodash(..).next() iterator

megawac:
As previously discussed this adds the opportunity for users to use a lodash instance in a `for...of` loop or in any native/non-native JS interface which accepts an iterator.

Further, as we've discussed, I'd like to see this as a stepping stone to implementing lodash chains through iterators (https://github.com/megawac/lodash/commit/22e7beb36382dab224ea091cdcd0f9218039d1bc#commitcomment-12960941)

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -1492,6 +1492,8 @@
       this.__wrapped__ = value;
       this.__actions__ = actions || [];
       this.__chain__ = !!chainAll;
+      this.__value__ = null;
+      this.__index__ = 0;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jdalton:
Woot. Thanks!

lodash.js
```diff
@@ -5868,6 +5870,42 @@
...
+     * // => {done: false, value: 2}
+     *
+     * console.log(wrapped.next());
+     * // => {done: true, value: undefined}
+     *
+     * var array = [1, 2];
+     * var wrapped = _(array);
+     *
+     * // Use with an ES6 methods which accept an iterable
+     * console.log(Array.from(wrapped))
```
jdalton: `Array.from` requires a `Symbol.iterator` function.

lodash.js
```diff
@@ -5868,6 +5870,42 @@
...
+     * // => [1, 2]
+     */
+    function wrapperNext() {
+      if (this.__value__ == null) this.__value__ = this.value();
```
gibson042: This means that calling `next` will terminate laziness, right?

jdalton: Yes.

megawac: Currently yes, I'm playing with an implementation where that isn't the case

megawac: @jdalton is it worth updating `.value()` to use/set this value? With this set we can also drop the original `__wrapped__` item for some potential space savings

jdalton: > is it worth updating .value() to use/set this value? With this set we can also drop the original __wrapped__ item for some potential space savings

I'm not following you. Can you clarify?

megawac: Sure,
1) The `__value__` can be set by `.value()` and returned by `.value()` when it exists.
2) Since the resulting value, `__value__`, is known we can delete `__wrapped__` as all  computations have completed (to this point).

jdalton: Caching the value in `__value__` may lead to some issues. At the moment the chaining creates a sequence of operations that are executed against a `__wrapped__` value. If changes are made to the underlying `__wrapped__` value the sequence could produce different results.

PR #1421, opened 2015-08-31
Combinepredicates

menelaos:
Related to
https://github.com/lodash/lodash/issues/1072
https://github.com/lodash/lodash-fp/issues/19

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -1393,7 +1393,8 @@
      * and `toArray`
      *
      * The chainable wrapper methods are:
-     * `after`, `ary`, `assign`, `assignWith`, `at`, `before`, `bind`, `bindAll`,
+     * `after`, `allPass`, `anyPass`, `ary`, `assign`, `assignWith`, `at`,
+     * `before`, `bind`, `bindAll`,
      * `bindKey`, `chain`, `chunk`, `commit`, `compact`, `concat`, `constant`,
      * `countBy`, `create`, `curry`, `debounce`, `defaults`, `defaultsDeep`,
      * `defer`, `delay`, `difference`, `drop`, `dropRight`, `dropRightWhile`,
--- a/lodash.js
+++ b/lodash.js
@@ -6888,6 +6888,40 @@
     }
 
     /**
+     * Combines multiple predicates into a single predicate that returns true
+     * if all predicates are satisfied and false otherwise.
+     *
+     * @static
+     * @memberOf _
+     * @category Function
+     * @param {Array} predicates The array of predicates.
+     * @returns {Function} Returns the combined predicate.
+     *
+     * var largeAndEven = _.allPass([x => x > 1000, x => x % 2 === 0 ]);
+     * largeAndEven(2000);
+     * // => true
+     */
+    function allPass(predicates) {
+      var args;
+      if (!predicates) {
+        args = [];
+      } else {
+        args = arguments;
+        isArray(args[0]) && (args = args[0]);
+      }
+      var length = args.length;
+      return function() {
+        var index = -1;
+        while (++index < length) {
+          if (args[index].apply(this, arguments) === false) {
+            return false;
+          }
+        }
+        return true;
+      };
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


jdalton:
I think the implementation of these could be simplified using `arrayEvery` and `arraySome`.
Also you can use `restParam` to avoid the array juggling.

jdalton:
@megawac Any opinions on naming? (`and` and `or`)?

megawac:
I think `allPass`/`anyPass` is better names than `and` & `or` which would be better suited to be simple proxies to the operators.

I don't have much of a preference for `anyPass` vs `somePass` or `allPass` vs `everyPass`, though `everyPass` seems distasteful somehow

Relevant pull request for underscore: https://github.com/jashkenas/underscore/pull/2124

jdalton:
Good point on the names. I think sticking with `every` and `some` variants is a better fit since v4 will be dropping the `any` and `all` aliases.

StreetStrider:
`conj` & `disj`.

jdalton:
It kind of feels like a shame to create new methods for this. I'd love a composible option.

menelaos:
The names [`allPass`](http://ramdajs.com/0.17/docs/#allPass) and [`anyPass`](http://ramdajs.com/0.17/docs/#anyPass) are taken from Ramda. `everyPass` and `somePass` would be more consistent with `lodash` naming conventions, you are right. 
A more generic solution could be something like:
~~~javascript
function liftA(combine) {
  return function(fs) {
    if (!isArray(fs)) { throw new Error("Could be extended to non-array arguments"); }
    var length = fs.length;
    return function() {
      var res = Array(length),
          index = -1;
      while (++index < length) {
        res[index] = fs[index].apply(this, arguments);
      }
      return combine(res);
    };
  };
}

  var allPass = liftA(every);
  var anyPass = liftA(some);
~~~

Drawback: if `combine` can be an arbitrary function of type `Array -> Value`, then I believe we would have to iterate over all `fs` which might not be necessary in the case of `fs` being predicates.

jdalton:
Avoiding `everyPass` as @megawac points out would be a good thing. Lets think of a name that avoids `some`, `every`, `all`, `any` like `conj` & `disj` but not those :)

stevewillard:
Any progress on this? Looking to use allPass in a project -- FWIW I like allPass and anyPass.

StreetStrider:
BTW, anyone interested in this can port [my code](https://github.com/StreetStrider/aux.js/commit/02285f9095f07aaa113e1bd6319c5745b55342c8) to some kind of lodash-mixin. It is well-tested and&nbsp;bla-bla-bla.

stevewillard:
Cool! Was able to port our code over to what I'm doing. Thanks a bunch

jdalton:
@stevewillard
> Any progress on this?

Still bike shedding a name.

jdalton:
If a name isn't settled on I'll have to close the PR to marinate a bit more.

stevewillard:
If you're going with every/all in v4, then probably makes sense to work that in. Any of these? I kind of like the last one.

```everySatisfy``` and ```someSatisfy```
```everyYield``` and ```someYield```
```everyAllow``` and ```someAllow```

jdalton:
`someTrue`, `everyTrue`?

jdalton:
Nm nm nm.

jdalton:
Ok more bikeshedding. Maybe it's a good idea for `anyPass` and `allPass` so they don't get confused with `some` and `every`. Seeing as at a glance you could name `some` as `anyPass` as in any element passes the predicate check.

I also want to add `ifElse` method so maybe there's a category they could fall under.

Wideshanks:
These are the same as Ramda's [allPass](http://ramdajs.com/0.17/docs/#allPass)/[anyPass](http://ramdajs.com/0.17/docs/#anyPass), aren't they? Seems like a good name.

jdalton:
Sorry for the hold up here. I'm closing this PR as I have it implemented it locally as `_.conj` and `_.disj` using the helper `createConnective`.

PR #1422, opened 2015-08-31
Add coverage for wrapper[Symbol.iterator]

megawac:


test/test.js
```diff
@@ -42,6 +42,7 @@
   var phantom = root.phantom,
       amd = root.define && define.amd,
       argv = root.process && process.argv,
+      jsArrayFrom = Array.from,
```
jdalton: No need for this. You can use `_.toArray(wrapped)`.

megawac: I thought it'd be better to ensure it conforms to an actual browser implementation

test/test.js
```diff
@@ -16598,7 +16599,7 @@
 
     var chainType = ' in an ' + (implict ? 'implict' : 'explict') + ' chain';
 
-    test('should produce an ES6 compliant object' + chainType, 7, function() {
+    test('should follow ES2015 iterator specification' + chainType, 7, function() {
```
jdalton: Don't worry about the test label here. I'm changing it in a cleanup pass.

test/test.js
jdalton: `_.toArray` conforms to it so that's fine and lets it be tested in Chrome and Node which lacks `Array.from`.

jdalton: I still need to add unit tests for `_.toArray` and the like (more for this evening).

jdalton:
Thanks Graeme!

PR #1424, opened 2015-09-02
[PR] Base implementation of `_.unset`.

falsyvalues:
Hey there!

I saw lodash/lodash/issues/1418 so here's my attempt to deliver ```unset```.
I wasn't sure how to handle paths that doesn't exist so there is a ```undefined``` as a third possible value in return.

CLA signed (submitted a CLA on 2015-09-02 09:19:55)

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -2738,6 +2738,21 @@
     }
 
     /**
+     * The base implementation of `_.unset`.
+     *
+     * @private
+     * @param {Object} object The object to query.
+     * @param {Array|string} path The path of the property to unset.
+     * @returns {boolean} Returns delete operator result.
+     */
+    function baseUnset(object, path) {
+      path = isKey(path, object) ? [path + ''] : toPath(path);
+      object = parent(object, path);
+      var key = last(path);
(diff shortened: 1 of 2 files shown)


lodash.js
```diff
@@ -2738,6 +2738,29 @@
...
+     * @param {Array|string} path The path of the property to unset.
+     * @returns {boolean} Returns delete operator result.
+     */
+    function baseUnset(object, path) {
+      path = isKey(path, object) ? [path + ''] : toPath(path);
+
+      var index = -1,
+          lastIndex = path.length - 1,
+          key;
+
```
jdalton: Since this is unsetting you can use `object = parent(object, path);` once the path is resolved then:
```js
if (object) {
  delete object[last(path)];
}
```

falsyvalues:
Thanks for review, I have updated commit. I'm still using ```has``` to distinguish case when key is missing on object.

jdalton:
Thanks @falsyvalues!

falsyvalues:
Thanks @jdalton for lodash! :+1:

PR #1427, opened 2015-09-02
Added deep filtering to compact function

rheh:
Proposed enhancement to allow compact function to perform a deep filter.  New test added and existing tests executed: all still pass.


jdalton:
Thanks for the PR @rheh! I want to keep `compact` simple. I'll keep an eye on this to see if there's demand. If there is it would likely land as another method `compactDeep`.

PR #1429, opened 2015-09-02
Replace œ by oe in _.deburr

NTag:
I updated _.deburr
Œ -> Oe
œ -> oe


jdalton:
Moved to #1430 for discussion first.

PR #1433, opened 2015-09-02
Addition of jscs

rheh:
Added JSCS to the project (http://jscs.info/)  Initially imported the google code style preset (https://github.com/jscs-dev/node-jscs/blob/master/presets/google.json) and modified to suite the current code style.  This should enforce such gems as the automation of tabs equating to 2 spaces, missing semi-colons, curly braces enforment etc.

To automate or run manually one can execute "npm run lint"

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -1913,7 +1913,7 @@
       return function(prototype) {
         if (isObject(prototype)) {
           object.prototype = prototype;
-          var result = new Object;
+          var result = new object;
           object.prototype = undefined;
         }
         return result || {};
--- a/lodash.js
+++ b/lodash.js
@@ -1976,7 +1976,8 @@
             }
           }
           result.push(value);
-        } else if (indexOf(values, value, 0) < 0) {
+        }
+        else if (indexOf(values, value, 0) < 0) {
           result.push(value);
         }
       }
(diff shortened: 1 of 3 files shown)


lodash.js
```diff
@@ -1976,8 +1976,7 @@
             }
           }
           result.push(value);
-        }
-        else if (indexOf(values, value, 0) < 0) {
+        } else if (indexOf(values, value, 0) < 0) {
```
jdalton: I'm not nit picky on the `} else if` bit if it's able to just let that slide (I use whichever is more readable for the situation).

rheh:
I've reversed the 'else' statement changes and modified the rules according to accept the either style.

rheh:
@falsyvalues, ideally yes; but, @jdalton fed-back a desire for this rule to be a little looser.

lodash.js
```diff
@@ -7328,8 +7327,7 @@
             }
             lastCalled = stamp;
             result = func.apply(thisArg, args);
-          }
-          else if (!maxTimeoutId) {
```
jdalton: just a couple of new-line nits left over.

lodash.js
```diff
@@ -2932,8 +2932,7 @@
             seen.push(computed);
           }
           result.push(value);
-        }
```
jdalton: '' ''

jdalton:
Thanks @rheh! Have you signed [our cla](http://dojofoundation.org/about/claForm)?

rheh:
@jdalton sure have, a few months back when I did my first pull request.

jdalton:
Rock!

Issue #1455, opened 2015-09-10
`_.merge`: Unexpected behaviour in merging `class` instance properties.


rhys-vdw:
When providing instances of a class as the `object` and `sources` arguments for `merge`, their properties are merged as expected. However, any similar properties are overridden without being merged.

```js
class Thing {
    constructor(properties) {
        _.extend(this, properties);
    }
}

const x = new Thing({a: 'a', thing: new Thing({b: 'b'})});
const y = new Thing({thing: new Thing({c: 'c'})});

const merged = _.merge(x, y);
// {"a":"a","thing":{"c":"c"}}
```

[fiddle](https://jsfiddle.net/xfv3fq4g/)

Using the following customizer fixes the problem:

```js
function thingCustomizer(a, b) {
    if (a instanceof Thing || b instanceof Thing) {
        return _.merge(a, b);
    }
}
```

This is unexpected as `_.merge` behaves recursively for "plain objects". This behaviour is not specified in the documentation.

jdalton:
This is because when providing a `source` the properties on the source are iterated over to merge, the `source` itself is not assigned, but when you attempt to assign and merge a non-plain object we punt on it because non-plain objects may be DOM elements or other things.

You're approach of using a `customizer` callback to handle it is correct.

rhys-vdw:
Makes sense. Would you accept a PR to clarify the `merge` docs?

jdalton:
Yep!

PR #1458, opened 2015-09-12
Clarify `_.merge` description.

rhys-vdw:
Mention behaviour difference between merging plain objects and objects with non-Object prototypes. As discussed in #1455.


jdalton:
Thanks @rhys-vdw!

PR #1476, opened 2015-09-17
Fix typo in docs for _.contains()

stewx:
Should say "_.includes" instead of "_.include"


jdalton:
The alias's for `_.includes` is `_.contains` and `_.include`. In master branch lodash.js I have removed most aliases.

PR #1481, opened 2015-09-23
Add `_.isJSON` function

hairyhenderson:
Here's an `_.isJSON` function to help clean up code that would otherwise be littered with `try { JSON.parse(...); } catch (e) { }`.

I'm pretty sure that the performance of this function could be improved, but this is a simple first cut.

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

### DIFF
--- a/lodash.core.js
+++ b/lodash.core.js
@@ -537,7 +537,7 @@
    * `findLastIndex`, `findLastKey`, `first`, `floor`, `get`, `gt`, `gte`,
    * `has`, `hasIn`, `identity`, `includes`, `indexOf`, `inRange`, `isArguments`,
    * `isArray`, `isArrayLike`, `isBoolean`, `isDate`, `isElement`, `isEmpty`,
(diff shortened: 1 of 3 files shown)


jdalton:
Thanks @hairyhenderson!

You can use

```js
if (!_.isError(_.attemp(JSON.parse, value))) {
```

I'll keep an eye on this request to see if it gets traction.

hairyhenderson:
Thanks @jdalton - I wasn't aware of `_.attempt`. I still think something like

```js
if (_.isJSON(value)) {
```

is a lot cleaner.

Also, `JSON.parse` actually won't throw on certain non-JSON inputs (by which I mean inputs which aren't strictly valid JSON strings - I'm thinking of `JSON.parse(null)` or `JSON.parse(42)`, which return `null` and `42` even though they _aren't JSON strings_.

jdalton:
In the meantime you can create a `mixin`.

hairyhenderson:
@jdalton - good point, thank you!

Issue #1478, opened 2015-09-21
Floor not rounding properly for 2 decimals


pootzko:
Hi,

I found out that the `_.floor` has one case where it does not work properly:

```javascript
_.floor(4.1, 2)
// -> 4.09 (which is not expected - I would expect this to be 4.1)
```

where this is good:
```javascript
_.floor(4.11, 3)
// -> 4.11 (which is good)
```

I looked into the source code, and this is what gets executed:

```javascript
    function createRound(methodName) {
      var func = Math[methodName];
      return function (number, precision) {
          debugger;
        precision = precision === undefined ? 0 : (+precision || 0);
        if (precision) {
          precision = pow(10, precision);
          return func(number * precision) / precision;
        }
        return func(number);
      };
    }
```

what happens is that `number` is 4.1 and precision is `100` at the return inside `if`. Then `number * precision` is 409.99999999999994 (which should be 410 instead) and when you divide that by 100 you obviously get 4.09. So, I understand that this is really a JS problem with numbers with lots of decimals and I guess this is considered as "accepted" behavior by lodash (and it's not the kind of thing that lodash would handle)?

Cheers

jdalton:
I think this falls under "by design". I tried to think of a way to counter this (like doing a max compare between the result and the original number) but non panned out. \@thejameskyle.

gibson042:
@jdalton:
```patch
     function createRound(methodName) {
       var func = Math[methodName];
       return function(number, precision) {
         precision = precision === undefined ? 0 : (+precision || 0);
         if (precision) {
           precision = pow(10, precision);
-          return func(number * precision) / precision;
+          return func(round(number * precision)) / precision;
         }
         return func(number);
       };
     }
```

And if you want to get fancy, you can omit the redundant call from `_.round` by checking `methodName`.

gibson042:
Edit: The above breaks some negative-precision cases... better to use the algorithm at [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/floor#Decimal_adjustment), which will handle everything modulo toString-based precision loss.

jdalton:
@gibson042 Are you up for a PR?

gibson042:
Sure, but I don't think I understand your git structure... where do I branch from, and what does the workflow look like?

jdalton:
Branch from master, update the lodash.js file, and add a test to test/test.js.

jdalton:
[...]

PR #1487, opened 2015-09-24
Fix rounding issue with floor

tejokumar:
This is a fix for https://github.com/lodash/lodash/issues/1478.
I have used [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/floor#Decimal_adjustment) logic as mentioned in the issue.
Unit tests are also included.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -3699,12 +3699,13 @@
       return function(number, precision) {
         precision = precision ? toInteger(precision) : 0;
         if (precision) {
-          // Shift
-          number = number.toString().split('e');
-          number = func(+(number[0] + 'e' + (number[1] ? (+number[1] + precision) : precision)));
+          // Shift decimal point with exponential operations
+          number = (number + 'e').split('e'); // Force a split
+          number = func(number[0] + 'e' + (+number[1] + precision)); // Use ToNumber conversions
           // Shift back
-          number = number.toString().split('e');
-          return +(number[0] + 'e' + (number[1] ? (+number[1] - precision) : -precision));
+          number = (number + 'e').split('e');
+          return +(number[0] + 'e' + (number[1] - precision));
+
         }
         return func(number);
       };
--- a/lodash.js
+++ b/lodash.js
@@ -3700,8 +3700,8 @@
         precision = precision ? toInteger(precision) : 0;
         if (precision) {
           // Shift decimal point with exponential operations
-          number = (number + 'e').split('e'); // Force a split
-          number = func(number[0] + 'e' + (+number[1] + precision)); // Use ToNumber conversions
+          number = (number + 'e').split('e');
+          number = func(number[0] + 'e' + (+number[1] + precision));
           // Shift back
           number = (number + 'e').split('e');
           return +(number[0] + 'e' + (number[1] - precision));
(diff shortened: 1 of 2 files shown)


lodash.js
```diff
@@ -3699,8 +3699,12 @@
...
         precision = precision ? toInteger(precision) : 0;
         if (precision) {
-          precision = pow(10, precision);
-          return func(number * precision) / precision;
+          // Shift
+          number = number.toString().split('e');
+          number = func(+(number[0] + 'e' + (number[1] ? (+number[1] + precision) : precision)));
+          // Shift back
+          number = number.toString().split('e');
+          return +(number[0] + 'e' + (number[1] ? (+number[1] - precision) : -precision));
```
gibson042: This looks too verbose and formal for lodash... I was planning to compress the logic, e.g.:
```js
// Shift the decimal point with exponential operations
number = (number + 'e').split('e');	// force a non-trivial split
number = func(number[0] + 'e' + (+number[1] + precision));	// rely on ToNumber conversions
// Shift back
number = (number + 'e').split('e');
return +(number[0] + 'e' + (number[1] - precision));
```

gibson042:
Looks like you beat me to it. Which tests fail on the old code?

test/test.js
```diff
@@ -21071,6 +21071,95 @@
   }());
 
   /*--------------------------------------------------------------------------*/
+  QUnit.module('lodash.floor');
```
gibson042: There are already tests for these methods: https://github.com/tejokumar/lodash/blob/bug/1478/test/test.js#L15784

I don't think a new block is necessary, just a new assertion template.

tejokumar:
I have created a new assertion for failing scenario at https://github.com/tejokumar/lodash/blob/bug/1478/test/test.js#L15784 and removed new blocks

lodash.js
```diff
@@ -3699,8 +3699,13 @@
       return function(number, precision) {
         precision = precision ? toInteger(precision) : 0;
         if (precision) {
-          precision = pow(10, precision);
-          return func(number * precision) / precision;
+          // Shift decimal point with exponential operations
+          number = (number + 'e').split('e'); // Force a split
+          number = func(number[0] + 'e' + (+number[1] + precision)); // Use ToNumber conversions
```
gibson042: I don't know if the end-of-line comments add value here; they were just to explain my deviations from the MDN algorithm. Other than that, LGTM.

lodash.js
```diff
@@ -3699,8 +3699,13 @@
       return function(number, precision) {
         precision = precision ? toInteger(precision) : 0;
         if (precision) {
-          precision = pow(10, precision);
-          return func(number * precision) / precision;
+          // Shift decimal point with exponential operations
+          number = (number + 'e').split('e');
```
gibson042: `number + 'e'` should probably become `+number + 'e'`, otherwise this change will start accepting invalid input that looks like valid input followed by "e" (e.g., `"5e"` or `"7e55e…"`—solid test cases, FYI :wink:).

jdalton:
@tejokumar  have you [signed our CLA](http://dojofoundation.org/about/claForm)?

tejokumar:
@jdalton 
Yes, I have signed just now.

PR #1488, opened 2015-09-24
[docs] make priority of sources in _.defaults unambiguous

timruffles:
I always have to check this in the console as the documentation isn't explicit.

This updates both the example and the documentation to make priority of sources explicit.


jdalton:
Thanks!

PR #1489, opened 2015-09-25
Add `_.consists`.

jshanson7:
Hello!

I often find myself using this mixin to test if collections consist of the same members:

```javascript
_.consists([1,1,2,3], [3,1,2,1]);
// => true

_.consists([1,1,2,3], [3,1,2]);
// => false

_.consists('hello', ['o','l','l','e','h'], { a:'l', b:'h', c:'e', d:'o', e:'l' });
// => true
```

The common use case is to simply test if two arrays are equivalent ignoring order.  I've yet to find a terse way accomplish this using any combination of existing methods like `_.difference`, `_.intersection`, or `_.union` while handling:
 - collections with duplicate values
 - collections with non-primitive values

A couple more examples:

```javascript
var a = { a: 'a' };
var b = { b: 'b' };
var b2 = { b: 'b' };

_.consists([a, b, b], [b, a, b]);
// => true

_.consists([a, b, b], [a, b, b2]);
// => false
```

Not sure if there's interest in this, but thought I'd share it nonetheless.

[Here's a more real-world example of how `consists` might be useful.](https://gist.github.com/jshanson7/98083803bddd66d21a90)

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -6264,7 +6264,7 @@
     /*------------------------------------------------------------------------*/
 
     /**
-     * Tests if the members of two or more collections are equal. Uses
+     * Tests if two or more collections have the same members. Uses
      * [`SameValueZero`](http://ecma-international.org/ecma-262/6.0/#sec-samevaluezero).
      *
      * @static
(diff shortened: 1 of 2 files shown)


jdalton:
Thanks @jshanson7! In lodash v4 our `_.isEqual` method will support comparing sets and maps in an unordered fashion.

I think `_.consists` makes as a good mixin rather than core addition.

jshanson7:
`_.isEqual` is a good place for this, thanks!

PR #1497, opened 2015-09-29
Expose `baseHas` as `_.hasOwn`

benweet:
I was looking for the original `_.has` implementation from underscore. I just renamed the existing `baseHas`, added documentation and tests.

Please tell me if that's acceptable.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -1445,7 +1445,7 @@
      * `cloneDeepWith`, `cloneWith`, `deburr`, `endsWith`, `eq`, `escape`,
      * `escapeRegExp`, `every`, `find`, `findIndex`, `findKey`, `findLast`,
      * `findLastIndex`, `findLastKey`, `first`, `floor`, `get`, `gt`, `gte`,
-     * `has`, `hasIn`, `identity`, `includes`, `indexOf`, `inRange`, `isArguments`,
+     * `has`, `hasIn`, `hasOwn`, `identity`, `includes`, `indexOf`, `inRange`, `isArguments`,
      * `isArray`, `isArrayLike`, `isBoolean`, `isDate`, `isElement`, `isEmpty`,
      * `isEqual`, `isEqualWith`, `isError`, `isFinite` `isFunction`, `isInteger`,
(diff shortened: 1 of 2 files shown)


jdalton:
Thanks but I think our `_.has` meets and exceeds the Underscore use case.

benweet:
Ok but how if I want this behavior:

```js
_.has({'a.b': true},'a.b') // true
_.has({'a':{b: true}},'a.b') // false
```

jdalton:
I'm cool with deferring to the built-in:
```js
object.hasOwnProperty('a.b')
```

benweet:
Ok, that's part of a discussion I had today with a colleague. `object.hasOwnProperty('a.b')` is:
- slow, as it goes through the proto chain (better use `Object.prototype.hasOwnProperty.call(object, 'a.b')` which is not so nice)
- unsafe, as it can be overridden (as mentioned [in the underscore doc](http://underscorejs.org/#has))

I think the lodash `_.has` function is just not the same use case as the orginal one.

jdalton:
I think those are edge concerns. For general use in this scenario where you want to distinguish between plain and deep it should be fine or you could reference the internals via `require('lodash/internal/baseHas')` if needed.

benweet:
Well I could also create my own module, but that's precisely what I don't want to do.

jdalton:
Ah doh, you can also also provide a path as an array to get the behavior you want:
```js
_.has({'a':{b:true}}, ['a.b']) 
// => false
```

benweet:
Thank you but creating an array is not acceptable to me.

StreetStrider:
@benweet, demethodize std method
```js
js > var demethodize = require('demethodize')
js > var has = demethodize({}.hasOwnProperty)
js > has
[Function: dm]
js > has({x:1}, 'x')
true
js > has({x:1}, 'y')
false
```
Maybe this helps…

Btw this issue is a negative side of implementing dot-notation in accessors. I personally was negative to this (well, we have object-path, isn't we?). But this feature has good applications in something like `map` and family. And seemes that people use it.

benweet:
Ok bye.

benweet:
@StreetStrider thanks.

PR #1499, opened 2015-09-30
Ensure capitalization works for all-caps text

tquetano-r7:
.startCase() does not give the expected result with all caps input:

```
_.startCase("TEST"); // TEST, expected Test
```

Technically you could do the same conversion to your input in the call, but considering this is so computationally minimal and the result now aligns with all other inputs ... figured it was valuable.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -10504,6 +10504,7 @@
       if (!string) {
         return string;
       }
+      string = string.toLowerCase();
       if (reAdvSymbol.test(string)) {
         var strSymbols = stringToArray(string);
         return strSymbols[0].toUpperCase() + strSymbols.slice(1).join('');


jdalton:
Start case is a simplified variant of title case, where all words start with a capital letter. The casing rules don't include lowercasing the rest.

loris:
@jdalton As of `lodash 4.6.1`, it appears there was a non documented breaking change on that behavior:

loris:
So actually, `4.0.0` documented that change `Made _.capitalize uppercase the first character & lowercase the rest` and `startCase` is making use of `capitalize`

jdalton:
Yep. That's a bug.

jdalton:
@loris Patched https://github.com/lodash/lodash/commit/9735961090f6f4d1c404246c1fce55efe02fa2cb.

PR #1541, opened 2015-10-20
range: Produce more exact fractional results

### LINKED ISSUE #1539: Rounding error in _.range
Using version 3.10.1:

`_.range(0, 0.6, 0.1)`

returns

`[0, 0.1, 0.2, 0.30000000000000004, 0.4, 0.5]`

That's not right.

exogen:
Previously, `range` performed naïve summation, accumulating a larger and larger error on each iteration. This produced some annoying inexact values in the resulting array, but it also led to the final values of the array possibly being off by a significant amount.

This new method checks for fractional `step` values and attempts to generate the most exact results possible.

Fixes #1539.

/cc @zachhale @jdalton @jridgewell 

NOTE for anyone who didn't read the whole #1539 thread: the method used here provides *significantly* better results than attempting error compensation using Kahan's method or any others attempted (unless I just implemented error compensation incorrectly several times).

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -12480,24 +12480,42 @@
         end = +end || 0;
       }
       var n = nativeMax(nativeCeil((end - start) / (step || 1)), 0);
-      var factor = 1;
+      var scaleFactor = 1;
       // Only do extra work for fractional `step` values.
       if (step % 1 !== 0) {
+        // Separate sigificant digits from exponent, e.g. ['-4.005', '+4']
         var exponentParts = step.toExponential().split('e');
-        // If the step looks inexact already, and would result in a very large
-        // `factor`, don't even attempt to get more exact results.
-        if (exponentParts[0].length < 16) {
-          var integerStep = +(exponentParts[0].replace('.', ''));
-          // We know `factor` should be a power of ten. Do we need to potentially
-          // use `round` or `floor` here to ensure this float division doesn't
-          // screw that up?
-          factor = integerStep / step;
-          step = integerStep;
-          start *= factor;
+        // Remove the decimal to get a string version of `step` scaled up by
+        // some power of 10.
+        var stepDigits = exponentParts[0].replace('.', '');
+        // Let's find out how exact `step` is, to determine if it's worth
+        // scaling up in the first place. `toExponential` already gave us a
+        // hint at how exact it is, because the number of digits after the
+        // decimal point defaults to the number of digits necessary to
+        // represent the value uniquely. Even though 0.1, for example, can NOT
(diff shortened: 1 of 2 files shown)


lodash.js
```diff
@@ -12424,8 +12424,24 @@
...
+        // If the step looks inexact already, and would result in a very large
+        // `factor`, don't even attempt to get more exact results.
+        if (exponentParts[0].length < 16) {
+          var integerStep = +(exponentParts[0].replace('.', ''));
+          // We know `factor` should be a power of ten. Do we need to potentially
+          // use `round` or `floor` here to ensure this float division doesn't
+          // screw that up?
+          factor = integerStep / step;
+          step = integerStep;
+          start *= factor;
```
jdalton: I think this block needs some cleanup and solidify the logic (be sure it does what you want).

exogen:
Thanks for the extra push, @jdalton! After much digging around in JS numbers and testing, I think the latest update will increase confidence in the code. :)

exogen:
@gibson042 Updated with corrected float scaling, `+string` style instead of `parseInt`/`parseFloat` to match, and new test cases that failed before.

exogen:
One more note: there's still a tricky case that, intuitively, this method still seems to hiccup on (note: so does lodash's current method and a few others I've tried) – but I think might actually be the technically-correct-in-floating-point-land result:

```
> _.range(0, 7/3, 1/3);
[ 0,
  0.3333333333333333,
  0.6666666666666666,
  1,
  1.3333333333333333,
  1.6666666666666665,
  2,
  2.333333333333333 ]
```

I'm not concerned with intermediate inexact terms – like `5/3` != `5 * (1/3)` – that seems unavoidable whether we sum or multiply, even with error correction. But rather what that leads to: we've picked up an extra result – intuitively there should be 7 items, not 8. It's because of `n`, which after calculating `(7/3) / (1/3)` to be `7.000000000000001` and running `ceil` on that, tells us we should produce 8 results.

Even if we compared the last item to `end`, it'd still tell us everything is OK, because:
```
> 7 * (1/3) < 7/3
true
```
So technically, that last item "fits" within the requested range, and the user should have just specified `7 * (1/3)` as their `end`.

One option that fixes this specific case, but introduces a subtraction (and thus potential error): calculate a sentinel, `end - step`, and on the last iteration, if a previous result has already met or exceeded the sentinel, don't append the last result.

jdalton:
I'm going to punt on this for now as it creeps into other methods and complicates things for what's really an edge case concern. I'm ok tackling it in the rounding methods because we have a param explicitly for `precision` but elsehwere I'm for keeping it simple.

Issue #1552, opened 2015-10-22
Enhancement: Extending Math and Number categories for commonly used methods.


Xotic750:
I would like to see a few more common Math and Number methods available in lodash. I'm not talking about creating a Math or Number library, just having some commonly used stuff. And I'm not talking about fixing IEEE 754 and implementing arbitrary precision solutions, the methods would be subject to its limits. I initiated a small conversation in https://github.com/lodash/lodash/issues/1550 and now I wanted to extend on that specific conversation and review Math and Number a little more generally. Of course I would be willing to help out with all or any of these. The basics I would like to see are:
```
/**
 * Same as Math.sign can defer to native if available.
 *
 * @see https://en.wikipedia.org/wiki/Sign_function
 * @see http://www.ecma-international.org/ecma-262/6.0/#sec-math.sign
 */
function sgn(value) {
    value = Number(value);
    return value && (value >= 0 || -1);
}

/**
 * The value's direction, towards -Infinity (-1) or Infinity (1).
 * Allows isPositive and isNegative tests in one function.
 *
 * @see https://github.com/lodash/lodash/issues/1550
 */
function dir(value) {
  value = sgn(value);
  if (value) {
    value = value < 0 ? -1 : 1;
  } else if (value === value) {
    value = 1 / value !== Infinity ? -1 : 1;
  }
  return value;
}

/**
 * Same as Math.trunc can defer to native if available.
 *`
 * The name trunc is already used in lodash.
 *
 * @see https://en.wikipedia.org/wiki/Truncation
 * @see http://www.ecma-international.org/ecma-262/6.0/#sec-math.trunc
 */
function trc(value, precision) {
  value = Number(value);
  return value < 0 ? _.ceil(value, precision) : _.floor(value, precision);
}

/**
 * The fractional part.
 *
 * @see https://en.wikipedia.org/wiki/Floor_and_ceiling_functions
 * @see https://en.wikipedia.org/wiki/Sawtooth_wave
 */
function saw(value, precision) {
  value = Number(value);
  return value - _.floor(value, precision);
}

/**
 * Truncated division
 * In Javascript the % operator but more generally a - n * Math.trunc(a / n)
 *
 * @see https://en.wikipedia.org/wiki/Modulo_operation
 * @see http://www.ecma-international.org/ecma-262/6.0/#sec-applying-the-mod-operator
 */
function mod(a, n, precision) {
  a = Number(a);
  n = Number(n);
  return a - n * trc(a / n, precision);
}

[...]

PR #1555, opened 2015-10-22
Added `toSafeInteger` and tests as discussed …

Xotic750:
https://github.com/lodash/lodash/issues/1552#issuecomment-150225955

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -54,7 +54,6 @@
   /** Used as references for various `Number` constants. */
   var INFINITY = 1 / 0,
       MAX_SAFE_INTEGER = 9007199254740991,
-      MIN_SAFE_INTEGER = -MAX_SAFE_INTEGER,
       MAX_INTEGER = 1e308;
 
   /** Used as references for the maximum length and index of an array. */
@@ -9572,7 +9571,7 @@
      * // => false
      */
     function isSafeInteger(value) {
-      return isInteger(value) && value >= MIN_SAFE_INTEGER && value <= MAX_SAFE_INTEGER;
+      return isInteger(value) && value >= -MAX_SAFE_INTEGER && value <= MAX_SAFE_INTEGER;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -19083,6 +19083,29 @@
...
+  QUnit.module('lodash.toSafeInteger');
+
+  (function() {
+    QUnit.test('should convert values to safe integers', function(assert) {
+      assert.expect(6);
+
+      assert.strictEqual(_.toSafeInteger(-5.6), -5);
+      assert.strictEqual(_.toSafeInteger('5.6'), 5);
+      assert.strictEqual(_.toSafeInteger(), 0);
+      assert.strictEqual(_.toSafeInteger(NaN), 0);
```
megawac: Can the rationalization of this case be explained?

Xotic750: Which particular case?

megawac: NaN ~= 0

Xotic750: toInteger is based on http://www.ecma-international.org/ecma-262/6.0/#sec-tointeger and toSafeInteger is based on toInteger

megawac: If we're following that spec, why do we want to deviate on infinity and
-infinity?

Xotic750: I can say that for me this is a commonly used function. I don't think lodash is following that spec per say but like in the comments https://github.com/lodash/lodash/blob/master/lodash.js#L9720, it is loosely basing some of it's methods on that spec, why not?

jdalton: > If we're following that spec, why do we want to deviate on infinity and
-infinity? For safeness?

The internal `ToInteger` doesn't enforce that the return value is an integer. It gets away with it because it's internal and not exposed. Our `_.toInteger` on the other hand is exposed and anything named `toThing` should always return some `thing`.

lodash.js
```diff
@@ -54,6 +54,7 @@
   /** Used as references for various `Number` constants. */
   var INFINITY = 1 / 0,
       MAX_SAFE_INTEGER = 9007199254740991,
+      MIN_SAFE_INTEGER = -MAX_SAFE_INTEGER,
```
jdalton: No need for `MIN_SAFE_INTEGER` we can just stick with the sign.

Xotic750: Ok, I'll remove that micro optimisation.

jdalton:
Thank you @Xotic750!

Xotic750:
@jdalton  No problem, if you want PRs for any of the others at some point then just shout and I'll be happy to help. If I come across anything else then I will float them with you, I still haven't looked through my string collection. ;)

PR #1571, opened 2015-10-26
Add function to make chains callable with a new value

fdegiuli:
Add `_().callable()`, which returns a function that executes the actions
of the chain `callable()` was called on its argument.

This enables the passing of chains to functions which take an interatee.

`callable()` works by cloning the wrapper and replacing the first `__wrapped__`
property which is not itself a wrapper with the argument passed to
the function. It then calls `value()` on the resulting wrapper object
and returns the result.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -6736,6 +6736,39 @@
       return baseWrapperValue(this.__wrapped__, this.__actions__);
     }
 
+    /**
+     * Creates a function which executes the chained sequence on its argument.
+     * The original wrapped value of the chain is ignored.
+     *
+     * **Note:** Many `lodash` methods (`sum`, `first`, etc...) are not chainable by default. 
+     * It's good practice to always use `_.prototype.chain` with `_.prototype.callable`,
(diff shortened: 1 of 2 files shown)


jdalton:
This is an interesting idea. I think it could be implemented by way of [_#plant](https://lodash.com/docs#prototype-plant) and you could make it auto `chain()`.

I'm not sure it's approp for core but dig the idea. Do you have a use case?

fdegiuli:
Huh. Totally missed _#plant. I couldn't figure out a way to force `chain()` after the fact, since you lose all the wrappers when you call a non-chainable function.

falsyvalues:
Interesting idea :+1:

fdegiuli:
I don't want to beat a dead horse, but I keep running into places where having this would be useful. (Especially when using Promises.)  The problem is that I haven't found a really clean way to implement it from the outside.  Should I make an issue to gauge whether there is any other interest?

jdalton:
> Should I make an issue to gauge whether there is any other interest?

This is that issue.

PR #1576, opened 2015-10-28
Preserve sign of `0`.

Xotic750:
Part of https://github.com/lodash/lodash/issues/1574

Issue #1574, opened 2015-10-27
Enhancement: toString and toNumber, possibly isBinary and isOctal


Xotic750:
```
// Regexs not taking into account leading or trailing whitespace.
var reBinary = /^0b[0..1]+$/i,
  reOctal = /^0o[0..7]+$/i;

// Some older browsers don't convert null and undefined to humanised strings.
// Opera 9 (I think) doesn't like the use of toString as a name.
// Symbol should convert to value, but many not up to spec yet, so '' is
// sensible for exposed toString.
// Javascript toString doesn't preserve sign of 0.
function toStr(value) {
  if (value === 0) {
    return 1 / value !== Infinity ? '-0' : '0';
  }
  var type = typeof value;
  // Exit early for strings to avoid a performance hit in some environments.
  if (type === 'string') {
    return value;
  }
  return type !== 'symbol' ? value + '' : '';
}

// Expose or internal only?
// Should leading and trailing whitespaces be accounted for, i.e. '   0b0101   ', probably?
function isBinary(value) {
  return reBinary.test(toStr(value));
}

// Expose or internal only?
// Should leading and trailing whitespaces be accounted for, i.e. '   0b0101   ', probably?
function isOctal(value) {
  return reOctal.test(toStr(value));
}

// Internal use.
// I believe there is also a hex conversion bug in some old browsers (not sure),
// which could also be fixed/avoided here.
// https://github.com/es-shims/es5-shim/blob/master/es5-shim.js#L1600
function str2num(value) {
  value = toStr(value);
  // Should leading and trailing whitespaces be accounted for, i.e. '   0b0101   ', probably?
  if (isBinary(value)) {
    // returns or fallthrough, I don't know what your preference is?
    value = parseInt(value.slice(2), 2);
  } else if (isOctal(value)) {
    // returns or fallthrough, I don't know what your preference is?
    value = parseInt(value.slice(2), 8);
  }
  return +value;
}

// Number() gives 0, but Number(undefined) and +(undefined) gives NaN.
// The ES6 spec says throw a TypeError for symbols but NaN seems more
// appropriate for an exposed toNumber method.
// Spec tries to convert by valueOf then toString.
// At minimum Opera 9 throws if object has valueOf that is not a function.
// Many host objects are `Object` objects that can coerce to strings
// despite having improperly defined `toString` methods.
function toNum(value) {
  if (value === null) {
    return 0;
  }
  var type = typeof value;
  if (value === 'undefined' || type === 'symbol') {
[...]

PR #1577, opened 2015-10-28
Add `toNumber` and cleanup `toString`. Plus a wide range of tests.

Xotic750:
https://github.com/lodash/lodash/issues/1574
https://github.com/lodash/lodash/pull/1576#discussion_r43206357

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -712,7 +712,22 @@
   }
 
   /**
-   * The base implementation of `_.reduce` and `_.reduceRight` without support
+   * The base implementation of `_.pairs` and `_.pairsIn` which creates an array
+   * of key-value pairs for `object` corresponding to the property names of `props`.
+   *
+   * @private
+   * @param {Object} object The object to query.
+   * @param {Array} props The property names to get values for.
+   * @returns {Object} Returns the new array of key-value pairs.
+   */
+  function basePairs(object, props) {
+    return arrayMap(props, function(key) {
+      return [key, object[key]];
+    });
+  }
+
+  /**
+   * The base implementation of `_.reduce` and `_.reduceRight`, without support
    * for callback shorthands, which iterates over `collection` using the provided
    * `eachFunc`.
    *
@@ -10174,8 +10189,8 @@
     }
 
     /**
-     * Creates an array of function property names from all enumerable properties,
-     * own and inherited, of `object`.
+     * Creates an array of function property names from own enumerable properties
+     * of `object`.
      *
      * @static
      * @memberOf _
@@ -10184,10 +10199,42 @@
      * @returns {Array} Returns the new array of property names.
      * @example
      *
-     * _.functions(_);
-     * // => ['after', 'ary', 'assign', ...]
+     * function Foo() {
+     *   this.a = _.constant('a');
+     *   this.b = _.constant('b');
+     * }
+     *
+     * Foo.prototype.c = _.constant('c');
+     *
+     * _.functions(new Foo);
+     * // => ['a', 'b']
      */
     function functions(object) {
+      return object == null ? [] : baseFunctions(object, keys(object));
+    }
+
+    /**
+     * Creates an array of function property names from own and inherited
+     * enumerable properties of `object`.
+     *
+     * @static
+     * @memberOf _
+     * @category Object
+     * @param {Object} object The object to inspect.
+     * @returns {Array} Returns the new array of property names.
+     * @example
+     *
+     * function Foo() {
+     *   this.a = _.constant('a');
+     *   this.b = _.constant('b');
+     * }
+     *
+     * Foo.prototype.c = _.constant('c');
+     *
+     * _.functionsIn(new Foo);
+     * // => ['a', 'b', 'c']
+     */
+    function functionsIn(object) {
       return object == null ? [] : baseFunctions(object, keysIn(object));
     }
 
@@ -10593,8 +10640,7 @@
     }
 
     /**
-     * Creates an array of the key-value pairs for `object`,
-     * e.g. `[[key1, value1], [key2, value2]]`.
+     * Creates an array of own enumerable key-value pairs for `object`.
      *
      * @static
      * @memberOf _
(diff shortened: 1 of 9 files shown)


Xotic750:
Ok, something funky with node v0.8, but don't have the time to investigate just now, must sleep.

Xotic750:
That should cover most of the basic use cases. There are plenty more that could be added to catch other bugs, like I mentioned with Opera where an object has valueOf defined but it's not a function.

Xotic750:
To catch even more sting issues, the type value === 'string' could be extended to String objects by using isString(value) instead.

Xotic750:
Finally (for now), do we keep this early return for number types?
```
    function toNumber(value) {
      // Possibly not required, not sure if this improves performance or not?
      if (typeof value === 'number') {
        return value;
      }
```

Xotic750:
So, node v0.8 has a problem with this case again converting bad hex strings.

Xotic750:
Now need to find a simple fixes without implementing a full toPrimitive shim.

lodash.js
```diff
@@ -9771,6 +9781,37 @@
...
+      }
+      if (isString(value)) {
+        var trimmed = _.trim(value);
+        if (reIsBinary.test(trimmed)) {
+          return parseInt(trimmed.slice(2), 2);
+        }
+        if (reIsOctal.test(trimmed)) {
+          return parseInt(trimmed.slice(2), 8);
+        }
+        if (reIsBadHex.test(trimmed)) {
```
jdalton: Shouldn't bad hex fall out naturally as `NaN` from the `+trimmed` below?

lodash.js
```diff
@@ -9771,6 +9781,37 @@
...
+          return NaN;
+        }
+        return +trimmed;
```
jdalton: Is the `return `+trimmed` needed at this point could it just rely on `+value`?

test/test.js
```diff
@@ -19423,6 +19425,386 @@
...
+      assert.deepEqual(_.toNumber(' +0X1A2B3X '), NaN);
+      assert.deepEqual(_.toNumber(' -0X1A2B3X '), NaN);
+      assert.deepEqual(_.toNumber(' -0b '), NaN);
+      assert.deepEqual(_.toNumber(' -0o '), NaN);
+      assert.deepEqual(_.toNumber(' -0x '), NaN);
+    });
+
+    QUnit.test('should convert string objects with whitespace accurately', function(assert) {
+      assert.expect(75);
+
```
jdalton: This seems like a large amount of tests for this method. Can you go through this and weed out anything extra.

test/test.js
```diff
@@ -19423,6 +19425,386 @@
...
+    })
+
+    QUnit.test('other objects', function(assert) {
+      assert.expect(14);
+
+      assert.deepEqual(_.toNumber({}), NaN);
+      assert.deepEqual(_.toNumber({
+        valueOf: '1.1'
+      }), NaN);
+      assert.strictEqual(_.toNumber({
```
jdalton: Same here way too many tests for this, confirm `toString` behavior and that's enough.

lodash.js
Xotic750: I will look at improving the flow.

lodash.js
Xotic750: I will look at improving the flow.

test/test.js
Xotic750: I will do, just making sure that I catch anything in common use.

test/test.js
Xotic750: I should be able to roll it into something much smaller once I make it robust for the environments that you are testing on Travis.

Xotic750:
All test cases are now satisfied but need to tackle the test reduction.

jdalton:
It may be easier to do a quick test for the basics of `+foo` then dig into the uniq bits of the method to test. We don't need to test all of `+foo` because our assumption is the language is stable enough.

Xotic750:
Looking at it now, and can probably do that. I think the values chosen for testing are a fairly good representation and limits and types, and would like to keep them as are. I'll see what I an come up with and then you can poke me some more. :P

Xotic750:
That's some improvement but clearly more can be done.

Xotic750:
How does that look, it has reduced the test count by several hundred?

Xotic750:
I would also like to see _.toInteger use this method so that it benefits from the binary and octal parsing.

jdalton:
Can you rebase this PR.

Xotic750:
I've created an optimised version, instead of calling lodash methods. I don't know if you'd prefer this instead, as it looks like I need to rebase again?
```
    function toNumber(value) {
      if (value != null) {
        var type = typeof value;
        if (type == 'number') {
          return value;
        }
        if (objToString.call(value) != numberTag) {
          if ((type == 'object' || type == 'function') && isFunction(value.valueOf)) {
            value = value.valueOf.call(value);
            type = typeof value;
            if (type == 'number') {
              return value;
            }
          }
          if (!!value && (type == 'object' || type == 'function')) {
            value = toString(value);
          }
          if (!!value && typeof value == 'string') {
            value = trim(value);
            if (value) {
              if (reIsBinary.test(value)) {
                return nativeParseInt(value.slice(2), 2);
              }
              if (reIsOctal.test(value)) {
                return nativeParseInt(value.slice(2), 8);
              }
              if (reIsBadHex.test(value)) {
                return NaN;
              }
            }
          }
        }
      }
      return +value;
    }
```

jdalton:
Not a fan of that. Less is more.

lodash.js
```diff
@@ -9903,6 +9931,40 @@
...
+            return nativeParseInt(value.slice(2), 8);
+          }
+          if (reIsBadHex.test(value)) {
+            return NaN;
+          }
+        }
+      }
+      return +value;
+    }
+
```
jdalton: This can be simplified:
```js
function toNumber(value) {
  var result = +value;
  if (typeof value == 'number' || !isString(value)) {
    return result;
  }
  value = trim(value);
  if (reIsBinary.test(value)) {
    return nativeParseInt(value.slice(2), 2);
  }
  if (reIsOctal.test(value)) {
    return nativeParseInt(value.slice(2), 8);
  }
  return result;
}
```

Xotic750:
Ok, I'll just rebase it.

jdalton:
Updated to even less code.

lodash.js
Xotic750: That is going to fail on at least node v0.8 I believe. But that should just be a case of adding the bad hex test back in. It's nice to see where your experience pays off wrt simplification. Let me change it and we can see.

jdalton: Ah cool, verified 0.8 does not return `NaN` :P

Xotic750:
I don't know if it is just me, but git doesn't seem very good at auto merging simple upstream changes?

jdalton:
It'll be easier in your branch to copy the lodash.js and test.js to another location then.
```bash
git reset --hard c142cfc81dce4c3cbd3ff4ceb33e13df13062be9
# add your lodash.js and test.js back
git commit -m "Add _.isNumber."
git push origin +master
```

Xotic750:
I guess I'll get better at it, at the end of the day I'm just a self taught enthusiast. ;)

jdalton:
Closed via https://github.com/lodash/lodash/commit/0539461ab25c6b6cd3d47b7ac33a937a76271846.

PR #1486, opened 2015-09-24
doc: typo occurence to occurrence

reggi:

PR #1580, opened 2015-10-29
Update README.md

nickbalestra:
Misspelling fixed


KenanY:
I think that file is generated from `lodash.js`, and that typo was already fixed in https://github.com/lodash/lodash/pull/1486.

jdalton:
Thanks @nickbalestra! This is already fixed in the lodash.js source.

Issue #1590, opened 2015-10-30
Enhancement: expose isIndex, isLength and complimentary toLength


Xotic750:
Change `reIsUint` to pick up number string literals beginning with `0`.
Change `isLength`, add length argument, use other lodash methods
Add `toLength` as it doesn't exist.
These are all useful methods that I find myself using and lodash mostly has them internally.
```
var MAX_SAFE_INTEGER = Number.MAX_SAFE_INTEGER || Math.pow(2, 53) - 1;

/** Used to detect unsigned integer values. */
var reIsUint = /^(\d|[1-9]\d+)$/;

/**
 * Checks if `value` is a valid array-like index.
 *
 * @param {*} value The value to check.
 * @param {number} [length=MAX_SAFE_INTEGER] The upper bounds of a valid index.
 * @returns {boolean} Returns `true` if `value` is a valid index, else `false`.
 */
function isIndex(value, length) {
  value = (typeof value == 'number' || reIsUint.test(value)) ? +value : -1;
  length = length == null ? MAX_SAFE_INTEGER : length;
  return value > -1 && value % 1 == 0 && value < length;
}

/**
 * Checks if `value` is a valid array-like length.
 *
 * **Note:** This function is based on [`ToLength`](http://ecma-international.org/ecma-262/6.0/#sec-tolength).
 *
 * @param {*} value The value to check.
 * @param {number} [length=MAX_SAFE_INTEGER] The upper bounds of a valid length.
 * @returns {boolean} Returns `true` if `value` is a valid length, else `false`.
 */
function isLength(value, length) {
  value = (typeof value == 'number' || reIsUint.test(value)) ? +value : -1;
  length = length == null ? MAX_SAFE_INTEGER : length;
  return value > -1 && value % 1 == 0 && value <= length;
}

/**
 * The abstract operation ToLength converts its argument to an integer
 * suitable for use as the length of an array-like object.
 *
 * **Note:** This function is based on [`ToLength`](http://ecma-international.org/ecma-262/6.0/#sec-tolength).
 *
 * @param {*} value The object to be converted to a length.
 * @param {number} [length=MAX_SAFE_INTEGER] The upper bounds of a valid length.
 * @return {number} If len <= 0 then 0 else if len is +INFINITY then 2^53-1
 *                  else min(len, 2^53-1).
 */
function toLength(value, length) {
  length = length == null ? MAX_SAFE_INTEGER : length;
  return _.clamp(_.toInteger(value), 0, length);
}
```

[...]

PR #1592, opened 2015-10-30
Add `toLength` and basic tests.

Xotic750:
https://github.com/lodash/lodash/issues/1590

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -1475,9 +1475,9 @@
      * `random`, `reduce`, `reduceRight`, `repeat`, `result`, `round`, `runInContext`,
      * `sample`, `shift`, `size`, `snakeCase`, `some`, `sortedIndex`, `sortedIndexBy`,
      * `sortedLastIndex`, `sortedLastIndexBy`, `startCase`, `startsWith`, `sum`,
-     * `sumBy`, `template`, `toLower`, `toInteger`, `toSafeInteger`, `toString`,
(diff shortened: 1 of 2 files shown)


lodash.js
```diff
@@ -9833,6 +9833,25 @@
...
+     *
+     * @static
+     * @memberOf _
+     * @category Lang
+     * @param {*} value The object to be converted to a length.
+     * @param {number} [length=MAX_SAFE_INTEGER] The upper bounds of a valid length.
+     * @return {number} If len <= 0 then 0 else if len is +INFINITY then 2^53-1
+     *                  else min(len, 2^53-1).
+     */
+    function toLength(value, length) {
```
jdalton: Since this isn't used internally you can remove the `length` param and just ensure the value is clamped to the `MAX_ARRAY_LENGTH` and greater than `-1`.

Xotic750: You don't think it would be useful to support MAX_SAFE_INTEGER for ES6 arrays and strings and be configurable in older ES5 environments for MAX_ARRAY_LENGTH or even less for strings?

jdalton: > You don't think it would be useful to support MAX_SAFE_INTEGER for ES6 arrays and strings and be configurable in older ES5 environments for MAX_ARRAY_LENGTH or even less for strings?

No, `isLength` is associated with array-like lengths so follows the rules of `isArrayLike`.

PR #1594, opened 2015-10-30
Improve `reIsUint` to filter strings like `0001 and add tests.

Xotic750:
https://github.com/lodash/lodash/issues/1590

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -132,7 +132,7 @@
   var reIsHostCtor = /^\[object .+?Constructor\]$/;
 
   /** Used to detect unsigned integer values. */
-  var reIsUint = /^(\d|[1-9]\d+)$/;
+  var reIsUint = /^(?:\d|[1-9]\d+)$/
 
   /** Used to match latin-1 supplementary letters (excluding mathematical operators). */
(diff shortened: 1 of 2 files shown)


lodash.js
```diff
@@ -132,7 +132,7 @@
   var reIsHostCtor = /^\[object .+?Constructor\]$/;
 
   /** Used to detect unsigned integer values. */
-  var reIsUint = /^\d+$/;
+  var reIsUint = /^(\d|[1-9]\d+)$/;
```
jdalton: You can use non capturing group please:
```js
/^(?:\d|[1-9]\d+)$/
```

lodash.js
```diff
@@ -132,7 +132,7 @@
   var reIsHostCtor = /^\[object .+?Constructor\]$/;
 
   /** Used to detect unsigned integer values. */
-  var reIsUint = /^\d+$/;
+  var reIsUint = /^(?:\d|[1-9]\d+)$/
```
jdalton: Since you're contributing a bit more I'm going to get a bit more nit picky (I'll easy into it).
Can you run `npm run style` and see if it warns of any style issues.

Xotic750: I had been running it, but I didn't on this change. :)

PR #1596, opened 2015-11-01
Add tests to check `toString` behaviour.

Xotic750:
http://jsfiddle.net/myahL788/1/
`String()` calls toString first, `value + ''` calls valueOf first.
Results may not be what are expected.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -19652,6 +19652,33 @@
         skipTest(assert);
       }
     });
+
+    QUnit.test('should return the `toString` result', function(assert) {
+      assert.expect(1);
+
+      var answer = {
+        valueOf: function () {
+          return true;
+        },
+        toString: function () {
+          return 'correct'
+        }
+      };
+
+      assert.strictEqual(_.toString(answer), 'correct');
+    });
+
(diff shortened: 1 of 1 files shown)


jdalton:
This is testing the language which I want to avoid.

Xotic750:
I just added the tests to indicate behaviour, not necessarily for merging. My point is that `toString` by name has an expectation associated with it and so far we have 3 anomalies. First, we are preserving sign of zero which Javascript does not, and second the order to get a string is valueOf then toString rather than the most likely expected toString then valueOf. And third, an empty string is returned for `null` and `undefined`. Nothing particularly wrong with our choices, documentation is the key.
```
     * Converts `value` to a string if it's not one.
     * An empty string is returned for `null` and `undefined` values.
```
Seems a little lacking at present?

jdalton:
Our `_.toString` has the normal behavior of coercing a value to a string as far as `valueOf` and `toString` invocation is concerned since we use do `value + ''` under the hood. If you'd like to expand the doc note on preserving the sign of `0` I'd accept a PR for that.

Xotic750:
I can add that. I just wanted to check/clarify.

Issue #1602, opened 2015-11-02
Enhancement: Identify implicit type conversions that would benefit from `toNumber` consistency


Xotic750:
An example may be
```
    function baseRandom(min, max) {
      return min + nativeFloor(nativeRandom() * (max - min + 1));
    }
```
as
```
    function baseRandom(min, max) {
      min = toNumber(min);
      max = toNumber(max);
      return min + nativeFloor(nativeRandom() * (max - min + 1));
    }
```
(Ok, `baseRandom` is an internal, it is just an example of meaning.)
therefore type conversion becomes more consistent (are not left to environmental chance) and benefit from the ES6 like string conversions implemented in `toNumber`.

jdalton:
The base methods is not the place for type coercion as they should already have the correct type of value passed to it.

Xotic750:
I agree. But if you would like, then I'd be willing to look through and submit individual PRs for approval.

jdalton:
Let's work through where it would be needed here and take it out in 1 PR.

Xotic750:
Ok, I'll begin making a list of possible candidates.

Xotic750:
I haven't gone very far yet but what I have noticed so far is that `baseSlice` and `baseFill` are performing type conversion rather than the methods that are using them.

Xotic750:
These are potential candidates for `toNumber`
createRound
clamp
inRange
random
range
add
sum
sumBy
(baseSum - but it was agreed base isn't the correct place)

I also saw areas where I thought, `toInteger` seems appropriate here, or `toSafeInteger` or `toLength` seem more appropriate there. Just a small selection of examples may be.

debounce
delay
throttle

 The other 2 major areas that seemed strange to me were (as mentioned) `baseSlice` and `baseFill`.

jdalton:
> createRound
clamp
inRange
random
range
add
sum
sumBy

Yes, `_.toNumber`!

> debounce
delay
throttle

Yes, `_.toInteger`!

>  baseFill

Lets move the `toInteger` use to `fill`.

Lets not touch `baseSlice` yet.

Xotic750:
Ok, I'll start working my way through them.

Xotic750:
I'm a little unsure about the intention with the method `add`, as the docs state
>      * @param {number} augend The first number to add.
     * @param {number} addend The second number to add.
     * @returns {number} Returns the sum.

But the code and the tests show something altogether different.

[...]

PR #1610, opened 2015-11-03
Enhance `debounce` to handle binary and octal strings and ms times to integers.

Xotic750:
Ref https://github.com/lodash/lodash/issues/1602

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -8145,8 +8145,7 @@
       if (typeof func != 'function') {
         throw new TypeError(FUNC_ERROR_TEXT);
       }
-      wait = toInteger(wait);
-      wait = wait < 0 ? 0 : wait;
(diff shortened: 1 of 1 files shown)


lodash.js
```diff
@@ -8145,10 +8145,11 @@
       if (typeof func != 'function') {
         throw new TypeError(FUNC_ERROR_TEXT);
       }
-      wait = wait < 0 ? 0 : (+wait || 0);
+      wait = toInteger(wait);
+      wait = wait < 0 ? 0 : wait;
```
jdalton: clamp?

Xotic750: I thought about it, and thought minimal change.I'll change it.

Xotic750:
As throttle uses debounce, do we need to do conversion there or should we just rely on debounce?

PR #1612, opened 2015-11-03
Enhance `delay` with `toInteger` and `clamp` to use binary and octal strings.

Xotic750:
Ref https://github.com/lodash/lodash/issues/1602


jdalton:
This isn't needed as it is handled by the built-in `setTimeout`.

jdalton:
Actually I'm cool with the `toInteger` use though. I'll take care of the clamp removal.

PR #1619, opened 2015-11-06
adding math subtract

J3remyD:


jdalton:
Hi @J3remyD! Have you signed [our CLA](http://dojofoundation.org/about/claForm)?

J3remyD:
Hi @jdalton ,
yes indeed I have

jdalton:
Opinions on the name? Would `minus` do?

StreetStrider:
@jdalton, maybe `_.sub`?
add ↔ sub. `subtract` is also ok, `sub` just shorter. Have looked in Ramda, they called it [subtract](http://ramdajs.com/docs/#subtract) as in PR.

jdalton:
I want to avoid `sub` as that can be seen as a shorthand for [substitute](http://api.prototypejs.org/language/String/prototype/sub/).

jdalton:
@J3remyD Do you have a use case for `_.subtract`?

J3remyD:
@jdalton I agree with you that sub could sometimes be misleading even though it's shorter and sounds great ; that's why I named it thinking with the same pattern like add is for addition I figured subtract would be for subtraction.

PR #1625, opened 2015-11-09
Support deepPaths as keynames for zipObject

### LINKED ISSUE #1151: deep path support for zipObject
The deep path support in 3.7.0 is exciting! After updating, my first use case was for zipping an array into an object, where some of the property names use bracket/subscript notation. (More specifically, my use case is a quick one-liner to parse a query string into an object.)

```js
_([
  [ "a[b]", 1 ],
  [ "a.c", 2 ]
]).object().value() //=>  { a: { b:1, c:2 } }
```
There are a couple of questions as to how this would be handled for `zipObject`.

Clearly when zipping objects, the deep path parsing would need to be optional. But we also have a number of different sub-features:

1. handling subscript notation with string key: `"a[b]"`
2. handling dot notation: `"a.b"`
3. handling empty subscript (to result in an array property): `"b[]"`
4. handling explicit numerical subscript: (result in array? or object? object with string keys or numbers?): `"b[0]"`

And there's the interplay with `_.pairs` to think about, as well.

jasonkarns:
Closes #1151

This is implemented by simply using baseSet within the body of zipObject
when building up the object

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -6671,9 +6671,9 @@
       while (++index < length) {
         var key = props[index];
         if (values) {
-          result[key] = values[index];
+          baseSet(result, key, values[index]);
         } else if (key) {
-          result[key[0]] = key[1];
+          baseSet(result, key[0], key[1]);
         }
       }
       return result;
--- a/test/test.js
+++ b/test/test.js
@@ -21236,6 +21236,24 @@
       assert.deepEqual(actual, expected);
     });
 
+    QUnit.test('should accept deep-path key: subscript with string', function(assert) {
+      assert.expect(1);
+
+      assert.deepEqual(_.zipObject([['a[b]', 1]]), { 'a': { 'b': 1 }});
+    });
+
+    QUnit.test('should accept deep-path key: subscript with index', function(assert) {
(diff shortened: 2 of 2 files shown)


jdalton:
Hey @jasonkarns! Have you signed [our CLA](http://dojofoundation.org/about/claForm)?

jasonkarns:
signed

jdalton:
:tada:

jasonkarns:
It's worth mentioning that with this feature, lodash can reify an object from a querystring in a one-liner. (For common querystring structures, anyway. The array syntax can't use repeated-keys or the empty bracket form.)

jdalton:
Neat!

aliak00:
Hi, this behaviour breaks a few situations where a dot is part of your key, unexpected results ensue. Is that intentional to have that kind of behaviour or an oversight ... or maybe just preferred?

Case I actually have is mapping files in a directory to their fs.Stats object. File names can have dots, and boom.

Any recommended work around?

jdalton:
This is resolved in the edge master branch with _.zipObjectDeep.

PR #1630, opened 2015-11-11
add parseBool function

DavidMoritz:
_.parseBool (or _.parseBoolean) is a utility designed to receive almost any variable and convert it into a proper boolean.

The most common use-case will be for converting strings such as "yes"/"no" or "true"/"false" into actual booleans.

In the fringe case that a boolean cannot be resolved, the string "NaB" is returned as JS doesn't have an absent boolean equivalent.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -11857,6 +11857,62 @@
     }
 
     /**
+     * Converts `any` to a boolean. If a boolean is not able to be determined, it
+     * returns the string `"NaB"`. This is unfortunate being that there is no js equivalent
+     * to `NaN` for booleans.
+     *
+     * @static
+     * @memberOf _
+     * @category Any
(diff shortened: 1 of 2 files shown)


jdalton:
Thanks @DavidMoritz! This looks a bit too specialized for lodash but would make a nice mixin.

DavidMoritz:
Thanks @jdalton! Can you provide a link to either an example of a PR that added a mixin or a guide for how do add a mixin?

jdalton:
Ah it wouldn't be a PR to add a mixin. It would be for you to create a mixin using [_.mixin](https://lodash.com/docs#mixin).

You can use [lodash-deep](https://github.com/marklagendijk/lodash-deep/blob/master/lodash-deep.js) as an example.

PR #1633, opened 2015-11-12
add enrich array function

J3remyD:
This function enrich every element of a provided array by concatenating
an input parameter value

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -5431,6 +5431,39 @@
         ? baseWhile(array, getIteratee(predicate, 3), true)
         : [];
     }
+	
+	 /**
+     * Creates an enriched array in which all values were enriched with the parameter value provided
+     *
+     * @static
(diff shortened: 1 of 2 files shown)


jdalton:
Thanks @J3remyD. In v4 you'll be able to do `_.map(arr, _.partial(_.add, prefix))`.

J3remyD:
Oh I didn't see that coming :) ! Cool !

J3remyD:
@jdalton would it be possible to combine the same with _.map(arr, _.partial( _.remove,prefix)) and does it handle suffixes to ??

jdalton:
You could could partial a combo of `_.bind` plus a `String#replace` method but it would end up looking pretty gross compared to just writing out a traditional `iteratee`.

J3remyD:
Indeed, thanks.

PR #1661, opened 2015-11-25
Add nonempty

aguspina:


aguspina:
This function is similar to `nonEmpty` in `scala.collection.immutable.List` from Scala

jdalton:
Thanks for the PR, @aguspina!

This can be tackled with `_.negate(_.isEmpty)`

aguspina:
@jdalton Yeah! I'm actually using this in my projects, but it's quite annoying, don't you think?

jdalton:
I haven't had a problem with doing `if (!_.isEmpty(...)) {`.

aguspina:
Ok, thanks for your answer! Regards!

PR #1707, opened 2015-12-16
Some issues with lodash-fp.

d10:
The methods in `keyMap` (curryN, getOr, sampleSize) don't seem to work properly too, probably some bind wrapper thing going wrong?

### DIFF
--- a/lib/fp/mapping.js
+++ b/lib/fp/mapping.js
@@ -73,11 +73,11 @@ module.exports = {
   /** Used to map ary to method names. */
   'aryMethodMap': {
     1: (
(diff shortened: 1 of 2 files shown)


jdalton:
No worries on the `sampleSize`, that's a method we have now so it probably just needs proper mapping.

d10:
Ah, didn't notice `sampleSize` was split out.

jdalton:
The curry methods were fine. Lodash curry methods can accept more than the specified arity. The arity signals when the wrapper should invoke `func`.

d10:
My bad, I've always thought it limits the number of arguments pased into the function. I don't think the docs are very clear on that.

jdalton:
I'll adjust the docs.

PR #1717, opened 2015-12-22
Improved example for _.times()

### LINKED ISSUE #1456: Incorrect example in _.times doc.
In the Example section for _.times documentation (https://lodash.com/docs#times), the comment does not match the code (`cast` vs `castSpell`):

```javascript
var diceRolls = _.times(3, _.partial(_.random, 1, 6, false));
// → [3, 6, 4]

_.times(3, function(n) {
  mage.castSpell(n);
});
// → invokes `mage.castSpell(n)` three times with `n` of `0`, `1`, and `2`

_.times(3, function(n) {
  this.cast(n); // <------- This should be this.castSpell(n)
}, mage);
// → also invokes `mage.castSpell(n)` three times

```

jmiziolek:
Comment now matches function name. It's easier to understand.

Fixes #1456 (which was closed but the not really fixed)

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -7794,7 +7794,7 @@ _.times(3, function(n) {
 // => invokes `mage.castSpell(n)` three times with `n` of `0`, `1`, and `2`
 
 _.times(3, function(n) {
-  this.cast(n);
+  this.castSpell(n);
 }, mage);
 // => also invokes `mage.castSpell(n)` three times
 ```


jdalton:
Hi @jmiziolek!

Thanks but this was fixed.
You're looking in the wrong place. Our documentation is generated from the lodash.js source.

jmiziolek:
Hi @jdalton :smile: 
Thanks for explaining!

PR #1735, opened 2016-01-05
Update _.slice() docs

jareware:
Add a small note regarding negative start positions working as they do in Array.slice().

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -6217,14 +6217,15 @@
     /**
      * Creates a slice of `array` from `start` up to, but not including, `end`.
      *
-     * **Note:** This method is used instead of [`Array#slice`](https://mdn.io/Array/slice)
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @jareware!

I think this can be handled by modifying the doc note:
```
     * **Note:** This method is used instead of [`Array#slice`](https://mdn.io/Array/slice)
     * to ensure dense arrays are returned.
```

To say that this method is based on  [`Array#slice`](https://mdn.io/Array/slice) but used instead...

jdalton:
@jareware would you mind updating the PR?

jareware:
Is that better? Or should I drop the negative-index-mention altogether?

jdalton:
You can drop the negative index as it's covered by the `Array#indexOf` link.

jdalton:
Thanks!

Issue #1530, opened 2015-10-14
_.sortBy([...], 'string', this); does not work in 3.10.1


ThisIsManta:
```
_.sortBy(['a', 'ccc', 'bb'], 'length');
=> ["a", "bb", "ccc"]
```

Adding `this` as the third argument made `sortBy` function did nothing at all.
```
_.sortBy(['a', 'ccc', 'bb'], 'length', this);
=> ["a", "ccc", "bb"]
```

However, I was using version 2.4.1 and both mentioned statements worked perfectly.

Note that the later statement appears in *Backbone.js* inside `Collection.sort()` so that's the reason why I can't simply remove the third argument myself.

jridgewell:
That's caused by lodash's shorthand `_.matcher` syntax: `_.filter(col, 'prop', val)`. I think this has been resolved with https://github.com/jashkenas/backbone/pull/3732, which landed in Backbone v1.2.2.

ThisIsManta:
Ah! It seems like the time of upgrading *Backbone.js*. Thanks!

jdalton:
This has been resolved in v4 (edge master). Our prop shorthand is now an array of key/value pair and all methods have dropped the `thisArg` param.

jridgewell:
> Our prop shorthand for matchesProperty is now an array (key/value pair)

What does that look like?

jdalton:
```js
_.filter(objects, [key, prop]);
```

jridgewell:
Won't that conflict with deep properties?

jdalton:
Deep property shorthand is:
```js
_.filter(objects, { key1: value1, key2: value2 });
```

jridgewell:
But that's `_.matcher` shorthand...

jdalton:
Right (in lodash `_.matcher` is `_.matches`) so there's the `_.matches` shorthand and `_.matchesProperty` shorthand.

jridgewell:
But what's the deep property shorthand? `_.filter([{a: { b: true }}], ['a', 'b'])` now does `matchesProperty` (`obj.a === 'b'`).

jdalton:
There's `matches` and `matchesProperty` shorthands:
```js
var objects = [
  { 'a': { 'b': true },  'c': 'foo', 'd': 1 },
  { 'a': { 'b': true },  'c': 'bar', 'd': 0 },
  { 'a': { 'b': false }, 'c': 'foo', 'd': 1 }
];

// matchesProperty
[...]

PR #1739, opened 2016-01-10
add test case for `sortBy` with property syntax and context

krawaller:
I was recently bit by the regression mentioned in #1530 where using `_.sortBy` with a property shorthand and a supplied context stopped working. The bug has since been fixed, but this PR adds a test case to help discover future regressions.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -17995,7 +17995,7 @@
     QUnit.test('should work with "_.property" style `iteratee` and context', function(assert) {
       assert.expect(1);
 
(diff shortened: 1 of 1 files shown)


jdalton:
Thanks for making this @krawaller!

However, the test case is no longer valid because `_.sortBy` no longer accepts a `thisArg` parameter.

krawaller:
Hups, apologies, completely missed that!

Makes sense I guess since it's easy enough to bind the context directly.

jdalton:
> Makes sense I guess since it's easy enough to bind the context directly.

Yep, and it simplifies method implementations as well :grinning:

Issue #1704, opened 2015-12-13
ES6 and FP in Lodash 4


billyjanitsch:
I keep seeing this shim referenced in discussions about bringing FP to Lodash 4:

```js
require('lodash/fp')()
```

I assume that in ES6, this would have to look something like:

```js
import initFp from 'lodash/fp'
import {some, zip} from 'lodash'

initFp()
zip(...)
```

Perhaps I'm missing something, but this doesn't seem very idiomatic to me.
- Is that idea that this should be called once at the top of my project's main entry point(s)? If so, do I also need to include it at the top of all of my tests, assuming they don't import from the main entry?
- Either way, doesn't this add a fair bit of overhead on app init?
- What is the compatibility here with the new Lodash ES build? I'd hate to lose tree-shaking just because I want FP.

What I seem to want is an interface more like the one that exists for lodash-fp:

```js
import {some, zip} from 'lodash/fp'

zip(...)
```

What are your thoughts on this?

Thanks for your work on Lodash!

jdalton:
The current approach is:
```js
var _ = require('lodash/fp')();
```
so in ES6 that would be
```js
import fp from 'lodash/fp';
const _ = fp();
```
for cherry-picked methods you can do
```js
import {some as _some } from 'lodash/fp'
const some = fp('some', _some);
```

billyjanitsch:
Hm. Could you share your reasoning in preferring something like:

```js
import fp from 'lodash/fp'
import {map as _map, filter as _filter, reduce as _reduce, take as _take} from 'lodash'

const map = fp('map', _map)
const filter = fp('filter', _filter)
const reduce = fp('reduce', _reduce)
const take = fp('take', _take)
```

over:

```js
import {map, filter, reduce, take} from 'lodash/fp'
```

I realize that it takes more work to precompile the latter and you're looking to consolidate the Lodash builds, but the former seems incredibly verbose. It also incurs runtime overhead for each import in each file (if I want `map` in another file, I have to call `fp` again).

I'm aware of the advantages of Lodash over Ramda and I'd like to switch, but the former is going to be a tough sell when I can already just do:

```js
import {map, filter, reduce, take} from 'ramda'
```

jdalton:
The `fp` module is the convertor. I could change things up to allow a module to export a lodash that's already converted so a:
```js
import {map, filter, reduce, take} from 'lodash/smth'
```
[...]

PR #1779, opened 2016-01-13
Make fp modules cherry-pickable

epeli:
As discussed in https://github.com/lodash/lodash/issues/1704#issuecomment-170259732
Only a script at this point. Not hooked up in any builds yet.

Easiest way to test this now is just to make it spit out the module wrappers to an existing lodash installation:

```
$ node lib/fp/precompile.js node_modules/lodash/fp
$ node
> var includes = require("lodash/fp/includes");
> includes(1)([2,3,4,1])
true
```

The `includes.placeholder` seems to be undefined. Bug in the `convert` function?


jdalton:
> The `includes.placeholder` seems to be undefined. Bug in the `convert` function?

Not a bug. The `placeholder` is wired up depending on if it's included in the `lodash.js` build where it would be assigned `lodash`. If not the it's up to the user to specify what they want as a placeholder.

jdalton:
@epeli Awesome! Have you signed [our CLA](http://dojofoundation.org/about/cla)?

epeli:
Now I have.

jdalton:
Rock!

lib/fp/precompile.js
```diff
@@ -0,0 +1,100 @@
...
+`.trim();
+
+var collectionTemplate = name => `
+var convert = require("./convert");
+module.exports = convert(require("../${name}"));
+`.trim();
+
+var passThroughTemplate = name => `module.exports = require("../${name}");`;
+
+var moduleTemplate = name => {
```
jdalton: No arrow functions plz. We support Node 0.10.0

lib/fp/precompile.js
```diff
@@ -0,0 +1,100 @@
...
+        path.join(target, aliasName + '.js'),
+        moduleTemplate(origName)
+      );
+    }, aliases);
+  }, mapping.aliasMap);
+}
+
+module.exports = precompileFpWrappers;
+
+if (require.main === module) {
```
jdalton: No need for the cli branch at the moment.

epeli:
Added support for node 0.10 and removed the cli branch. Rebased and force pushed this PR.

jdalton:
Awesome!

PR #1803, opened 2016-01-15
Update contributing.md about how to launch tests

jfmengels:
In e255818, the way to launch tests was updated to include the fp tests, but the contributing.md file was not updated with the new way to launch tests.


jdalton:
Nice thank you!

PR #1807, opened 2016-01-15
Update README.md

binhqx:
Missing a '`' in the chainable wrapper methods section that breaks the formatting.


jdalton:
Thanks @binhqx!

The docs are generate [from the source](https://github.com/lodash/lodash/blob/master/lodash.js) and I believe this has been patched there.

PR #1815, opened 2016-01-16
Fix inconsistency in _.map documentation

iemejia:
I noticed that the applied function didn't correspond to the 'expected'
result. I fixed and added an additional value.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -7722,11 +7722,11 @@
      *   return n * n;
      * }
      *
-     * _.map([1, 2], square);
-     * // => [3, 6]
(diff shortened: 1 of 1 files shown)


iemejia:
I don't know how to do it, but probably a way to avoid such issues is to find a way to take these code snippets from the real js tests. I noticed that one of the tests had a really similar code but with the function 'tripled' instead of square, but I didn't want to change the function.

d10:
Thanks for the fix! Related: https://github.com/dtao/autodoc.

PR #1817, opened 2016-01-17
fix doc: "return" to "returns"

think2011:


jdalton:
Thanks!

think2011:
@jdalton  Could you tell me  how to convert lodash.js to doc HTML? 

I'm using this https://github.com/jldec/lodash-doc-src

jdalton:
We use [php-markdown](https://michelf.ca/projects/php-markdown/) and [docdown](https://github.com/jdalton/docdown). Neither are super great.

think2011:
@jdalton Thank you very much！

PR #1821, opened 2016-01-17
Update old filepath.

d10:


jdalton:
Thanks @d10, but this is all getting changed.

PR #1830, opened 2016-01-18
Updated doc for partition

nacmartin:
If I am not mistaken, the docs for `partition` are wrong. It is only invoked with an argument, not three. See this fiddle https://jsfiddle.net/y8fcwa6m/

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -2627,7 +2627,7 @@ _.orderBy(users, ['user', 'age'], ['asc', 'desc']);
 Creates an array of elements split into two groups, the first of which
 contains elements `predicate` returns truthy for, while the second of which
 contains elements `predicate` returns falsey for. The predicate is invoked
-with three arguments: (value, index|key, collection).
+with one argument: (value).
 
 #### Arguments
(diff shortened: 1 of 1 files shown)


jdalton:
Thanks @nacmartin!

hughlomas:
@jdalton Was the bug really that the documentation needed to be changed, or that the function should actually support all three of `value, key|index, collection` ?

I'm trying to use partition with the index (`index % 2 == 0`) of the array, and searching issues led me to this curious decision. 

I may be missing something obvious, but why does partition's predicate not support those?

jdalton:
@hughlomas Our `partition` method is a form of the methods like `groupBy`, `countBy`, & `keyBy` which transform or group values by their transformed forms which is why they're limited to one param.

hughlomas:
@jdalton It would be useful to have the index/key available for those methods, for example the use case above, or partitioning/grouping/counting an object by key.

Is it just due to some arbitrary categorization of the methods, or is it necessitated by support for some higher level abstraction?

jdalton:
This was to unify, simplify, and create other "xyzBy" method variant  APIs.
It's an intentional design choice in v4.

Issue #1823, opened 2016-01-17
possibility to add update method


hzamani:
What about having an `update` method as in [clojure](https://clojuredocs.org/clojure.core/update)? Specially useful in functional programming. Something like this:  
```js
function update(object, path, func, ...args) {
  set(object, path, func(get(object, path), ...args);
  return object;
}
```

jdalton:
I think that's a pretty neat idea. Let's wait a bit to see if others dig it.

In the meantime can you post a usage example?

hzamani:
For example in a todomvc app with lodash/fp one would have:
```js
var state = {todos: {}, uid: 1}

function logicalNot(x) { return !x; }
function inc(n) { return n+1; }

function addTodo(text) {
  var id = state.uid;
  _.update(state, 'uid', inc);
  _.update(state, 'todos', _.set(id, {text: text, id: id, completed: false}));
}

function toggleTodo(id) {
  _.update(state, ['todos', id, 'completed'], logicalNot);
}

function toggleAllTodos(checked) {
  _.update(state, 'todos', _.mapValues(_.set(checked, 'completed')));
}

function destroyTodo(id) {
  // Also currently unset is not added to lodash/fp :-\
  _.update(state, 'todos', _.unset(id));
}

function clearCompleted() {
   // omitBy wrongly places in aryMethodMap 3 :-|
  _.update(state, 'todos', _.omitBy(_.get('completed')));
}
```

hzamani:
Is it necessary to open separate issues for mentioned bugs on fp?

StreetStrider:
Ramda has [`over`](http://ramdajs.com/docs/#over) which works in complex with various [Lenses](http://ramdajs.com/docs/#lens).

jdalton:
> Is it necessary to open separate issues for mentioned bugs on fp?

Yep. Though check to see if [it's been corrected](https://github.com/lodash/lodash/blob/master/fp/_mapping.js) in the edge master branch.

hzamani:
In the [edge master branch](https://github.com/lodash/lodash/blob/master/fp/_mapping.js) unset is in `aryMethod[1]` witch is wrong, it must be in 2

jdalton:
Thanks pached https://github.com/lodash/lodash/commit/5f78d3ae26ff27407b21ac0cd0498981cc1e50a3.

jfmengels:
[...]

PR #1832, opened 2016-01-19
implement update

hzamani:
Implement update method as proposed in #1823.

### DIFF
--- a/fp/_mapping.js
+++ b/fp/_mapping.js
@@ -70,7 +70,7 @@ exports.aryMethod = {
     'getOr', 'inRange', 'intersectionBy', 'intersectionWith', 'isEqualWith',
     'isMatchWith', 'mergeWith', 'orderBy', 'pullAllBy', 'pullAllWith', 'reduce',
     'reduceRight', 'replace', 'set', 'slice', 'sortedIndexBy', 'sortedLastIndexBy',
-    'transform', 'unionBy', 'unionWith', 'xorBy', 'xorWith', 'zipWith'
+    'transform', 'unionBy', 'unionWith', 'update', 'xorBy', 'xorWith', 'zipWith'
   ],
   '4': [
     'fill', 'setWith'
@@ -121,7 +121,8 @@ exports.iterateeAry = {
   'takeRightWhile': 1,
   'takeWhile': 1,
   'times': 1,
-  'transform': 2
+  'transform': 2,
+  'update': 1
 };
 
 /** Used to map method names to iteratee rearg configs. */
@@ -141,6 +142,7 @@ exports.methodRearg = {
   'setWith': [3, 1, 2, 0],
   'sortedIndexBy': [2, 1, 0],
   'sortedLastIndexBy': [2, 1, 0],
+  'update': [2, 1, 0],
   'zipWith': [1, 2, 0]
 };
 
@@ -175,7 +177,8 @@ exports.mutate = {
   'set': {
     'set': true,
     'setWith': true,
-    'unset': true
+    'unset': true,
+    'update': true
   }
 };
 
--- a/lodash.js
+++ b/lodash.js
@@ -3450,6 +3450,19 @@
     }
 
     /**
+     * The base implementation of `_.update`.
+     *
+     * @private
+     * @param {Object} object The object to query.
+     * @param {Array|string} path The path of the property to update.
+     * @param {Function} func The updater function.
+     * @returns {Object} Returns `object`.
+     */
+    function baseUpdate(object, path, func) {
+      return baseSet(object, path, func(baseGet(object, path)));
+    }
+
+    /**
      * The base implementation of `setData` without support for hot loop detection.
      *
      * @private
@@ -11894,6 +11907,37 @@
     }
 
     /**
+     * Updates the value at `path` of `object` with given `func`. If a portion of `path` doesn't exist
+     * it's created. Arrays are created for missing index properties while objects
+     * are created for all other missing properties. The `func` is invoked with
+     * `value` at `path` of `object`: (value).
+     *
+     * @static
+     * @memberOf _
+     * @category Object
+     * @param {Object} object The object to modify.
+     * @param {Array|string} path The path of the property to set.
+     * @param {Function} func The updater function.
+     * @returns {Object} Returns the updated `object`.
+     * @example
+     *
+     * var object = { 'a': [{ 'b': { 'c': 3, d: '11' } }] };
+     *
+     * _.update(object, 'a[0].b.c', function (n) { return n + 10; });
+     * console.log(object.a[0].b.c);
+     * // => 13
+     *
+     * _.update(object, 'x[0].y.z', function (n) { return n ? n + 1 : 0; });
+     * console.log(object.x[0].y.z);
+     * // => 0
+     */
+    function update(object, path, func) {
+      return object == null
+        ? object
+        : baseUpdate(object, path, baseCastFunction(func));
+    }
+
+    /**
      * This method is like `_.set` except that it accepts `customizer` which is
      * invoked to produce the objects of `path`.  If `customizer` returns `undefined`
      * path creation is handled by the method instead. The `customizer` is invoked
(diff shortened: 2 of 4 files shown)


jdalton:
Thanks for this @hzamani, but we'll hold off until more interest is shown.

jdalton:
@hzamani Have you signed [our CLA](http://dojofoundation.org/about/cla)?

hzamani:
I have just signed it.

lodash.js
```diff
@@ -11405,6 +11424,37 @@
     }
 
     /**
+     * Updates the value at `path` of `object` with given `func`. If a portion of `path` doesn't exist
+     * it's created. Arrays are created for missing index properties while objects
+     * are created for all other missing properties.
```
jdalton: Can you add a sentence on that params the`func` expects. You can reference other methods like `_.every` for an example. Also can you rename the param `updater`.

lodash.js
```diff
@@ -3290,6 +3290,25 @@
       return object;
     }
 
+
+    /**
+     * The base implementation of `_.update`.
+     *
+     * @private
+     * @param {Object} object The object to query.
+     * @param {Array|string} path The path of the property to update.
```
jdalton: The `args` param is missing a doc entry.

lodash.js
```diff
@@ -11405,6 +11424,37 @@
...
+     * console.log(object.x[0].y.z);
+     * // => 0
+     */
+    function update() {
+      var args = Array.prototype.slice.call(arguments),
+          funcargs = args.splice(3);
+      args.push(funcargs);
+      return baseUpdate.apply(null, args);
+    }
+
```
jdalton: Lets simplify this a bit and remove the partial args stuff as that can be done with `_.partialRight` and friends.

hzamani: Good idea

jdalton:
Can you rebase this and squash the commits.

hzamani:
With removed partial args?

jdalton:
The PR should be one commit. Git allows you to rebase (to fix the merge conflict) and squash (it's a git thing) the commits to a single commit.

hzamani:
Rebase would make push non-fast-forward which will cause a pull and merge, and after that the PR would be two commits (one for changes, and one for merge). Am I wrong?

jfmengels:
You should squash the commits to only have one, then rebase it to the `master` branch, so that the conflicts are handled (right now, it's unmergeable). When this gets merged, there will indeed be two commits (one for changes, and one for merge), but @jdalton will rebase it later so that there is only 1 commit (but you'll keep the credit in the git history).

hzamani:
Thanks

jdalton:
You're close, just need to rebase a bit more. MOAR BASE!

jdalton:
@hzamani  Execute these commands
```bash
git checkout update
git pull -f https://github.com/lodash/lodash
git reset --hard 608d11f33582641
git cherry-pick 99bc993938661d2
git push origin +update
```

jdalton:
Sweet!

lodash.js
```diff
@@ -3450,6 +3450,27 @@
...
+     * The base implementation of `_.update`.
+     *
+     * @private
+     * @param {Object} object The object to query.
+     * @param {Array|string} path The path of the property to update.
+     * @param {Function} func The updater function.
+     * @param {Array} args The provided extra arguments for updater `func`.
+     * @returns {Object} Returns `object`.
+     */
+    function baseUpdate(object, path, func, args) {
```
jdalton: `arg` param can be removed.

lodash.js
```diff
@@ -3450,6 +3450,27 @@
...
+      if (object == null) {
+        return null;
+      }
```
jdalton: You can move the `null` check to the `update` function out of `baseUpdate`.

lodash.js
```diff
@@ -3450,6 +3450,27 @@
...
+      var value = baseGet(object, path),
+          updated = isFunction(func)
+            ? func.apply(object, [value].concat(args))
```
jdalton: The `update` function should use `baseCastFunction` so that `baseUpdate` will always get a function and the `isFunction` check can be removed. Since there's no args the `func` invocation can be simplified.

hzamani:
What about update and updatePath in fp?

fp/_mapping.js
```diff
@@ -213,7 +217,8 @@ exports.remap = {
   'getOr': 'get',
   'trimChars': 'trim',
   'trimCharsEnd': 'trimEnd',
-  'trimCharsStart': 'trimStart'
+  'trimCharsStart': 'trimStart',
+  'updatePath': 'update'
```
jdalton: `updatePath` should be in the `aliasToReal` map not `remap`.

jdalton: actually scratch that. Lets not add an alias unless its warranted for cross-lib migration.

hzamani: What about update rearg? Should it be [2, 1, 0] or the default [2, 0, 1]?

fp/_mapping.js
```diff
@@ -70,7 +70,8 @@ exports.aryMethod = {
     'getOr', 'inRange', 'intersectionBy', 'intersectionWith', 'isEqualWith',
     'isMatchWith', 'mergeWith', 'orderBy', 'pullAllBy', 'pullAllWith', 'reduce',
     'reduceRight', 'replace', 'set', 'slice', 'sortedIndexBy', 'sortedLastIndexBy',
-    'transform', 'unionBy', 'unionWith', 'xorBy', 'xorWith', 'zipWith'
+    'transform', 'unionBy', 'unionWith', 'update', 'updatePath', 'xorBy', 'xorWith',
```
jdalton: `'updatePath'` should be removed from this list ^

fp/_mapping.js
jdalton: 2, 1, 0.
`(func, path, object)` which is `(c, b, a)`

lodash.js
```diff
@@ -11894,6 +11907,38 @@
...
+     * console.log(object.a[0].b.c);
+     * // => 13
+     *
+     * _.update(object, 'x[0].y.z', function (n) { return n ? n + 1 : 0; });
+     * console.log(object.x[0].y.z);
+     * // => 0
+     */
+    function update(object, path, func) {
+      if (object == null) {
+        return null;
```
jdalton: It should just return `object` not `null`.

jdalton:
Btw thank you for being amenable to these changes. It helps out.

hzamani:
My pleasure.

jfmengels:
Thanks a lot @hzamani @jdalton!
Not sure why the fp mapping is different than for `_.set` though, do you mind explaining the reasoning behind it?

jdalton:
It mad sense to me for the `func` to come first.

jdalton:
Is there a counter example in an existing FP lib?

jdalton:
```js
fp.update(function(x) { return { 'c2': 2 }; })('a.b')(object);
fp.setWith(Object)('a.b.c')(2)(value);
```

jfmengels:
[Immutable.js](https://facebook.github.io/immutable-js/#nested-structures) has (path, func)
```js
nested2.updateIn(['a', 'b', 'd'], value => value + 1)
```
DId not get your example with `setWith` sorry

jdalton:
"get" as in understand?

jfmengels:
Yes, as in understand.

jdalton:
The `setWith` method allows you to specify how namespaces are created.
Its customizer function comes first in fp.

jfmengels:
Oh, ok, got it.
How I see it:
the `value` argument in `set` is like the `func`argument in `set` (you could replicate `set` with `update` by passing `_.constant(someValue)` as `func`), so it'd make sense having them at the same position. So that, for example, if you need to modify your code to use `update` rather than `set`, you do not have to change the order of the arguments.
I also think it is more readable to have as first argument *what* you want to change, rather than *how* to change it.

IMO the `customizer` function from `setWith` is not that relevant to this, as it has another purpose than describing the replacement value.

jdalton:
@jfmengels Mind opening a PR to change it?

jfmengels:
Not at all #2061

PR #1839, opened 2016-01-19
fix doc: _.mergeWith

think2011:


jdalton:
Thanks, this is fixed in the source already.

think2011:
Ok

PR #1866, opened 2016-01-22
options.timeout argument was added to _.memoize function

silviom:
Hello, I would like to have an option to expire memoized values.


jdalton:
We expose the memoize cache for you to modify as you see fit.

silviom:
Thanks you!

PR #1873, opened 2016-01-25
Remove explicit .chain()

carlbennettnz:
For consistency


jdalton:
The `.chain()` call is required to avoid implicit `value()` of string methods ending the chain.

PR #1885, opened 2016-01-26
chore(scripts): Run scripts in parallel

kentcdodds:
Any scripts that don't have a dependency on something else running first could utilize `&` rather than `&&` and therefore run in parallel :+1:

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,11 +28,11 @@
     "webpack": "^1.12.12"
   },
   "scripts": {
-    "build": "npm run build:main && npm run build:fp",
+    "build": "npm run build:main & npm run build:fp",
     "build:fp": "node lib/fp/build-dist.js",
     "build:fp-modules": "node lib/fp/build-modules.js",
     "build:main": "node lib/main/build-dist.js",
-    "style": "npm run style:main && npm run style:fp && npm run style:perf && npm run style:test",
+    "style": "npm run style:main & npm run style:fp & npm run style:perf & npm run style:test",
     "style:fp": "jscs fp/*.js lib/**/*.js",
     "style:main": "jscs lodash.js",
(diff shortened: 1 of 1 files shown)


jdalton:
It's not needed for this change but have you signed [our cla](http://dojofoundation.org/about/claForm)?

kentcdodds:
I don't know if I had. Done 👍

package.json
```diff
@@ -28,11 +28,11 @@
     "webpack": "^1.12.12"
   },
   "scripts": {
-    "build": "npm run build:main && npm run build:fp",
+    "build": "npm run build:main & npm run build:fp",
     "build:fp": "node lib/fp/build-dist.js",
     "build:fp-modules": "node lib/fp/build-modules.js",
     "build:main": "node lib/main/build-dist.js",
-    "style": "npm run style:main && npm run style:fp && npm run style:perf && npm run style:test",
+    "style": "npm run style:main & npm run style:fp & npm run style:perf & npm run style:test",
```
jdalton: jscs gives a report or errors. Will running in parallel muck that up?

kentcdodds: Ah, yes, it might. Not sure whether jscs streams those or if it just dumps them when its done.

kentcdodds: Never used jscs. But I'm guessing this will raise a flag about errors, then when you're figuring them out you can run each one individually...

jdalton: >  or if it just dumps them when its done

It dumps them. I think it's ok this just won't ensure order which is fine.

kentcdodds: Yeah, I figured this was more about getting the work done (quicker the better) than getting the order correct. Thanks for merging my first PR to lodash :+1:

PR #1887, opened 2016-01-26
Add npm scripts to build docs

jfmengels:
Adding npm scripts to build docs, based on [this gist](https://gist.github.com/jdalton/938ea49ab3919f1c046c)

Leveraging the fact that scripts ran using `npm run` have the package's version set in the environment variables (as `npm_package_version`). This way, the version won't have to be manually updated every time the version changes.

I felt obligated to create a `build-doc` bash file containing the whole command, as it made the package.json huge horizontally (and it's not possible to create multiline commands). I put the bash file in doc/ as it was to me the most sensible, but I don't know it is a problematic solution.

### DIFF
--- a/doc/build-doc
+++ b/doc/build-doc
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+docdown \
+  ./lodash.js  \
+  ./doc/README.md  \
+  toc="categories"  \
+  url="https://github.com/lodash/lodash/blob/$npm_package_version/lodash.js"  \
+  title="<a href=\"https://lodash.com/\">lodash</a> <span>v$npm_package_version</span>"  \
+  hash="$1"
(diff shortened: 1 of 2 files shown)


jdalton:
Wow, awesome! I still manually correct `_.prototype.Symbol.iterator` to `_.prototype[Symbol.iterator]` because of a bug in `docdown` but this helps a bunch!

jdalton:
BTW have you signed our [cla](http://dojofoundation.org/about/cla)?

jfmengels:
For the CLA, I just have (sorry about that).

It's probably worth making a script out of the other commands in that gist that are used on a regular basis. It's pretty non-newcomer, as you pretty much have to search that certain issue where you give the link to the gist (and I can't remember how I did it off the top of my head).

And you're welcome :)

jfmengels:
Btw, not sure what you meant by the bug in docdown or how it relates to this. Do you mean that you don't use the npm in the node_modules but a patched local version?

jdalton:
I mean docdown doesn't know how to handle properties of "Symbol.iterator". It doesn't know to enclose it in brackets.

PR #1888, opened 2016-01-27
Make `isEmpty` method faster on objects

nkt:
Here is the benchmark: http://jsperf.com/fastest-check-if-object-is-empty/2.


jdalton:
I can see optimizing it a bit, maybe not like this but I'll kick it around a bit.

nkt:
The main idea is that current algorithm has `O(n)` complexity, when it could be `O(1)`.

jdalton:
Have you signed [our cla](http://dojofoundation.org/about/cla)?

nkt:
Yes, as `nkt`, just now.

PR #1897, opened 2016-01-29
Add scripts to generate the docs for lodash/fp

jfmengels:
This PR adds scripts to generate docs for lodash/fp rather than lodash. This is made in the hope of having them available on the [documentation page](https://lodash.com/docs) of the website.

The parameters are reordered:
`_.set(object, path, value)` --> `_.set(path, value, object)`
And also the code samples
```js
var object = { 'a': 1, 'b': '2', 'c': 3 };
_.pickBy(object, _.isNumber);
-->
_.pickBy(_.isNumber, object);
// => { 'a': 1, 'c': 3 }
```
even though some examples mutate values, making the code sample not right
```js
_.set('x[0].y.z', 5, object);
console.log(object.x[0].y.z); // --> object is not mutated, so the example is not 100% correct
// => 5
```
Also, some examples (for `transform` and `reduce`) did not specify their optional third argument, which is not optional anymore in `lodash/fp`. I added the third argument so as not to cause problems with this, I hope that's okay.

I tried to do this without forking `docdown`, and it works, except for one tiny little thing: `docdown` would need change [one line](https://github.com/jdalton/docdown/blob/master/lib/entry.js#L51) from
`this.getExample = _.memoize(getExample)` to `this.getExample = _.memoize(this.getExample)`, which would allow overriding the "default" behavior.

Currently, the generated document is written at the same location as the regular doc and would need to be changed if the website is updated to present both. And I haven't gone through all the code samples and functions to verify that everything was reordered correctly.

Tell me what you think!

(submitted a CLA on 2016-01-26 22:19:22)


jdalton:
Let's do it!

jfmengels:
Haha! Didn't expect this merge to be this quick (or to happen). Do note that it will not work without an update to docdown (which I feel that you're already working on). Thanks anyway! I hope it'll become something useful!

jdalton:
Thank you! Thank you! Thank you!

It may not be in place for the v4.1.0 but it's a great start to it.

jfmengels:
You're welcome! Tell me if there's anything else I can do to help with the docs ;)

demoneaux:
:+1:

PR #1907, opened 2016-01-30
_.attempt() should preserve custom errors

VanTanev:
Previously, custom errors were wrapped in a `new Error(e)`.

CLA signed as VanTanev (submitted a CLA on 2016-01-30 00:30:18).

Thanks!


jdalton:
Thank you Ivan!

Issue #1911, opened 2016-01-31
lodash/fp alias set to assoc?


billyjanitsch:
Lodash/fp aliases `omit` to `dissoc`, so it's kind of surprising that it doesn't alias `set` to `assoc`.

jdalton:
Moved to #1912

jfmengels:
Would it make sense now (because I feel like `assoc` and `dissoc` should be polar functions with opposite behaviors) to add deeply nested omitting with `omit`?
SInce you can do
```js
var object = { a: { c: 20 } };
_.set(object, 'a.b', 100)
// { a: { b: 100, c: 20 } }
```
but not
```js
_.omit(object, ['a.b'])
// { a: { c: 20 } }
```

PR #1912, opened 2016-01-31
Alias set to assoc in FP

billyjanitsch:
Re: #1911


jdalton:
Thank you!

jdalton:
You could also add `assocPath` as `set` accepts a path too.

billyjanitsch:
Hm, it'd be weird to have `assocPath` but not `dissocPath`, no? And `dissocPath` wouldn't be as straightforward as aliasing `omit`.

PR #1913, opened 2016-01-31
Alias omit to unset in FP

billyjanitsch:
I realize that `omit` is already aliased to `dissoc`, but currently lodash/fp's `unset` is not immutable, which is surprising and arguably incorrect for an FP method.

You could make `unset` immutable, but then it would be useless to return a boolean, so it might as well just be overwritten by `omit` to provide the same functionality as `dissoc`. Then it would pair more sensibly with `set`, which currently *is* immutable.

### DIFF
--- a/fp/_mapping.js
+++ b/fp/_mapping.js
@@ -27,6 +27,7 @@ exports.aliasToReal = {
   'somePass': 'overSome',
   'unapply': 'rest',
   'unnest': 'flatten',
+  'unset': 'omit',
   'useWith': 'overArgs',
   'whereEq': 'filter',
   'zipObj': 'zipObject'


jfmengels:
From what I can see, `unset` is immutable, but it simply returns a boolean which is kind of useless behavior in FP. But unsetting a deep value seems to not be possible in FP with the current implementation of `unset`, which is too bad and should probably be fixed  (or be made available under a new name). Aliasing it to `omit` would bury the functionality though.

Note: Considering that `set`is now aliased to `dissoc`, `unset` would actually be a better alias than `omit` (since `unset` supports deep unsetting, just like `set` supports deep setting).

jdalton:
Looks like `unset` need to be fixed.

jdalton:
Also https://github.com/lodash/lodash/commit/cc20163cf76e8b02f5e745b11c7a987cb16bb057.

jfmengels:
What's the value in having both `assoc` and `assocPath` (and their dissoc equivalents) if they all alias to `set`/`unset`? Is it closer in naming convention to other names/aliases?

jdalton:
It covers Ramda equivalents.

billyjanitsch:
@jdalton Will this make fp's `unset` return the object instead of a boolean?

@jfmengels Also, in standard FP, `assoc` takes a string and `assocPath` takes an array (`set` covers both). Even though they're just aliases, it's nice for devs who dislike overloading and enforce these types by convention.

jdalton:
> Will this make fp's unset return the object instead of a boolean?

Yep.

PR #1918, opened 2016-02-01
Update indexOf documentation

onumossn:
I think the option to pass `true` for indexOf to use binary search was removed when sortedIndexOf was added.


jdalton:
Yep it was! Thanks.

PR #1922, opened 2016-02-01
Reword docs for _.tap and _.thru

jjt:
Recomend `_.thru` for altering/replacing intermediate values and `_.tap`
for triggering side-effects

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -7320,8 +7320,9 @@ var youngest = _
 
 This method invokes `interceptor` and returns `value`. The interceptor is
 invoked with one argument; (value). The purpose of this method is to "tap into"
-a method chain in order to perform operations on intermediate results within
-the chain.
+a method chain in order to perform operations based on intermediate results
+within the chain. To mutate, alter, or replace intermediate results, use
+`_.thru`.
 
 #### Arguments
 1. `value` *(&#42;)*: The value to provide to `interceptor`.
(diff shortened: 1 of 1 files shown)


jdalton:
I like the `array.pop()` as it was. That example hits a use case devs ask about. The issue here is both allow mutating the value but really `tap` is for mutation and `thru` is for replacing.

jdalton:
So the wording should reflect that.

jdalton:
That `tap` is for mutation and `thru` is for replacing.

jjt:
I don't think mutation of this sort is a good model to espouse in general. A chaining pipeline is easier to reason about if each step in the pipeline is a function that takes the current value and returns a new value to the next step.

`tap` for mutation and `thru` for replacement forces a conceptual divide between altering objects and primitive data types. There's no way to "mutate" a string using `_.tap`, you have to use `_.thru`.

It seems conceptually cleaner to me to recommend `_.thru` for any alteration of the intermediate value, and `_.tap` for side effects.

jjt:
Another thing that bothers me about the `array.pop()` example is that there are better ways to do that (calling `.initial` directly on the chain). 

I think my intent behind this PR might be too opinionated toward FP ideas of pure functions and immutability (where practical in JS), which might not be suitable for recommendation in the docs.

jdalton:
> I think my intent behind this PR might be too opinionated toward FP ideas of pure functions 

I'll do a compromise while still maintaining your commit cred.

jjt:
Sure, that sounds reasonable to me. Pushed a new commit.

PR #1924, opened 2016-02-02
fix _.sortedUniqBy example

laispace:
fix output of the example.


jdalton:
Thank you!

jdalton:
For future reference please modify the source and not the generated markdown.

laispace:
@jdalton Got it.

PR #1961, opened 2016-02-08
added support for variable-depth flatten

msmorgan:
This commit adds an optional `maxDepth` argument to `_.flatten`, allowing a nested array to be recursively flattened to a variable degree.

Please double-check I made the correct change in fp/_mapping.js, because it was hard to tell what was going on in that file.


msmorgan:
I forgot to save a file so I'm closing and re-submitting this.

jfmengels:
You can force-push your branch instead of closing this one.

msmorgan:
Done. Didn't know you could do that.

jdalton:
Could you remove the `dist/` files.

jfmengels:
I'm pretty sure you should not commit the dist/ and doc/ folders.
As for the mapping, looks good, but it depends on the behavior that you want. If you put it in the aryMethod[1] category, then you won't be able to use the maxDepth option out of the box. If you put it in the aryMethod[2], then you will be forced to add a maxDepth value every time.

I'm guessing the common case is that people want it to be flattened only by one level, and would use flattenDeep otherwise. I'd argue it would actually make more sense to have the maxDepth option in flattenDeep function (then put that one in the 2-args mapping, and flatten in the 1 or keep it uncapped).

lodash.js
```diff
@@ -2494,14 +2494,19 @@
...
      * @param {boolean} [isStrict] Restrict flattening to arrays-like objects.
      * @param {Array} [result=[]] The initial result value.
      * @returns {Array} Returns the new flattened array.
      */
-    function baseFlatten(array, isDeep, isStrict, result) {
+    function baseFlatten(array, maxDepth, isStrict, result) {
+      maxDepth = maxDepth === undefined ? 1 : nativeMax(toInteger(maxDepth), 0);
       result || (result = []);
 
+      if (maxDepth < 1) {
```
jdalton: The min depth should be `1`. A flatten that doesn't flatten isn't a flatten.

msmorgan: I intentionally made it callable with `maxDepth == 0` for flexibility purposes. It defaults to 1. I'll change it if you think it's necessary.

msmorgan:
If the consensus is that I should move this functionality to `flattenDeep` (with a default `maxDepth` value of `Infinity`?), I can change it tomorrow.

lodash.js
jdalton: Ya it shouldn't go below `1`.

jdalton:
> If the consensus is that I should move this functionality to `flattenDeep` (with a default `maxDepth` value of `Infinity`?), I can change it tomorrow.

I think that a separate method would work. I like `_.flatten` because it's simple and doesn't have any extra params that muck up iteration. I've tried to split functionality into separate methods like `_.flatten` and `_.flattenDeep`. If this is added I imagine it would be under a `_.flattenDepth` or something.

jdalton:
@msmorgan Have you signed [our cla](http://dojofoundation.org/about/cla)?

lodash.js
megawac: didn't look too closely, but is the case needed for recursion?

msmorgan: It's not, since it won't be called recursively once `maxDepth` reaches `1`.

jdalton: Naw just base functionality flatten should flatten at least 1 level. I knew a depth level [was coming for a while](https://github.com/mout/mout/commit/5957b66589f13f73de86b273f6d0040106cbd7a1) :P

msmorgan:
I just signed it.

If this will be a separate method, what do you think of the name `_.flattenBy`?

jdalton:
@msmorgan 
> If this will be a separate method, what do you think of the name `_.flattenBy`?

No the "By" postfix has another meaning. I think `_.flattenDepth` is fine.

msmorgan:
@jdalton 

Okay, `_.flattenDepth` it is.

What should it do if a depth of `0` is passed? I think it should simply copy the array. Wanting to recursively flatten `0` times might not come up that often, but if it does for someone, that behavior seems correct.

However, I concede that `baseFlatten` should probably default to `1` if non-numeric data is passed.

jdalton:
> What should it do if a depth of `0` is passed?

It's gross but consistent to do a simple `arrayCopy`. I checked and other methods that default to `n` of `1` still allow `0` and you get what you get.

StreetStrider:
`flattenTo`?
`flattenUnder`?
`flattenAt` (at level)?

jdalton:
I like `flattenTo` but we already have a pattern of taking a param and making it part of the name

* `xyzDeep`
* `xyzSize`
* `xyzN`

So turning in to `_.flattenDepth` is fine

msmorgan:
I've moved the new functionality into a separate `_.flattenDepth` function.

Please verify that I am defaulting the `depth` argument to `baseFlatten` correctly.

test/test.js
```diff
@@ -27,6 +27,7 @@
 
   /** Used as a reference to the global object. */
   var root = (typeof global == 'object' && global) || this;
+  debugger;
```
jdalton: rogue `debugger`

lodash.js
```diff
@@ -11457,7 +11484,7 @@
      * This method is like `_.merge` except that it accepts `customizer` which
      * is invoked to produce the merged values of the destination and source
      * properties. If `customizer` returns `undefined` merging is handled by the
-     * method instead. The `customizer` is invoked with seven arguments:
+     * method instead. The `customizer` is invoked with six arguments:
```
jdalton: Address this doc nit in a separate doc PR.

lodash.js
```diff
@@ -2494,12 +2494,17 @@
...
      * @param {Array} array The array to flatten.
-     * @param {boolean} [isDeep] Specify a deep flatten.
+     * @param {number} [depth=1] The maximum recursion depth.
      * @param {boolean} [isStrict] Restrict flattening to arrays-like objects.
      * @param {Array} [result=[]] The initial result value.
      * @returns {Array} Returns the new flattened array.
      */
-    function baseFlatten(array, isDeep, isStrict, result) {
+    function baseFlatten(array, depth, isStrict, result) {
+      depth = isNaN(toNumber(depth)) ? 1 : toInteger(depth);
```
jdalton: Move `toInteger` use to `flattenDeep` to keep it out of `baseFlatten`. The `baseFlatten` should assume receiving an integer depth of > 0. In `flattenDeep` no need for the `isNaN` and `toNumber` juggle just use `toInteger`.

test/test.js
msmorgan: Whoops! Should have noticed that.

At one point, I had a bug where `baseFlatten` was defaulting `depth` to `0` instead of `1` (because `toInteger(undefined)` behaved differently than I assumed), and the test suite wouldn't even load.

msmorgan:
New commit fixes those three things.

lodash.js
```diff
@@ -2494,12 +2494,17 @@
...
-     * @param {boolean} [isDeep] Specify a deep flatten.
+     * @param {number} [depth=1] The maximum recursion depth.
      * @param {boolean} [isStrict] Restrict flattening to arrays-like objects.
      * @param {Array} [result=[]] The initial result value.
      * @returns {Array} Returns the new flattened array.
      */
-    function baseFlatten(array, isDeep, isStrict, result) {
+    function baseFlatten(array, depth, isStrict, result) {
+      depth = depth === undefined ? 1 : depth;
+      if (depth < 1) {
```
jdalton: You can move the `depth < 1` and `copyArray()` bit to `flattenDepth`.

lodash.js
```diff
@@ -5832,12 +5837,34 @@
...
      *
-     * _.flattenDeep([1, [2, 3, [4]]]);
-     * // => [1, 2, 3, 4]
+     * _.flattenDeep([1, [2, [3, [[4]]], 5]]);
+     * // => [1, 2, 3, 4, 5]
      */
     function flattenDeep(array) {
       var length = array ? array.length : 0;
-      return length ? baseFlatten(array, true) : [];
+      return length ? baseFlatten(array, MAX_SAFE_INTEGER) : [];
```
jdalton: `INFINITY` use is fine.

jdalton:
Thank you @msmorgan!

PR #1963, opened 2016-02-09
Improve fp doc generation

jfmengels:
This is a follow-up to the work done in #1897, and adds a lot of changes to improve the signature of the methods and the examples in the generated docs.
I have compiled a list of changes in this [gist](https://gist.github.com/jfmengels/6b973b69c491375117dc), and the current result of the generation is available there too.

There is still a few things to do/fix (also listed in the gist, including converting the tests I wrote to QUnit), and ultimately, some examples will have to be adapted (either in this PR or later) because they won't make that much sense. For example: `pullAt`, `set`. Mostly the methods mutating the entry and therefore using console.logs(), and the examples that use too many arguments than are supported in fp.

Anyway, I was told it would be nice to share my progress, so there you go ;)


jdalton:
Can you rebase this and any opportunity to squash some commits is  a good thing.

jfmengels:
Rebased and squashed it. Got it down to four (relatively big) commits.

jfmengels:
Ok, well, I think I've ended at a somewhat pretty satisfying point. I've done all the improvements that I currently could think of.

Current pain points are all the examples that are still broken, mostly because the examples given don't make sense in FP:
- they use arguments that are now ignored ([`trim core`](https://lodash.com/docs#trim) vs [`trim fp`](https://gist.github.com/jfmengels/6b973b69c491375117dc#_trimstring))
- they pass too many arguments now that they are capped ([`assign core`](https://lodash.com/docs#assign) vs [`assign fp`](https://gist.github.com/jfmengels/6b973b69c491375117dc#_assignobject-sources))
- they use `console.log()` because they wanted to show mutated values ([`set core`](https://lodash.com/docs#set) vs [`set fp`](https://gist.github.com/jfmengels/6b973b69c491375117dc#_setpath-value-object), [`pullAt core`](https://lodash.com/docs#pullAt) vs [`pullAt fp`](https://gist.github.com/jfmengels/6b973b69c491375117dc#_pullatindexes-array))

jdalton:
That's cool. I wouldn't worry too much about it. We could always inject a "Note" into the mutation methods that state the example should be taken with a grain of salt.

jfmengels:
Agreed, but I'm sure some examples could be rewritten so that it works in both modes, but this can be an ongoing process.

Do you have any idea on when this could get published on the website? Waiting for lodash/lodash.github.io#8 to advance?

Thanks for merging the PR, hope you like it, and let me know if you want some improvements on this part :)

jdalton:
> Do you have any idea on when this could get published on the website? 

I'll clean it up this go around. I imagine the bump after next will see it. I generate our current docs by pointing its markdown to html converter to a https://rawgit.com of a gist of our documentation. I could do the same for the fp docs (not a ton of work there). Beyond this I'd dig entries in the [build-differences](https://github.com/lodash/lodash/wiki/Build-Differences) or a separate wiki just for fp. I'd like to cover the what, why, how of it and touch on the new fp convert options.

doug-numetric:
[pullAt fp](https://gist.github.com/jfmengels/6b973b69c491375117dc#_pullatindexes-array) says:

> Removes elements from `array` corresponding to `indexes` and returns an array of removed elements.

But the fp version of `pullAt` actually returns a new array without the removed elements...

PR #1969, opened 2016-02-10
Fix code style error and doc nit

jfmengels:
Fix code style error introduced in 7811d59 and doc nit.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -11528,7 +11528,7 @@
      * @category Object
      * @param {Object} object The source object.
      * @param {...(string|string[])} [props] The property names to omit, specified
-     *  individually or in arrays..
+     *  individually or in arrays.
      * @returns {Object} Returns the new object.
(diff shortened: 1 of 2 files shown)


jdalton:
Thank you!

captbaritone:
Couldn't linting as part of Travis help catch this kind of error?

jdalton:
I don't want to fail a test for linting. We should add it to the prepublish script though.

jdalton:
@captbaritone Would you be open to creating a PR to add it add a prepublish npm script to run "npm run style" ?

jfmengels:
Would it be possible to make Travis run the tests in the 7 different jobs as it is doing now, and add a new job that only runs the linting. That way, tests wouldn't break, but you could still see if there are linting errors.

captbaritone:
Sure, I'll take a look.

PR #1977, opened 2016-02-10
Add a note about in-browser usage of lodash-fp

ArturBaybulatov:


jdalton:
Aw yea! Thank you for creating the wiki entry! :tada: 

It's not needed for this specific PR but have you signed [our cla](http://dojofoundation.org/about/cla)?

PR #1986, opened 2016-02-11
Add paragraph about making feature requests in docs

jfmengels:
I'm seeing plenty of:
- feature requests with people asking if it is okay to annoy @jdalton with a proposal
- poorly documented feature requests
- duplicate feature requests

So I thought it'd be a good idea to write in the docs how feature requests work (it will often not be read in time, but it can't hurt much), the most important point being that they should search for similar proposals in the closed issues (and maybe mentally prepare them to the fact that their issue will be closed).

I'm not sure about the wording, but I think it has its place in the docs.

(Also, I capitalized Lodash wherever I thought was fitting, which I can move to another PR)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,11 @@
-# Contributing to lodash
+# Contributing to Lodash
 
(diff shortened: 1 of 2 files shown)


jdalton:
Great!

PR #1993, opened 2016-02-12
Add _.asArray method

### LINKED ISSUE #953: Ensure something is array
I think this is a feature request, but it's possible it already exists and I just can't find it.

I often find myself writing the following:

```js
_.isArray(something) ? something : [something]
```

It would be useful if there were a function called something like ensureArray that did this:

```js
_.ensureArray([1, 2]); // [1, 2]
_.ensureArray(1); // [1]
```

I'd be willing to write this if other people think it would be useful in the core. Also, not a fan of `ensureArray`—if anyone could think of a better name, that would be useful.

megawac:
I did it!

How should `null` and `undefined` be handled?
Empty array or `[null]`, `[undefined]`?
How about `_.asArray()`?

Resolves #953


callumacrae:
Damn I did it too :frowning: 

Your implementation is probably better! :D

megawac:
@callumacrae :racehorse:

callumacrae:
Plus I had no idea what I was doing :sweat_smile:

lodash.js
```diff
@@ -10424,6 +10424,41 @@
...
+     * // => [1, 2, 3]
+     *
+     * _.toArray('abc');
+     * // => ['abc']
+     *
+     * _.toArray(1);
+     * // => [1]
+     *
+     * _.toArray(null);
+     * // => [null]
```
jfmengels: Might be worth it to add as an example when the value is an array 
```js
// something like
_.toArray([1, 2, 3])
// => [1, 2, 3]
```

test/test.js
```diff
@@ -21172,6 +21172,47 @@
...
+      assert.expect(wrapCases.length * 2);
+
+      _.each(wrapCases, function(val) {
+        var result = _.asArray(val);
+        assert.deepEqual(result, [val], 'for value ' + val);
+        assert.strictEqual(result[0], val, 'should not copy the value');
+      });
+    });
+
+    QUnit.test('should an array copy if provided array', function(assert) {
```
jfmengels: 'should an array copy if provided array' --> 'should copy the value if value is an array'
(or should at least be 'should copy an array ...'

jdalton:
Related to

<a href="https://twitter.com/jdalton/status/686610113126256640">

test/test.js
jdalton: > 'should an array copy if provided array' --> 'should copy an array ...'
 
Lets not. In the fp form we can make it clone.

lodash.js
```diff
@@ -10424,6 +10424,41 @@
...
+     */
+    function asArray(value) {
+      if (isArray(value)) {
+        return copyArray(value);
+      }
+      if (iteratorSymbol && value && value[iteratorSymbol] && !isString(value)) {
```
jdalton: ^ That seems more like a `toArray` thing. Lets leave it out for now.

jfmengels:
Maybe add a test for `undefined` --> `[]` (if that's the expected behavior).
Also, maybe add a @since tag (4.4.0 I guess)?

lodash.js
```diff
@@ -10424,6 +10424,41 @@
...
+        return iteratorToArray(value[iteratorSymbol]());
+      }
+      return [value];
+    }
+
+    /**
```
jdalton:
```js
function asArray() {
  if (!arguments.length) {
    return [];
  }
  var value = arguments[0];
  return isArray(value) ? value : [value];
}
```

megawac: Don't all similar methods clone in lodash/_? I feel like not cloning may
confuse the user.

I'll remove the iterator case but it seemed appropriate for this method

jdalton: Cool. Adding shallow cloning is easy enough before we bump. I'll ask for feedback on the tweets.

megawac: So no clone in this PR?

jdalton: No clone for now.

test/test.js
```diff
@@ -21172,6 +21172,47 @@
...
+  (function() {
+    var wrapCases = [null, void 0, 1, '', 'ab', 0, {a: 1}, Object('xyz'), NaN];
+
+    QUnit.test('should wrap non array items in an array', function(assert) {
+      assert.expect(wrapCases.length * 2);
+
+      _.each(wrapCases, function(val) {
+        var result = _.asArray(val);
+        assert.deepEqual(result, [val], 'for value ' + val);
+        assert.strictEqual(result[0], val, 'should not copy the value');
```
jdalton: `strictEqual` won't work with `NaN` but `deepEqual` will. Maybe `NaN` to a separate test.

jridgewell:
:-1: until the name is [changed](https://github.com/lodash/lodash/issues/953#issuecomment-73746507). :stuck_out_tongue_winking_eye:

jdalton:
Ya know, part of an API is a fun API. Folks :heart: `_.kebabCase`, I :heart: `_.rearg`, lets do `_.castArray`. I'll laugh every time and it sets the tone for not cloning the value.

melnikov-s:
I knew this looked familiar! jashkenas/underscore#1467

Anyways +1 to an empty array for `undefined` not sure how `null` should be handled though.

Returning an array copy vs passing back the reference is also tricky. The later is obviously faster but can have surprising behavior in edge cases.

test/test.js
leobalter: test262's `sameValue` handles NaN comparison, it makes me want to extend QUnit to have a similar assertion, including a `sameValueZero` (where it compares `-0` as equal to `+0`).

Would it be useful on LoDash?

leobalter:
As I mentioned on Twitter, I would like to see some consistency with the native `Array.of` (and `_TypedArray_.of`) where it creates an array containing undefined and null values. Examples:

```js
Array.of(); // []
Array.of(undefined); // [undefined]
Array.of(null); // [null]
```

This is also interesting if you the function receives a list of parameters:

```js
Array.of(undefined, undefined); // [undefined, undefined]

// why not?

_.asArray(undefined, 42, undefined); // [undefined, 42, undefined] (better than [])
```

jdalton:
`Array.of` serves a different purpose.

leobalter:
That would be consistent with the spread operator as well:

```js
let fn = (...args) => args;
fn(undefined, 42, undefined);
```

(these examples are not covering arrays intances on the arguments)

leobalter:
> Array.of serves a different purpose.

I know, and that's why I like `_.asArray`, at least how it deal with arrays and other objects. My suggestion is just related to keep it consistent on the other parts.

jdalton:
:shipit:

jdalton:
Renamed to `_.castArray` and exposed `_.castFunction`, and `_.castPath`.

megawac:
+, you beat me to rebasing

On Sat, Feb 13, 2016 at 3:58 PM, John-David Dalton <notifications@github.com

MeoMix:
Thanks for adding this @megawc :D Funny to see you around, but happy to use your methods!

PR #2002, opened 2016-02-16
Use Object.create for baseCreate

megawac:


lodash.js
```diff
@@ -2373,17 +2374,9 @@
...
-        if (isObject(prototype)) {
-          object.prototype = prototype;
-          var result = new object;
-          object.prototype = undefined;
-        }
-        return result || {};
-      };
-    }());
+    function baseCreate(proto) {
+      return objectCreate(!isObject(proto) ? null : proto);
```
jdalton: Not `null` but `{}`.

jdalton: A translation of old to new would be:
```js
function baseCreate(proto) {
  return isObject(proto) ? objectCreate(proto) : {};
}
```

megawac: I thought `null` would be the default, why `{}`? Are we going to allow `baseCreate(null)` when Object.create is supported, or ignore because non shimmable?

jdalton: >  Are we going to allow `baseCreate(null)` when Object.create is supported, or ignore because non shimmable?

ignore it because its non-shimmable.

megawac: done

Issue #1892, opened 2016-01-28
Is this the behaviour of _.remove


SridharGS:
**_.remove**

_Arguments_

_array (Array): The array to modify.
[predicate=_.identity] _(Function|Object|**string**)_: The function invoked per iteration.
[thisArg] (*): The this binding of predicate._

Consider

var temp=[ {"test":"test"},{"test1":"test1"}];
_.remove(temp,{"test":"test"})

this removes object {"test":"test"}

but when

var temp =["test","test1"]

_.remove(temp,"test")

this wont remove anything. Is this a bug please let me know
 
![screen shot 2016-01-28 at 9 29 19 am](https://cloud.githubusercontent.com/assets/8552839/12634671/aa4d979a-c5a1-11e5-89e0-ddf77c37b15e.png)

jdalton:
The `_.remove` method accepts a predicate function or predicate shorthand not the actual value to remove. What you're looking for is `_.pull`.

bsonntag:
I bashed into this, too.
Could maybe `_.remove`'s documentation point to `_.pull`?

jdalton:
> Could maybe `_.remove`'s documentation point to `_.pull`?

Yes! Would you be up for sending a PR? All you would have to do is modify the jsdoc in the [lodash.js](https://github.com/lodash/lodash/blob/master/lodash.js) file. No doc building needed.

bsonntag:
> Yes! Would you be up for sending a PR?

Sure!

By the way, I noticed the CONTRIBUTING.md file is missing in master. Is [this](https://github.com/lodash/lodash/blob/3.10.1/CONTRIBUTING.md) still valid?
Also, this broke "Contributing" links in lodash.com and the repo's readme.

jdalton:
> Links are still broken, tho.

Updated the site link :)

PR #2027, opened 2016-02-19
Links _.remove and _.pull in the docs.

bsonntag:
It seems that `_.remove` is confusing people on how it works (https://github.com/lodash/lodash/issues/1892), so I added a note to its documentation pointing to `_.pull` and a note in `_.pull` pointing to `_.remove`.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -6187,6 +6187,7 @@
      * for equality comparisons.
      *
      * **Note:** Unlike `_.without`, this method mutates `array`.
+     * **Note:** To remove elements from an array based on a predicate, see `_.remove`.
      *
      * @static
      * @memberOf _
@@ -6295,6 +6296,7 @@
      * three arguments: (value, index, array).
      *
      * **Note:** Unlike `_.filter`, this method mutates `array`.
+     * **Note:** To remove elements from an array based on their values, see `_.pull`.
      *
      * @static
      * @memberOf _


lodash.js
```diff
@@ -6208,6 +6208,7 @@
      * This method is like `_.pull` except that it accepts an array of values to remove.
      *
      * **Note:** Unlike `_.difference`, this method mutates `array`.
+     * **Note:** To remove elements from an array based on a predicate, see `_.remove`.
```
megawac: This is `_.pullAll` not `_.pull`

jdalton: Imma call `pull` on that one. 
You hit the `pull`s eye there.
`pull` hockey I don't believe it.

jdalton: The reason for the variety of pulls is that `pull` accepts any value or values to remove. The value could be an array. So to make composition easier there's a `pullAll` which accepts an array of values to pull. This means `pull` is just `rest(pullAll)`. Sometimes you want to pull by a transform so then there's `pullAllBy`. We avoid the `pullBy` form because that would be a variable arg signature depending on how many values you want to pull. Finally `pullAt` is for removing by index and not value.

bsonntag: Woops.
I didn't see it _at_ _all_.

bsonntag: Fixed.

jdalton:
Can you squash these commits.

bsonntag:
Done.

jdalton:
Though not required for this PR have you signed [our cla](http://dojofoundation.org/about/cla)?

bsonntag:
Well, it states donation of "fixes or *documentation*", so I guess it's actually required.
Signed it just now, anyway.

jdalton:
Your kind of submission isn't enough to require it but thanks for schooling me on my own cla.

PR #2060, opened 2016-02-26
Add failing test for mergeWith with multi-ref source object

dannyfallon:
This seems pretty straightfoward to me. I've got a source object
I want to merge which has two keys, only one of which exists in
the object I want to merge into. Both keys in the source object
point to the same object, an array.

What I expect to come out of the merge is an object where the
key that existed before has been modified to contain the values
of the array too, and the key that didn't exist just got added
with the values of the array *only*.

What actually happens is the two keys in the final object are
exactly the same, a strict equals comparison is true. This is
unexpected.

We can take this problem a little further and start with an initial
object with both keys from the source object defined and set to
different things:
```js
var food = {
  vegetables: ['carrot'],
  greens: ['cabbage']
};
```
The result of `mergeWith` will actually end up destroying data because
`actual.greens` will no longer contain the element `cabbage`. This initial
state also produces an ordering problem. If you swap the order of the
keys in the source object, the final result is different - you end up with
`actual.vegetables` equal to `['cabbage', 'brocolli', 'peas']` and have
lost the `carrot` element. Merges with the same keys and data in the
source object feels to me like something that should be deterministic.

The problem appears to stem from `baseMergeDeep`'s stack because
the customiser function is only executed once, for the first key.
It feels like a bad assumption to make is that if you've seen a
source key's value once, the merge result will be the same no
matter what initial key you're merging into.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -14386,6 +14386,24 @@
 
       assert.deepEqual(actual, { 'a': { 'b': ['c'] } });
     });
+
+    QUnit.test('should merge correctly when merging a source object with two keys referencing the same object', function(assert) {
+      assert.expect(1);
+      var vegetables = ['peas', 'brocolli'];
+      var foodToMerge = {
+        vegetables: vegetables,
+        greens: vegetables
+      };
+      var food = {
+        vegetables: ['carrot']
+      };
+
+      var actual = _.mergeWith(food, foodToMerge, function(a, b) {
+        return lodashStable.isArray(a) ? a.concat(b) : undefined;
+      });
+
+      assert.deepEqual(actual, { vegetables: ['carrot', 'peas', 'brocolli'], greens: ['peas', 'brocolli'] });
+    });
   }());
 
   /*--------------------------------------------------------------------------*/


dannyfallon:
You can play around with an example of this on [this jsFiddle](http://jsfiddle.net/nwvxn9uq/2/) and this behaviour has been present for quite some time since I initially discovered it on v2.4.1

dannyfallon:
@jdalton I lost about 3 hours to this one, I'd really love if you could explain to me what's invalid about my expectation or the behaviour of _mergeWith that makes this impossible.

jdalton:
And I've put 4yrs into the project. So maybe let me get through something before jumping down my gd throat.

jdalton:
Hey @dannyfallon! Thanks for the failing test repro

This is a known limitation of our circular reference detection. I'm not sure how to go about resolving it at this time. If I find a way to fix it then your PR will be merged. Have you signed [our cla](http://dojofoundation.org/about/cla)? That would make things quicker to merge if a solution is found.

jdalton:
Actually it was easier to fix than I thought. If you've signed the cla I'll merge.

dannyfallon:
Thanks for the explanation and for the fix, I've been thinking about this one on and off all day :+1: I've signed the CLA just now, my Dojo name is `dannyfallon`.

jdalton:
:rocket: !

PR #2061, opened 2016-02-26
Remove custom rearg for `update`

jfmengels:
Remove custom rearg for `update` to have the same arg order as `set`
As discussed in #1832:
> How I see it:
the value argument in `update` is like the func argument in `set` (you could replicate set with update by passing `_.constant(someValue)` as func), so it'd make sense having them at the same position. So that, for example, if you need to modify your code to use `update` rather than `set`, you do not have to change the order of the arguments.
I also think it is more readable to have as first argument what you want to change, rather than how to change it.


jfmengels:
Thanks!

jdalton:
:coffee:

Issue #2065, opened 2016-02-28
Add _.toBoolean()


GreenPioneer:
Hello, 

I have a use case were i am parsing what the user sends me in the string of the url 
```javascript
// http://localhost:3000/api/v1/campaigns?where=funded&ne=true
req.query  =  { where:'funded' , ne:'true' }
// http://localhost:3000/api/v1/campaigns?where=paid&equals=false
req.query  =  { where:'paid' , equals:'false' }
// It always comes in as a string
```
In my Package [BuildREQ](https://github.com/GreenPioneer/buildreq) i will know based off the schema that it is a boolean field so all i need to do next is evaluate what the object is saying from the user. i was gonna just write a small checker but i already use lodash for everything so i figured i would contribute and hopefully see this get added.

GP

jfmengels:
When is see `toBoolean`, the first thing I think of is an equivalent to `!!value`. In this case, especially in regard of strings (and the implementation I see in #2066), it's more like a `parseBoolean`, isn't it?

jdalton:
Hi @GreenPioneer!

This looks like a better fit as a [mixin](https://lodash.com/docs#mixin).

StreetStrider:
@jfmengels, this time it's not `!! value`.
What he wants is not actually a `!! value`, but `value == 'true'`.

jfmengels:
Yeah. I just meant it was a confusing name, since I would think that toBoolean is ~ equivalent to `!!value`

GreenPioneer:
Hey  
I looked at what all of you were saying and placed it in a [JsFiddle](https://jsfiddle.net/greenpioneer/rs9upLqd/) . @StreetStrider 
 hit the nail on the head though because of my use case i cant get away from not evaluating  a true string .
@jfmengels what would you think of calling this then in this use case ?

@jdalton is the js fiddle kinda what you had in mind?

Let me know you thoughts

jdalton:
> is the js fiddle kinda what you had in mind

Yep.

jfmengels:
@GreenPioneer Something like `parseBoolean`? In effect, what you are looking for is an improved `JSON.parse()`, so it would make sense IMO to call it that. Anyway, I agree that this would be a better fit as a mixin.

jwerre:
You could user underscore.string [`toBoolean`](https://github.com/epeli/underscore.string#tobooleanstring--boolean):

    npm install underscore.string
    _.mixin(require('underscore.string').toBoolean);

[...]

PR #2066, opened 2016-02-28
#2065 Added toBoolean()

GreenPioneer:
#2065 Added toBoolean(). It Trims, Replace & Lowercases strings that
need to be evaluated for being a Boolean. I also added the proper tests
cases too.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -10669,6 +10669,40 @@
     }
 
     /**
+     * Converts `value` to an Boolean.
+     *
+     * @static
+     * @memberOf _
+     * @category Lang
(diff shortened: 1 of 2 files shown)


jwhitmarsh:
@jdalton sorry to comment on an old PR, but following #2065 I'd like to implement this as a mixin but I can't see how to access `objectToString`. Is there a way to use this PR as a mixin, or does it need to be implemented differently?

jdalton:
Hi @GreenPioneer!

For the v4 releases you should be able to do `lodash/_objectToString`.

PR #2067, opened 2016-02-28
Update README.md

wezzy:
in the documentation _.eq(object, other); return true and not false as expected

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -4104,7 +4104,7 @@ _.eq(object, object);
 // => true
 
 _.eq(object, other);
-// => false
+// => true
 
 _.eq('a', 'a');
 // => true


jdalton:
Please update the source file and not the generated doc file.

jdalton:
The example is correct. In v4 the `_.eq` method is for [`SameValueZero`](http://ecma-international.org/ecma-262/6.0/#sec-samevaluezero) comparisons. In v4 it was an alias for `_.isEqual`.

wezzy:
Ops my bad thanks for the clarification I'm still using V3

jdalton:
No worries. I've updated the [changelog](https://github.com/lodash/lodash/wiki/Changelog#v400) with a note on this.

PR #2080, opened 2016-03-03
Add _.isEqualTo

not-an-aardvark:
This is similar to _.isEqual, but rather than accepting two parameters and returning a boolean, it accepts one parameter and returns a curried function. Its typical use would be as an iteratee for function parameters.

For example, the following two samples are equivalent:

```javascript
var object = { a: 1, b: 1, c: 2, d: 3 };
// Select only the properties that have value 1
_.pickBy(object, function(value) {
  return value === 1;
})
// => { a: 1, b: 1 }
```
```javascript
var object = { a: 1, b: 1, c: 2, d: 3 };
_.pickBy(object, _.isEqualTo(1))
// => { a: 1, b: 1 }
```

I think this would be a convenient shorthand for common boilerplate code.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -9918,6 +9918,24 @@
     }
 
     /**
+    * Creates a function that checks whether its provided argument is deeply
+    * equal to `value`.
+    *
+    * @static
+    * @memberOf _
+    * @category Function
+    * @param {*} value The value to compare.
+    * @returns {Function} Returns the new function.
+    * @example
+    *
+    * _.filter([1, 2, 1, 3, 1, 4], _.isEqualTo(1))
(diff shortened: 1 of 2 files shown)


jdalton:
Hi @not-an-aardvark 

You're not fooling me


For this you can use our `fp` module:
```js
var isEqual = require('lodash/fp/isEqual');
_.pickBy(object, isEqual(1))
```
For simple comparisons of primitives you can use [_.eq](https://lodash.com/docs#eq) or the `fp` form of it. For more info on the `fp` module see the [FP Guide](https://github.com/lodash/lodash/wiki/FP-Guide).

PR #2083, opened 2016-03-03
Add reconvert method to fp functions

epeli:
Today I was happily hacking using the cherry picked `lodash/fp/<func>` functions but then I came to a situation where I really needed the second iteratee argument `key` of the `mapValues` function.

I seem to have two options:

1. Use the non-fp version
2. Manually do the conversion

But I find both ugly and I'm pretty sure it will be very confusing to whomever will be reading my code next time. So I propose the a `.convert(options)`  method to be added to all fp functions which will allow reconverting the function like this:

```js
import mapValues from "lodash/fp/mapValues";
const mapValuesWithKey = mapValues.convert({cap: false});

// or even simpler with commonjs
const mapValuesWithKey = require("lodash/fp/mapValues").convert({cap: false});
```

What do you think?

I'm not sure this is the best way to implement it. It works, but consider this PR just as a proposal for a feature like this.

### DIFF
--- a/fp/_baseConvert.js
+++ b/fp/_baseConvert.js
@@ -279,6 +279,9 @@ function baseConvert(util, name, func, options) {
       setPlaceholder = true;
       func.placeholder = result.placeholder = placeholder;
     }
+    result.convert = function(options) {
+      return baseConvert(util, name, func, options);
+    };
     return result;
   };
 
--- a/lib/fp/template/doc/wiki.jst
+++ b/lib/fp/template/doc/wiki.jst
@@ -26,8 +26,25 @@ var extend = require('lodash/fp/extend');
 ## Convert
 
 Although `lodash/fp` & its method modules come pre-converted, there are times when
-you may want to convert another lodash package or create a customized conversion.
-That’s when the `convert` module comes in handy.
+you may want to customize the conversion. That’s when the `convert` method comes in handy.
+
+
+```js
+// Disable capping of the iteratee arguments so it is possible to access
+// the `key` argument
+var mapValuesWithKey = require("lodash/fp/mapValues").convert({cap: false});
+
+mapValuesWithKey(function(value, key) {
+  if (key === "foo") {
+    return -1;
+  }
+  return value;
+}, {foo: 1, bar: 1});
+// => {foo: -1, bar: 1}
+```
+
+It's also possible to use the convert function directly to convert functions or
+objects manually.
 
 ```js
 var convert = require('lodash/fp/convert');
@@ -45,7 +62,7 @@ var fp = convert({
 var fp = convert(lodash.runInContext());
 ```
 
-It’s even customizable.
+It’s customizable too.
 ```js
 // Every option is `true` by default.
 var filter = convert('filter', _.filter, {


jdalton:
That is super cool!

jdalton:
I like it!

jdalton:
Thoughts? @billyjanitsch @jfmengels

megawac:
Will it extend the default convert rules for a method with the ones provided in options?

epeli:
This implementation doesn't extend but I think it should.

jfmengels:
> I'm pretty sure it will be very confusing to whomever who will be reading my code next time

I think this will too ^^'
That said, it's much simpler than converting all of Lodash if you only need one such method. I think it can be a pretty useful tool though, for instance for mutating methods where you need a small change in the options (though I can't think of any off the top of my head).
I hope users won't abuse it too much as it is IMO still more complicated than re-importing the non-fp method, but it can be a nice addition :+1:

jdalton:
@megawac @epeli 
>> Will it extend the default convert rules for a method with the ones provided in options?

> This implementation doesn't extend but I think it should.

:+1:

@epeli 
Could you also modify the convert blurb in the [FP Guide template](https://github.com/lodash/lodash/blob/master/lib/fp/template/doc/wiki.jst).

epeli:
> Could you also modify the convert blurb in the FP Guide template.

Added!

I took a look on how to use the existing mapping as the base but it's kinda tricky because the default options in the mapping file are reversed.

---


@jfmengels
> I think this will too ^^'

Fair point. But

> I hope users won't abuse it too much as it is IMO still more complicated than re-importing the non-fp method, but it can be a nice addition

I think the key is consistency. If everywhere in the code base the fp versions are used and just in one file a function is imported from `lodash/fun` and not `lodash/fp/fun` that's really easy to miss. It can be very confusing if some of the functions are used data first and some not. 

```js
const mapValuesWithKey = mapValues.convert({cap: false});
```
Might not be self explanatory but the variable name will give you a clue and that's much harder to miss. It explicitly says that we now do something funny with it (go check to docs).

jfmengels:
@epeli :+1:

fp/_baseConvert.js
```diff
@@ -279,6 +279,9 @@ function baseConvert(util, name, func, options) {
       setPlaceholder = true;
       func.placeholder = result.placeholder = placeholder;
     }
+    result.convert = function(options) {
+      return baseConvert(util, name, func, options);
```
jdalton: can you make this build off the existing options.

epeli: The `config` object in that scope has all it's [values set to true](https://github.com/lodash/lodash/blob/89f891ac86aa1b81a47598088c5188b4c358a4e6/fp/_baseConvert.js#L35-L40).

It seems that the mapping file is queried only when the actual conversion is being done. So `assing({}, config, options)` won't work here.

jdalton: If `convert` was used previously then `config` may have non-`true` values.

jdalton: A test for this could be something like `mapValues.convert({cap:false}).convert({ rearg: false})`.

epeli: I actually tried that and it didn't work. While skimming the file didn't see any code mutating the `config` after it's being created.

epeli: Wait. Doesn't this actually already extend the mapping file rules as hoped?

epeli: But now I see that it remembers the last customization rules. Isn't that a bad thing?

Let's assume all the default rules are true for the function `fun`

```js
var fun1 = require("lodash/fp/fun").convert({cap: false})
var fun2= require("lodash/fp/fun").convert({rearg: false})
```
I would expect that `fun2` has `cap:true` but it seems that it would be `false`?

epeli: Sorry, I got confused by the chaining here. I scrapped my misled comments.

jdalton:
@epeli As the Dojo/jQuery foundations are working through there merge could you please sign the [jQuery Foundation CLA](https://contribute.jquery.org/CLA/).

epeli:
Signed.

jdalton:
Thanks!

olsonpm:
Any reason you guys landed on cap taking a boolean as opposed to an integer?  The only two I could think of were complexity and that convert builds from previous conversions - where capping multiple times with different integers wouldn't make a lot of sense.  Considering the fp functions cap at various arities however, I thought an integer would be more fitting.

jdalton:
> Any reason you guys landed on cap taking a boolean as opposed to an integer?

I thought about adding that at the time but opted for simplicity and consistency. I figured I'd wait for it to be requested before digging into it. You should open an issue to request it so we can track interest in it.

olsonpm:
Will do - and I'm not downplaying simplicity as maintaining two cases would be much simpler than a varying amount.

PR #2093, opened 2016-03-07
Debounce bug

### LINKED ISSUE #2089: debounce with maxWait starts emitting on "leading" edge after first time
This is using lodash 4.6.1.  If I use code like the following:

```es6
var d = _.debounce(func, 1000, { maxWait: 2000 });

// or, this way:
// var d = _.debounce(func, 1000, { maxWait: 2000, leading: false, trailing: true });
```

Then the first time I call d() a bunch of times, it works exactly like I'd expect:

`func` is not initially called.  But it is called every 2000ms until the activity stops for at least 1 second, then it is called a final time.

But, if after that initial test, I call `d` again, `func` is called immediately.  If I then wait a few seconds and call `d` again, `func` again is called immediately.  In other words, after the first burst of activity, subsequent bursts cause my function to get called on both the leading and trailing edge.

Here's a jsFiddle showing the problem: https://jsfiddle.net/m2LgLxtL/

And here's an interactive version: http://jsfiddle.net/ex9ej87k/2/
Click a bunch on the 2nd meter then stop.  It works as expected.  Now click a single time and the next marker highlights immediately.  Wait a while.  click again.  It highlights immediately again...and so on.

bman654:
fixes #2089

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -9000,14 +9000,12 @@
      * jQuery(window).on('popstate', debounced.cancel);
      */
     function debounce(func, wait, options) {
-      var args,
-          maxTimeoutId,
+      var lastArgs,
+          lastThis,
+          lastCallTime = 0,
+          timerId,
+          lastInvokeTime = 0,
           result,
-          stamp,
-          thisArg,
-          timeoutId,
-          trailingCall,
-          lastCalled = 0,
           leading = false,
           maxWait = false,
           trailing = true;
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -21289,9 +21357,13 @@
       }, 192);
 
       setTimeout(function() {
+        assert.strictEqual(callCount, 1);
```
bman654: Once I fixed the other bug, this test started failing.  After reviewing the test, it seems like it had some bad assertions.  Or maybe I am misunderstanding the way throttle works.

What do you think?

jdalton:
@bman654 Thanks for this! Have you [signed our CLA](https://contribute.jquery.org/CLA/)?

bman654:
yes

jdalton:
:rocket: !

test/test.js
jdalton: I'm not sure how an added assert changes the behavior of a failing test.
Can you explain a bit more what the change does?

lodash.js
```diff
@@ -9000,14 +9000,12 @@
      * jQuery(window).on('popstate', debounced.cancel);
```
jdalton: The debounce code can be a bit complex and understanding the ins and outs of its async-ness can be challenging at times. Would you mind walking me through the changes step by step (baby step it)?

test/test.js
bman654: I changed the timings of the tests.  The commit message for this change has a bit more details: 43d71efd6e56d9f4041434737fe8b3286b780296

lodash.js
bman654: Sure.  Let me add an additional commit that adds some more detailed comments.  I could also walk you through the code in a screenshare.

jdalton: > I could also walk you through the code in a screenshare.

:heart_eyes:

jdalton:
Thanks for the screenshare!

bman654:
I studied #1952 use case and the OP was describing a sustained period of activity and the `maxWait` firings were not very accurate.

I took his code and applied it to this new version (replaced his click handlers with setInterval).

https://jsfiddle.net/7atuxyc3/

It looks like the new code is just as accurate.  The throttle calls are occurring every 400ms +/- 8ms, which is all the accuracy you can hope for when using setTimeout.

lodash.js
```diff
@@ -9022,95 +9020,137 @@
...
 
-      function maxDelayed() {
-        complete(trailing, timeoutId);
+      function checkTimes(time) {
+        var timeSinceLastInvoke = time - lastInvokeTime,
+            waitTime = time - lastCallTime,
+            remainingWait,
+            shouldInvoke;
+
+        if (waitTime >= wait) {
```
megawac: Is this condition for time zone switches?

jdalton: That bit is a little lower

Issue #2103, opened 2016-03-12
repeat should break when n equals 0


libreninja:
_.repeat(string, n) concatenates the input string with itself on every iteration of the do-while loop, including the last iteration when the concatenated result will never be used. If n, or the the length of the input string is large enough, repeat will generate a "RangeError: Invalid string length" exception. Breaking before `string += string` in the final iteration would prevent exceptions in some cases.

```javascript
do {
  if (n % 2) {
    result += string;
  }
  n = nativeFloor(n / 2);
  
  // guard against unnecessary "Invalid string length" exceptions
  if(n === 0) {
    break;
  }
  string += string;
} while (n);
```

jdalton:
Thanks, moving to #2104.

PR #2104, opened 2016-03-12
break before unnecessary string concatenation

libreninja:
Pull request for issue #2103. Break before string concatenation that will never be used.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -13116,6 +13116,9 @@
           result += string;
         }
         n = nativeFloor(n / 2);
+        if(n === 0) {
+            break;
+        }
         string += string;
       } while (n);
 


jdalton:
Hi @libreninja! (I'm in Woodinville btw :P) 

Have you signed [our CLA](https://contribute.jquery.org/CLA/)?
Would you mind creating a unit test for the range error if possible.

libreninja:
I haven't signed the CLA yet. I'll sign that and write a test shortly.

libreninja:
I signed the CLA.

I'm not sure how to go about unit testing this change, as the range error depends on the max string length which is implementation specific.

PR #2132, opened 2016-03-19
Re-ordered arguments for fp.isEqualWith + some additional tests for lodash.isEqualWith

mhelvens:
Here's the fix. It seems to work OK.

I wasn't sure how many tests you'd expect for an FP version of an existing operation (there aren't any tests for `fp.isMatchWith`), so I converted some of the basic ones from `test.js`.

Also, I was surprised that the original `lodash.isEqualWith` is tested for expected positive results, but not for expected negative results, so I added some additional tests there, too.

### DIFF
--- a/fp/_mapping.js
+++ b/fp/_mapping.js
@@ -143,6 +143,7 @@ exports.methodRearg = {
   'assignInWith': [1, 2, 0],
   'assignWith': [1, 2, 0],
   'getOr': [2, 1, 0],
+  'isEqualWith': [2, 1, 0],
   'isMatchWith': [2, 1, 0],
   'mergeWith': [1, 2, 0],
   'padChars': [2, 1, 0],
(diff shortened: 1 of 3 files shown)


jdalton:
Thanks! Have you signed [our CLA](https://contribute.jquery.org/CLA/)?

mhelvens:
I have now.

Pf... legal stuff. Just FYI, I technically did not waive any copyrights by signing the CLA, because this project does not have a `LICENSE.txt` file. :-)

jdalton:
We have a `LICENSE`. But I'm fine not excepting your contribution.

mhelvens:
You closed the pull request, so I'm afraid you're actually being serious. What's wrong?

jdalton:
One) You're probably not a lawyer.
Two) I've raised the wording issue with the foundation.
Three) There's lots of ways to raise a concern and the way you did was off putting so I'm ending it.

PR #2145, opened 2016-03-22
Make _.isEmpty work as expected on buffers

fdegiuli:


jdalton:
Hi @fdegiuli!

Thanks for this PR! I can dig it.
You're the first dev to hit the jQuery bot. It's checking if you've signed [our CLA](http://contribute.jquery.org/CLA/)

jdalton:
On review a buffer will be detected as array-like by `isArrayLike` so should hit the code path anyways (tested in Node 0.10 and Node 5.0).

fdegiuli:
The problem is that the checks after `isArrayLike` only allow for arrays, strings, functions, and arguments (because of the `&&`) So even though a buffer `isArrayLike`, it never reaches the return statement.

Here's what it looks like in the cli:

```
nesh> var u = require('lodash');
undefined
nesh> buf = new Buffer('');
<Buffer >
nesh> buf.length
0
nesh> u.isArrayLike(buf);
true
nesh> u.isEmpty(buf)
false
```

I have a new commit that clarifies this and adds a test case, I can open a new PR, if you want. (I guess github doesn't update closed PRs)

jdalton:
Ah, thanks for clarifying.
We don't currently support iterating buffers so returning `false` for `_.isEmpty` would be misleading:

> Checks if `value` is an empty object or collection. A value is considered
empty if it's an `arguments` object, array, string, or jQuery-like collection
with a length of `0` or has no own enumerable string keyed properties.

fdegiuli:
I don't quite follow. Right now calling `isEmpty` with a buffer _always_ returns `false`.  Are you saying that we should change it, or that `isEmpty` shouldn't work on buffers at all?  In that case, maybe `false` is the wrong thing to return anyway?

jdalton:
Wow hmm I'm messing this up pretty well :)
I forget that buffers in Node land are just typed arrays and we support typed arrays. You shouldn't need to remove the `isArrayLike` check just add an `isBuffer` check to the mix.

jdalton:
@fdegiuli Rock now it's just the CLA bit.

fdegiuli:
Ah, of course. Buffers only exist in node. Let me fix that.

jdalton:
:tada: Thanks for your patience!

fdegiuli:
My pleasure. Thanks for the quick replies.

fdegiuli:
Looks like the build is failing because I used `assert(true)` instead of `skipAssert` I've fixed it if you want to reopen this a second time... Sorry about that.

fdegiuli:
Wait, it's been merged. Should I open a new PR?

jridgewell:
@fdegiuli: He's [already fixed it](https://github.com/lodash/lodash/commit/43942356e85385d1a4570e3728025632ea8af122). :wink:

fdegiuli:
Sweet.

jdalton:
No worries. I've got your back.

PR #2156, opened 2016-03-26
Adds support for _.clone(DataView)

justsml:
I feel this might need better test coverage for both NodeJS & Browser... Any suggestions?

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -85,6 +85,7 @@
       weakSetTag = '[object WeakSet]';
 
   var arrayBufferTag = '[object ArrayBuffer]',
+      dataViewTag = '[object DataView]',
       float32Tag = '[object Float32Array]',
       float64Tag = '[object Float64Array]',
       int8Tag = '[object Int8Array]',
@@ -231,7 +232,7 @@
 
   /** Used to assign default `context` object properties. */
   var contextProps = [
-    'Array', 'Buffer', 'Date', 'Error', 'Float32Array', 'Float64Array',
+    'Array', 'Buffer', 'DataView', 'Date', 'Error', 'Float32Array', 'Float64Array',
     'Function', 'Int8Array', 'Int16Array', 'Int32Array', 'Map', 'Math', 'Object',
     'Reflect', 'RegExp', 'Set', 'String', 'Symbol', 'TypeError', 'Uint8Array',
     'Uint8ClampedArray', 'Uint16Array', 'Uint32Array', 'WeakMap', '_',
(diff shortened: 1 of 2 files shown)


lodash.js
```diff
@@ -3958,11 +3960,26 @@
...
+
+    /**
+     * Creates a clone of `dataView`.
+     *
+     * @private
+     * @param {DataView} dataView The data view to clone.
+     * @returns {DataView} Returns the cloned data view.
+     */
+    function cloneDataView(dataView) {
+      var buf = cloneArrayBuffer(dataView.buffer, dataView.byteOffset),
```
jdalton: Instead of cloning the `byteOffset` of the buffer wouldn't you clone the buffer whole and then provide the offset as:
```js
new dataView.constructor(buf, dataView.byteOffset, dataView.byteLength);
```

lodash.js
```diff
@@ -85,6 +85,7 @@
       weakSetTag = '[object WeakSet]';
```
jdalton: You should add `DataView` to the `contextProps` array too.

jdalton:
> I feel this might need better test coverage for both NodeJS & Browser... Any suggestions?

Ya, but don't worry about that. I'll take that bit on.

lodash.js
justsml: My thought was it'd be less memory usage. 
I may be wrong, I haven't done much profiling with these parts of JS.

jdalton: I've punted on optimized routes for now (because its such an edge case at the moment).

test/test.js
```diff
@@ -2634,6 +2636,20 @@
...
       });
 
+      QUnit.test('`_.' + methodName + '` should clone dataviews', function(assert) {
+        assert.expect(2);
+
+        if (DataView) {
+          var actual = func(dataView);
+
+          assert.strictEqual(actual.byteLength, dataView.byteLength);
+          assert.notStrictEqual(actual, dataView);
```
megawac: can you also check that it's  a `DataView` instance

justsml:
I am assuming that a cloned DataView should re-create the underlying view-able buffer, and then return a new DataView w/ the byteOffset now 'reset' to 0 with the new buffer as it's contents.

Am I making too many assumptions?

lodash.js
justsml: Also at first I thought I'd just make `cloneArrayBuffer` use `TypedArray.slice` - it seemed like an efficient way to get the data. 
@jdalton  I am guessing it isn't used because of IE10 support?

lodash.js
justsml: Added :heavy_check_mark:

justsml:
My thoughts on how this should behave are:

* Considering `DataView` doesn't contain much more than a couple integers (roughly speaking):
    1. **Pointer to memory addr** - potentially referencing a huge ArrayBuffer
    1. **Integer byteOffset** - which might limit what the DataView sees to, let's say only 8 bytes, out of perhaps many MBs.
    1. **Integer Length** - which reminds me, I forgot to account for this :astonished: 

* Would it be better to rely on `clone vs. cloneDeep` to achieve something like:
    * `_.cloneDeep(dv)` would copy the underlying buffer, resetting the offset & length? 
    * `_.clone(dv)` would simply `return new DataView(dv.buffer, dv.byteOffset, dv.byteLength)`. I think this seems a little pointless though - and it may not be what a typical user would expect?

lodash.js
jdalton: Correct. Older-modernish environments don't support it.

jdalton:
No worries. You've done a lot of work here. I can take it the rest of the way. The `DataView` is just another view like a typed array so can reuse the `cloneTypedArray` method (which I'll rename to `cloneArrayView`. The idea is if it's a shallow clone then just the view is cloned leaving the original array buffer in place. If it's a deep clone then the view and the buffer are cloned. No need to muck with resetting offsets we use the values of the original view being cloned.

justsml:
@jdalton After thinking about it more, cloning a `view` type seems pointless - because you can't do:
```js
// !!!! Non-example/pattern
var section1 = _.clone(dv);
section1.byteOffset += 1024;
var section2 = _.clone(section1);
section2.byteOffset += 1024;
```

I can't think of a real use-case for shallow cloning a `DataView`
If I want to clone the buffer, I'll do that.

But if I want to clone a DataView, **surely I mean to get a new DataView w/ a copy of the visible buffer?**

```js
// so in code-words:
var dv1 = _.clone(someDataView);
// Is almost no different than doing this:
var dv1 = someDataView;
// DataView members are immutable after all
```

jdalton:
Closed by https://github.com/lodash/lodash/commit/f6d8036a5e7a4b0945f79b0a021bf492dcfeaf96.

justsml:
@jdalton Did my last msg make sense? Or am I missing something here?

jdalton:
> Did my last msg make sense? Or am I missing something here?

Shallow cloning has limited value in some cases but it's a shallow clone nonetheless.

justsml:
Ok. Uniformity makes for a good argument.

For my presumptive use-case, to extract ONLY the data presented by the DataView, should something like this be possible:
```js
var freshView = _.castArray(dataView);
```

Issue #71, opened 2012-09-13
put lazy bind() in its own function; add lazy bind to bindAll()


benatkin:
I think that lazy bind is useful, but I think its current implementation makes for a confusing API. The purposes of the first and second arguments seemingly get switched. Here's what's really happening:
- _normal bind_
  - _first argument_ identifies function
  - _second argument_ identifies context object
- _lazy bind_
  - _first argument_ identifies context object and function
  - _second argument_ identifies function

http://lodash.com/docs#bind

So in both cases the first object is being used to identify the function. I think it would be better to reverse the arguments in lazy bind. The first argument, instead of _function or object_, would be _function or string_.
## 

But really I think it's better put into a separate function, to keep the API simple. I also think this would be equally useful, if not more so, for bindAll, and it would be harder to adjust the arguments to keep it down to one function. So it would also be something to move to a different function.

I apologize for overreacting on twitter.

PR #2177, opened 2016-03-30
ensure all background jobs finish before continuing

craigmichaelmartin:
Multiple scripts can be run in parallel by running all but one in the background. EG, `sleep 1` runs sleep in the foreground, `sleep 2 &` runs sleep in the background, and `sleep 2 & sleep 1` runs `sleep 2` in the background and `sleep 1` in the foreground (hence, in parallel). Currently in lodash, this is being utilized for scripts that can be run in parallel, but there is a racetime condition with the script being run in the foreground.

For example `echo 'sleeps beginning' && sleep 7 & sleep 1 && echo 'sleeps finished'` demonstrates this racetime condition where the process running in the foreground finishes before the process running with it in parallel in the background.

This is not demonstrating itself currently with the scripts employing background processes because the foreground process (the last piece of the script) happens to take longer. The place where you would see it would be with `npm test` which waits for `npm run build`. If `npm run build:fp` finishes before `npm run build:main`, the testing will begin with build:main incomplete.

Using `wait` solves this problem. Eg, `echo 'sleeps beginning' && sleep 7 & sleep 1 & wait && echo 'sleeps finished'`

```bash
cmartin:lodash martin$ time (echo 'sleeps beginning' && sleep 7 & sleep 1 & wait && echo 'sleeps finished')
sleeps beginning
sleeps finished

real	0m7.015s
user	0m0.003s
sys	0m0.006s
```

vs 

```bash
cmartin:lodash martin$ time (echo 'sleeps beginning' && sleep 7 & sleep 1 && echo 'sleeps finished')
sleeps beginning
sleeps finished

real	0m1.011s
user	0m0.001s
sys	0m0.004s
cmartin:lodash martin$ ps aux | grep sleep
martin         15895   0.0  0.0  2423356    196 s002  R+    2:03PM   0:00.00 grep sleep
martin         15893   0.0  0.0  2432752    512 s002  S     2:03PM   0:00.00 sleep 7
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "webpack": "^1.12.14"
   },
   "scripts": {
-    "build": "npm run build:main & npm run build:fp",
+    "build": "npm run build:main & npm run build:fp & wait",
     "build:fp": "node lib/fp/build-dist.js",
     "build:fp-modules": "node lib/fp/build-modules.js",
     "build:main": "node lib/main/build-dist.js",
@@ -41,7 +41,7 @@
     "doc:fp": "node lib/fp/build-doc",
     "doc:site": "node lib/main/build-doc site",
     "pretest": "npm run build",
-    "style": "npm run style:main & npm run style:fp & npm run style:perf & npm run style:test",
+    "style": "npm run style:main & npm run style:fp & npm run style:perf & npm run style:test & wait",
     "style:fp": "jscs fp/*.js lib/**/*.js",
     "style:main": "jscs lodash.js",
     "style:perf": "jscs perf/*.js perf/**/*.js",
@@ -49,7 +49,7 @@
     "test": "npm run test:main && npm run test:fp",
     "test:fp": "node test/test-fp",
     "test:main": "node test/test",
-    "validate": "npm run style & npm run test"
+    "validate": "npm run style & npm run test & wait"
   },
   "config": {
     "ghooks": {


jdalton:
Thanks @craigmichaelmartin! Have you signed [our CLA](https://contribute.jquery.org/CLA/)?

craigmichaelmartin:
Sure thing, signed.

jdalton:
Thanks!

jdalton:
Awesome!

package.json
```diff
@@ -49,7 +49,7 @@
     "test": "npm run test:main && npm run test:fp",
     "test:fp": "node test/test-fp",
     "test:main": "node test/test",
-    "validate": "npm run style & npm run test"
+    "validate": "npm run style & npm run test & wait"
   },
```
jdalton: @kentcdodds I think this change to "validate" is causing [this fail](https://travis-ci.org/lodash/lodash-cli/jobs/119714235).

kentcdodds: Hmmm, actually it looks like it may be an opt-cli bug. Thoughts  @ta2edchimp?

kentcdodds: Alternatively, you might try npm.im/npm-run-all. I've had great success with that in my projects.

jdalton: Cool. I'm temp removing it (while I bump to ensure tests pass) and will re-activate it after.

ta2edchimp: Hmm... it looks like opt-cli's command `opt` is interfering with another command with the same name, thus the rendered output in the logs (line 184+). I will have a more thorough look soon.

ta2edchimp: The issue seems to be that `ghooks` does not find its own `opt-cli` dependency when running its own `install` script on `npm@"^2.0.0."`.

[ghooks issue #71](https://github.com/gtramontina/ghooks/issues/71)

ta2edchimp: After spending some time on it, I found this to be an issue related only to npm versions `2.15.*`— versions `≤ 2.14.22` and `≥ 3.0.0` are not affected.

PR #2180, opened 2016-03-31
Simplify _.result

captbaritone:
Avoid an unneeded assignment.

It's possible that you are doing this on purpose because returning `value` feels more readable. But in the off-chance that this was just an oversight, I figured I'd propose the change.


jdalton:
If the tests pass then it's good by me.

PR #2188, opened 2016-03-31
Set the type of `apply`'s `args` parameter to `Array`

### LINKED ISSUE #2187: Incorrect JSDoc for `apply`
I would argue that [this JSDoc](https://github.com/lodash/lodash/blob/master/lodash.js#L412) describing `apply`'s `args` parameter isn't correct. `args` isn't a rest-parameter, and shouldn't be documented as such.

(Discovered in Microsoft/TypeScript#7749)

Victorystick:
Fixes #2187

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -409,7 +409,7 @@
    * @private
    * @param {Function} func The function to invoke.
    * @param {*} thisArg The `this` binding of `func`.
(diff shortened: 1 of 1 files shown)


jdalton:
Thanks!

Issue #2215, opened 2016-04-06
in map, being abble to call method of iteratee


AquariusFR:
example :
```
_(collection)
.map('::toJson');
```

will do the same as :
```
_(collection)
.map(function(o){return o.toJson());
```

Florian-R:
See [_.invoke](https://lodash.com/docs#invoke) and [_.invokeMap](https://lodash.com/docs#invokeMap)

demoneaux:
Yep, for your use case you can run:

```js
_(collection).invokeMap('toJson').value();
```

jdalton:
For the cases that don't map well to `invoke` or `invokeMap` there's also [_.method](https://lodash.com/docs#method).
```js
_(collection).map(_.method('toJSON'));
```

AquariusFR:
A suger would be nice to have a lighter code.
But this solution fits well.
Thanks @jdalton  !

PR #2220, opened 2016-04-07
feature :

AquariusFR:
setting an iteratee starting with '::' will invoke _.method(iteratee)
This will permit a cleaner code:

`_(collection).map('::toJson')`
instead of
`_(collection).map(_.method('toJson'))`

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -8536,9 +8536,19 @@
      */
     function map(collection, iteratee) {
       var func = isArray(collection) ? arrayMap : baseMap;
-      return func(collection, getIteratee(iteratee, 3));
+      return func(collection, getIteratee(iterateeSugar(iteratee), 3));
     }
 
+    /**
+     * this method transform a iteratee accorging to suggar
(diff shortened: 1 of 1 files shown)


jdalton:
Thanks @AquariusFR! The standard proceedure is to hold off on PRs until there's enough [user demand for a feature](https://github.com/lodash/lodash/blob/master/.github/CONTRIBUTING.md#feature-requests).

jdalton:
You can customize this without it being added to core by doing:
```js
_.iteratee = _.wrap(_.iteratee, function(iteratee, func) {
  if (typeof func == 'string' && _.startsWith(string, '::')) {
     return _.method(func.replace(/^::/, ''));
  }
  return iteratee(func);
});
```

jridgewell:
@AquariusFR: Have you tried [`_.invokeMap`](https://lodash.com/docs#invokeMap) yet?

jdalton:
Related to #2215.

PR #2263, opened 2016-04-18
Update lodash.js

jcataluna:
When array is already sorted and if it has too many elements this change makes the function way more performant. In my case with an array with 300k elements sorted by timestamp makes a huge difference.


jdalton:
Thanks @jcataluna!

I think this would be better as a separate method but also a sorted array could be `[1, 1, 1, 2, 2, 3, 3]` and we'd want to remove all the values the predicate returns truthy for.

jcataluna:
Totally agree! 

Would be nice to see it released because my BigData solution depends on this.

I can't live without this library ;);) Very tanks!

PR #2286, opened 2016-04-29
Test documentation with markdown-doctest

Widdershin:
Hello!

I'm the author of [markdown-doctest](https://github.com/Widdershin/markdown-doctest). markdown-doctest runs all of the code examples present in your markdown files to ensure that they are runnable.

I have to congratulate you, as lodash is the first project I've ever integrated markdown-doctest with that didn't have any typos! It's also the first project where no examples needed to be skipped. 🎉 

I think it might still be useful as a means of detecting when the documentation needs to be updated because of breaking changes (and catching any future typos).

markdown-doctest currently tests the documentation of RxJS v5, most.js and xstream, among other libraries.

I'm aware that your examples are written in jsdoc and compiled to markdown, but it seems to work fine regardless.

In other projects I have set markdown-doctest to run after npm test (so that it would run as part of the CI build), but as it seems you regenerate your documentation manually it seems more appropriate in this case to run it after documentation generation as part of `npm run doc`. Perhaps we can also add it to the travis config.

Let me know what you think! I should note that if this pull request isn't for you, that's totally okay 😄. Your documentation is pretty flawless to begin with.

screen shot 2016-04-29 at 10 29 07 am

### DIFF
--- a/.markdown-doctest-setup.js
+++ b/.markdown-doctest-setup.js
@@ -34,7 +34,7 @@ module.exports = {
 
     cwd: __dirname,
     mainText: '',
-    data: {user: 'Quinoa'},
+    data: {user: 'mock'},
 
     document: {
       body: {
--- a/.markdown-doctest-setup.js
+++ b/.markdown-doctest-setup.js
@@ -1,4 +1,4 @@
-function noop () {}
+var _ = require(__dirname);
 
 function jQuery () {
   return {
@@ -14,23 +14,23 @@ jQuery.each = function (items, f) {
 
 module.exports = {
   globals: {
-    '_': require(__dirname),
+    '_': _,
 
-    asyncSave: noop,
-    addContactToList: noop,
-    calculateLayout: noop,
-    createApplication: noop,
-    updatePosition: noop,
-    sendMail: noop,
-    renewToken: noop,
-    batchLog: noop,
+    asyncSave: _.noop,
+    addContactToList: _.noop,
+    calculateLayout: _.noop,
+    createApplication: _.noop,
+    updatePosition: _.noop,
+    sendMail: _.noop,
+    renewToken: _.noop,
+    batchLog: _.noop,
 
     setImmediate: setImmediate,
     Buffer: Buffer,
     EventSource: function () {},
 
-    fs: {writeFileSync: noop},
-    path: {join: noop},
+    fs: {writeFileSync: _.noop},
+    path: {join: _.noop},
 
     cwd: __dirname,
     mainText: '',
--- a/.markdown-doctest-setup.js
+++ b/.markdown-doctest-setup.js
@@ -1,0 +1,53 @@
+function noop () {}
+
+function jQuery () {
+  return {
+    on: function (ev, cb) {
+      cb()
+    }
+  }
+}
+
+jQuery.each = function (items, f) {
+  items.forEach(f)
+}
+
+module.exports = {
+  globals: {
+    '_': require(__dirname),
+
+    asyncSave: noop,
+    addContactToList: noop,
+    calculateLayout: noop,
+    createApplication: noop,
+    updatePosition: noop,
+    sendMail: noop,
+    renewToken: noop,
+    batchLog: noop,
+
+    setImmediate: setImmediate,
+    Buffer: Buffer,
+    EventSource: function () {},
+
+    fs: {writeFileSync: noop},
+    path: {join: noop},
+
+    cwd: __dirname,
+    mainText: '',
+    data: {user: 'Quinoa'},
+
+    document: {
+      body: {
+        childNodes: [],
+        nodeName: 'body'
+      }
+    },
+
+    jQuery: jQuery,
+
+    element: {},
+    window: {}
+  },
+
+  babel: false
+}
--- a/lodash.js
+++ b/lodash.js
@@ -13908,12 +13908,6 @@
      * compiled({ 'user': 'pebbles' });
      * // => 'hello pebbles!'
      *
-     * // Use custom template delimiters.
-     * _.templateSettings.interpolate = /{{([\s\S]+?)}}/g;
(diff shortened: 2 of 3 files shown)


lodash.js
```diff
@@ -13939,6 +13933,12 @@
      * //   return __p;
      * // }
      *
```
Widdershin: This is the only part of this PR that I'm not stoked with.

I had to reorder these examples because changing the interpolate regex means that all the following examples don't work correctly. I understand that these are intended to be run in isolation but markdown-doctest runs them as one big example.

Another option would be to reset the value of interpolate after this example, but that would add unnecessary noise to the docs.

Usually I would just skip this example set but because we're compiled from jsdoc I can't do that in the normal way (prepending the example with `<!-- skip-example -->` in the markdown).

If you have any ideas I would love to hear them 😄

Widdershin:
(Also, I've just signed your CLA 😄)

jfmengels:
I started doing a similar thing with https://github.com/jfmengels/example-validator (but haven't finished it yet...).

Does it only test that the code is runnable, or that the examples actually match? I.e., does the following pass?
```js
1 + 2
// => 4
```
That's what I started testing, but couldn't figure out a good solution for random tests (`_.uniqueId()`, `_.sample()`, ...).
Sounds pretty useful even without that, good job!

Widdershin:
Hi @jfmengels!

markdown-doctest only checks that the examples can be executed without throwing an error. It does nothing to actually test the correctness of the code. I had originally intended when I set out to write markdown-doctest to check that the output matched, but so far the simplicity of not doing that has been a boon.

It's intended to mimic the experience a user will have when copying and running code examples. You'd be surprised by how many projects have documentation that is littered with syntax errors!

lodash.js
jdalton: No worries. I'll kick something around.

.markdown-doctest-setup.js
```diff
@@ -0,0 +1,53 @@
+function noop () {}
+
+function jQuery () {
+  return {
+    on: function (ev, cb) {
+      cb()
+    }
+  }
+}
+
```
jdalton: Why is there a jQuery shim?

.markdown-doctest-setup.js
```diff
@@ -0,0 +1,53 @@
...
+    setImmediate: setImmediate,
+    Buffer: Buffer,
+    EventSource: function () {},
+
+    fs: {writeFileSync: noop},
+    path: {join: noop},
+
+    cwd: __dirname,
+    mainText: '',
+    data: {user: 'Quinoa'},
```
jdalton: What's this user setting `Quinoa` about?

.markdown-doctest-setup.js
```diff
@@ -0,0 +1,53 @@
+function noop () {}
```
jdalton: You can just use `_.noop`.
```js
var _ = require('lodash');
```

.markdown-doctest-setup.js
Widdershin: jQuery is used 12 times in the examples. We could add it to the devDependencies so we don't have to shim it, but this solution turned out to work quite well. Using the real jQuery would also mean we would need a browser context, or to shim out all the parts of document that jQuery uses.

jdalton:
Thanks for this @Widdershin!

> I have to congratulate you, as lodash is the first project I've ever integrated markdown-doctest with that didn't have any typos! It's also the first project where no examples needed to be skipped. 🎉

That's because of awesome contributors like yourself!

> In other projects I have set markdown-doctest to run after npm test (so that it would run as part of the CI build), but as it seems you regenerate your documentation manually it seems more appropriate in this case to run it after documentation generation as part of `npm run doc`. 

I totally :heart: the "test:doc" addition.

.markdown-doctest-setup.js
```diff
@@ -0,0 +1,53 @@
...
+
+jQuery.each = function (items, f) {
+  items.forEach(f)
+}
+
+module.exports = {
+  globals: {
+    '_': require(__dirname),
+
+    asyncSave: noop,
```
jdalton: What's all the mocks about?

jdalton:
I'm going to try to kick start our bot (ignore the close and reopen of the pr).

.markdown-doctest-setup.js
Widdershin: It's used in this example:

```js
// Use the `variable` option to ensure a with-statement isn't used in the compiled template.
var compiled = _.template('hi <%= data.user %>!', { 'variable': 'data' });
compiled.source;
// => function(data) {
//   var __t, __p = '';
//   __p += 'hi ' + ((__t = ( data.user )) == null ? '' : __t) + '!';
//   return __p;
// }
```

Quinoa is a bit of a running joke between myself and some of my coworkers. Usually when programmers pick names for examples, they pick something like 'Steve' or 'Mike'. We ended up picking 'Quinoa' as a name of one of our test customers, and it's been my go to ever since. I can change it to something more reasonable if you like!

.markdown-doctest-setup.js
Widdershin: Awesome, I'll update to use the builtin noop.

.markdown-doctest-setup.js
jdalton: Ah it's mocking out references from the docs. Niiice!

.markdown-doctest-setup.js
jdalton: Can you change it to "mock" so it's more clear there's no significance to it.

Widdershin: Done 😄

Widdershin:
All feedback addressed 😸

jdalton:
Thank you!

Widdershin:
Thank you @jdalton! Please let me know if you run into any issues, I'm happy to help. 😊

PR #2289, opened 2016-04-30
Remove duplicate method from FP mapping

jfmengels:
`find` was present multiple times in the `aryMethod` section of the FP mapping.


jdalton:
Thank you!

jfmengels:
Anytime ;)

PR #2332, opened 2016-05-12
Add `findIndexFrom` and `findLastIndexFrom` in FP.

### LINKED ISSUE #2329: Add `_.findIndexFrom` and `_.findLastIndexFrom` in Lodash/fp
The `fromIndex` param has been (re-)added to `_.findIndex` and `_.findLastIndex` in master, but the functionality is not available in `lodash/fp`.

I'm proposing `_.findIndexFrom(predicate, fromIndex, array)` for the param order.

Willing to try my hand at adding it :)

jfmengels:
Add `findIndexFrom` and `findLastIndexFrom` in FP. Closes #2329

I've added tests to make sure the iteratee was capped, wasn't sure that it would take the capping of its alias method, so better safe than sorry. And I've tried to make the tests look like the rest (`_.each`-y), I hope it doesn't make them too hard to understand.

### DIFF
--- a/fp/_mapping.js
+++ b/fp/_mapping.js
@@ -84,12 +84,12 @@ exports.aryMethod = {
   ],
   '3': [
     'assignInWith', 'assignWith', 'clamp', 'differenceBy', 'differenceWith',
-    'findIndexFrom', 'findLastIndexFrom', 'getOr', 'inRange', 'intersectionBy',
-    'intersectionWith', 'invokeArgs', 'invokeArgsMap', 'isEqualWith', 'isMatchWith',
-    'flatMapDepth', 'mergeWith', 'orderBy', 'padChars', 'padCharsEnd', 'padCharsStart',
-    'pullAllBy', 'pullAllWith', 'reduce', 'reduceRight', 'replace', 'set', 'slice',
-    'sortedIndexBy', 'sortedLastIndexBy', 'transform', 'unionBy', 'unionWith',
-    'update', 'xorBy', 'xorWith', 'zipWith'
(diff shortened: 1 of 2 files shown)


jdalton:
Up for adding `indexOf` and `lastIndexOf` forms too?

jfmengels:
Sure, `indexOfFrom` and `lastIndexOfFrom`? (Man that's a mouthful)

jdalton:
Yep!

jfmengels:
Tada!

jdalton:
Thanks!

jfmengels:
My pleasure :)

I see that `_.includes` also has a `fromIndex` method, might as well add it now too right?

By the way, I noticed that `_.find` doesn't have a `fromIndex` param that could be useful. Is that not available because it should also work on object?

jdalton:
> Is that not available because it should also work on object?

Right, which is why `includes` doesn't have it either. Collections methods. It's why in v5 we're splitting them all up.

jdalton:
Oh wait. We have it for `includes` and just shrug for objects. I guess we can do the same for `find` too.

jfmengels:
Yeah it has it. I'd be :+1: for that, sounds pretty useful to me

jfmengels:
Btw, are we :+1: with adding aliases to all methods with optional params? I see `result` that could have a `resultOr` friend, and I'm sure there are others. If we do want that, then I could make PRs as I find them.

jdalton:
> Btw, are we :+1: with adding aliases to all methods with optional params?

There are lots of variadic methods which could use alternatives. For now I'm going with the approach of waiting until the request comes in.

jfmengels:
Sure. Let me know if you want me to add others

PR #2344, opened 2016-05-17
Added method _.categorize

punit1108:
This method categorizes an array with heterogeneous elements according to their datatypes.

`_.categorize([1, "a", null, undefined, true, [1,2,3], { name: "Frank"}])`

gives `{ number: [ 1 ], string: [ 'a' ], object: [ null, [ 1, 2, 3 ], { name: 'Frank' } ], undefined: [ undefined ],boolean: [ true ] }`

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -6134,6 +6134,33 @@
       }
       return result;
     }
+    
+    /**
+     * Creates an object with its element categorized according to the datatype.
+     *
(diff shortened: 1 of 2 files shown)


jdalton:
Hi @punit1108!

I think this is a bit too specialized for core. It would make a great mixin though.

PR #2392, opened 2016-05-30
Update docs - _.every - no predicate argument

goldylucks:


jdalton:
Thanks @goldylucks!

I think the doc indicator of a default `predicate` of `_.identity` is enough here.

goldylucks:
it was due to [this](http://stackoverflow.com/questions/30900646/why-does-lodash-alltrue-true-true-true-return-false) SO thread

no worries :)

PR #2400, opened 2016-06-05
Useless export default rename

efender:
Eslint says: `Export default unnecessarily renamed.`


jdalton:
Thanks for the heads up @efender! But it should be:
```js
export default from './lodash.default.js';
```

Patched https://github.com/lodash/lodash-cli/commit/9fcf76468fb87a2297c1f8033bfcc9a3719a8065.

efender:
But if so `export { default } from './lodash.default.js';` ?
I found [stackoverflow question](http://stackoverflow.com/questions/32229947/is-there-any-one-line-analog-in-es6-for-es5-module-exports-require-inner-j). Babel and eslint are OK! )

PR #2422, opened 2016-06-15
Fixed display of transforms param in overArgs docs.

danjm:
There is currently a small bug in the documentation for `_.overArgs` that prevents the information about the `transforms` parameters from being displayed.

screen shot 2016-06-15 at 8 56 34 pm

This PR should fix that.


jdalton:
Thank you!

PR #2427, opened 2016-06-16
Add sortedFindBy and sortedLastFindBy functions

gudnm:
There's a pair of functions to search for a minimum index to insert an element while keeping array sorted - `sortedIndex` and `sortedIndexBy`. When it comes to checking whether element is in the array and returning its index when it is, there's no function that accepts an iteratee, just a lonely `sortedIndexOf`. It would be only logical to have `sortedIndexOfBy` or perhaps better named `sortedFindBy`. I also implemented `sortedLastFindBy`, to complement `sortedLastIndex` and `sortedLastIndexBy`.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -7475,6 +7475,43 @@
     }
 
     /**
+     * This method is like '_.sortedIndexOf' except that it accepts `iteratee`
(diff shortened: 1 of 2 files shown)


jdalton:
Thanks @gudnm!

PR #2428, opened 2016-06-16
Formalize which value is picked by _.unionBy and _.unionWith 

brandonhorst:
Formalize which value is picked by `_.unionBy` and `_.unionWith` to pick the result from the first array in which it occurs.

This is the way that the library has always worked - this simply adds that information to the tests and docs. This information is explicitly documented for `_.intersection*`.


jdalton:
Cool

PR #2453, opened 2016-06-28
Fixed broken link in README.md

realdubb:
Fixed broken link for lodash/fp in README

PR #2474, opened 2016-07-04
Fix link to "Contributing" doc

G07cha:


jdalton:
This is my bad for dragging out the release. See #2453.

PR #2495, opened 2016-07-12
Clarify _.sumBy iterating behaviour when using arrays and objects

pimterry:
The current `_.sumBy` behaviour with arrays and objects is not at all clear. I'm currently trying to fix up Lodash's type definitions for DefinitelyTyped, and I'm struggling to even work out what types are really valid, let alone how to use it.

After a little digging I think I've worked out how it works, so I've added details to iteratee explaining its behaviour generally, and a reference to that in `_.sumBy` explaining how that's used here.

I do think the array behaviour here (and probably everywhere that uses `_.iteratee`) is really kind of weird and surprising. For now, just documenting it seems best, but it might be worth considering future changes to that.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -14980,10 +14980,16 @@
 
     /**
      * Creates a function that invokes `func` with the arguments of the created
-     * function. If `func` is a property name, the created function returns the
-     * property value for a given element. If `func` is an array or object, the
-     * created function returns `true` for elements that contain the equivalent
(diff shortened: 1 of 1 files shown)


jdalton:
With v4.14.0 I'm simplifying the param signature and relying on code examples to get the gist across because it bloated the param signature and added noise. The iteratee shorthands are handled by `_.iteratee` and enable the rest.

PR #2507, opened 2016-07-18
Clarify _.isEmpty documentation

ryanve:
The phrase "string keyed" was confusing me


jdalton:
Thanks @ryanve!

Our use of string keyed is by design because there is now symbol keys in the mix.

PR #2525, opened 2016-07-24
closest() method implemented to find closest number in array

asduser:
```js
// Mock.
var collection = [0.12344, 0.12341, 0.12340], number = 0.12342;
```

```js
// Using "closest()" method, where 1st argument is a number, the 2nd one - an array of numbers
var result1 = _.closest(number, collection);   // will return a number '0.12341'
```

```js
// Arguments validation.
// If the first argument is not a number or the 2nd one is not an array, will be returned 'null'.
var result2 = _.closest("text", collection);   //null
var result3 = _.closest(number, 10);   //null
```

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -6335,6 +6335,32 @@
       return result;
     }
 
+    /*
+     * Finds closest number inside input collection.
+     * @param num {number} - required digit
+     * @param arr {array} - input set of digits
+     * @returns {number} - closest item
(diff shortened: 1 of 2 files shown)


megawac:
Have you considered using _.min?

_.min(collection, val => Math.abs(val - num))

jdalton:
Thanks @asduser!

PR #2530, opened 2016-07-25
Document behavior of _.debounce when wait is 0

islemaster:
Clarify that when `leading` is false a debounced method with `wait=0` will not invoke immediately, but on the next tick (like `setTimeout`), but when `leading` is true the invocation will be immediate.


jdalton:
Thanks!

Issue #999, opened 2015-02-25
Should _.matches(<emptything>)(<anything>) === true?


mpolichette:
Hey friends, love lodash, really great work!

I was just trying to use _.matches to do a _.reject.  However, sometimes my `source` object is undefined, and this causes the matches to always be true.  Is this by design?

I've walked the code and it looks like anything w/o any properties (E.G. empty array, empty object, undefined, null) will do this because it doesn't actually run any comparisons and instead immediately returns `true`.

If this is by design, I think there should be a note in the docs, since it is unexpected behavior to me.

Thanks!
Matt

megawac:
Yes, the behaviour was primarily designed for the case 

```js
var spec = {};
var obj = {a: 1};
_.matches(spec)(obj)
```

We decided this was true as every key:value pair is the same on the spec as the object as there are no keys on the spec. This is the same design as for `_.every` and `Array.prototype.every`, consider `[].every(_.constant(false))`

mpolichette:
Interesting, is there a function which behaves like _.matches but internally runs something more similar to `_.isEqual` to do the comparison?

I was trying to use `_.matches` as a convenience method for something like this:
```javascript
function deepWithout(array, value){
  var isValue = _.partial(_.isEqual, value);
  return _.reject(array, isValue)
}
```
Hoping i could replace that with a one liner
```javascript
_.reject(array, _.matches(value))
```
Am I overlooking a function lodash offers for that?

mpolichette:
I didn't mean to close :(

jdalton:
I'm in a live MVA course presenting all say so slow to respond

jdalton:
`_.match` performs a deep partial match do if you're wanting something like `_.isEqual` composition is your friend.

mpolichette:
Thanks for the fast and thorough response @megawac and @jdalton. You (and everyone helping) are doing a great job maintaining lodash.  I'm impressed.

PR #2548, opened 2016-07-31
Clarify empty source comparisons with _.matches, _.isMatch and _.matchesProperty.

danjm:
In the below code, that all the `falseMatches` are `false` is intuitive. However, it is much less intuitive that all the `trueMatches` are `true`.

```
var _ = require('lodash');
var assert = require('assert');

var objectToMatch = {a: {b: 2}};

var falseMatches = [
	_.isMatch(objectToMatch, {a: {d: null}}),
	_.matchesProperty('a', null)(objectToMatch),
	_.matchesProperty('a', [])(objectToMatch),
	_.matches({b: null})(objectToMatch)
];

assert.deepEqual(falseMatches, [false, false, false, false]);

var trueMatches = [
	_.isMatch(objectToMatch, {a: {}}),
	_.matchesProperty('a', {})(objectToMatch),
	_.matches(undefined)(objectToMatch),
	_.matches({})(objectToMatch)
];

assert.deepEqual(trueMatches, [true, true, true, true]);
```

A good explanation for the `trueMatches` was made here: https://github.com/lodash/lodash/issues/999#issuecomment-76028607.

This PR attempts to make this behavior obvious when reading the documentation.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -11344,8 +11344,9 @@
      * Performs a partial deep comparison between `object` and `source` to
      * determine if `object` contains equivalent property values.
      *
-     * **Note:** This method supports comparing the same values as `_.isEqual`
-     * and is equivalent to `_.matches` when `source` is partially applied.
+     * **Note:** If source is empty, this method returns true. Otherwise, it supports
+     * comparing the same values as `_.isEqual` and is equivalent to `_.matches`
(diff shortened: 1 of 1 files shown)


jdalton:
Thanks!

PR #2608, opened 2016-08-29
Update README.md

miczdem:
correction to forEach array example

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -2743,7 +2743,7 @@ or `_.forOwn` for object iteration.
 
 #### Example
 ```js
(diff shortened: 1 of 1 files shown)


jdalton:
Thanks @miczdem!

The use of chaining in the doc example was by design in Lodash v3 to show the special requirement of calling `.value`. This requirement was removed in Lodash v4 so we could tweak it.

That we'd modify the `lodash.js` source and not the generated docs.

miczdem:
I see.  I didn't realize that you could call the method in this way, just looked out of sync with the other examples.  cheers for clarifying it.

Issue #875, opened 2015-01-26
wish: document breaking changes in 3.0.0


markstos:
As a semantic version, 3.0.0 implies some sort of breaking change compared to 2.4 to 2.2. 

While a full ChangeLog would be nice, at least documenting compatibility issues from 2.2 to 2.4 and 2.4 to 3.0 would be helpful for those considering upgrading.

Thanks for lodash-- I use it nearly daily!

jdalton:
We're documenting the compat changes in our [changelog](https://github.com/lodash/lodash/wiki/Changelog) and releasing [lodash-migrate](https://github.com/lodash/lodash-migrate) to help give devs a headsup on compat issues in their code.

markstos:
@jdalton, thanks for the feedback. 

I was not aware that the wiki was in in use. Would you accept a patch to the [top level README](https://github.com/lodash/lodash/blob/master/README.md) that links to the docs available on the wiki? Something like this:

```markdown

## Further  Reading

  * Build Differences
  * Changelog
  * Resources
  * Roadmap
```

phated:
:+1: for referencing on the readme

jdalton:
@markstos Sure go for it. Don't forget to sign [the CLA](http://dojofoundation.org/about/claForm).

markstos:
I've now signed the CLA and submitted the pull request linked above.

PR #2648, opened 2016-09-15
Add CHANGELOG in the expected place and link it to where it's actually maintained.

markstos:
Ref: https://github.com/lodash/lodash/issues/875


jdalton:
Thanks!

PR #2716, opened 2016-10-07
Add repository to package.json

bhousel:
This allows tools to get information about the lodash repository. 
For example [greenkeeper](https://greenkeeper.io/) uses this link to include a diff summary in their pull requests. (e.g. "The new version differs by 12 commits.")

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,10 +5,7 @@
   "private": true,
   "main": "lodash.js",
   "engines": { "node": ">=4.0.0" },
-  "repository": {
-    "type": "git",
-    "url": "git://github.com/lodash/lodash.git"
-  },
+  "repository": "lodash/lodash",
   "scripts": {
     "build": "npm run build:main && npm run build:fp",
     "build:fp": "node lib/fp/build-dist.js",
--- a/package.json
+++ b/package.json
@@ -5,6 +5,10 @@
   "private": true,
   "main": "lodash.js",
   "engines": { "node": ">=4.0.0" },
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/lodash/lodash.git"
+  },
   "scripts": {
     "build": "npm run build:main && npm run build:fp",
     "build:fp": "node lib/fp/build-dist.js",


jdalton:
Thanks @bhousel! 

Can you use the shorthand form.

bhousel:
> Can you use the shorthand form.

Sure thing, I just updated it..

jdalton:
Do you know if greenkeeper would get that info from the package.json of the npm branch (the one we publish)?

bhousel:
> Do you know if greenkeeper would get that info from the package.json of the npm branch (the one we publish)?

Sorry I don't know!  I've just noticed that projects seem to need this property set, otherwise I don't get the summary.

cc @boennemann from greenkeeper team

boennemann:
We're using the metadata that's published to the npm registry, so that value should be present during `npm publish`. We're passing the value through https://github.com/tj/node-github-url-from-git to normalize it.
Additionally `npm publish` needs to happen from within the git repository, so that npm picks up the `gitHead` and attaches that to the published metadata as well.

As long as there is GitHub release and the tag `v${versionNumber}` (or without the v), we're also picking up release notes.

Thanks,
Stephan

jdalton:
I noticed the lodash update does seem to be missing the [commits summary](https://github.com/lodash/lodash-cli/pull/124):
https://github.com/katallaxie/generator-angular2-starter/pull/12
https://github.com/acburdine/gh-contrib-list/pull/24

jdalton:
@boennemann 
> Edit: I just checked, unfortunately we're not expanding the short form of the repo 😓 `repository.url` would have to be set.

You can use what `npm` uses to normalize data (and inflate `repo` shorthands):
[npm/normalize-package-data](https://github.com/npm/normalize-package-data)
```js
const normalizeData = require('normalize-package-data')
const packageData = fs.readFileSync("package.json")
normalizeData(packageData)
// packageData is now normalized
```

jdalton:
Thanks @bhousel! I'm going to close this as a greenkeeper issue. I think we're set from our end.

boennemann:
> You can use what npm uses to normalize data (and inflate repo shorthands):
> npm/normalize-package-data

Just now I remembered why we're not doing it – npm only stores normalized values in the registry :) 

`npm info lodash repository` returns what we expect.
The problem seems to be that `npm info lodash gitHead` doesn't return anything.
@jdalton Do you run your publishing scripts outside of the git repo?

jdalton:
Ah, we publish our [npm branch]()

jdalton:
So in that case then a `normalize-package-data` data would come in handy still as master is not the publish branch so info there could still need it.

boennemann:
> So in that case then a normalize-package-data data would come in handy still as master is not the publish branch so info there could still need it.

Not sure what you mean by that.

During `npm publish` npm looks up the HEAD sha of the current branch, of the repository it was executed in. It adds it as gitHead property to the metadata published: http://registry.npmjs.com/lodash
It looks like 4.11.2 is the only version where that worked before.

Greenkeeper adds a commit diff to the PR when the gitHead property is present for both the old and new version of the package, because only then we can figure out what the diff is.

jdalton:
> Not sure what you mean by that.

Ah. I `npm pack`, then untar, then cd into the `package` to review before publishing from there.

Our releases are tagged though [4.16.3...4.16.4](https://github.com/lodash/lodash/compare/4.16.3...4.16.4).

PR #2737, opened 2016-10-16
add CDNJS version badge in README.md

pvnr0082t:
This will show its version on CDNJS!


jdalton:
Hi @pvnr0082t!

I want to avoid badges. They clutter up things without much win.
I'd also like to stick with  1 CDN and jsdelivr is the most feature rich one out there.

pvnr0082t:
@jdalton okay, thanks for your explanation.

Issue #1459, opened 2015-09-12
How to get an object with all keys sorted?


stanleyxu2005:
Sometimes, I create an object (dictionary) with some index-able value as key. I want to make sure all keys are sorted alphabetically. Is there a built-in function of lodash to do that? Currently I have an own function to do that.
```
function objectWithKeySorted(object) {
  var result = {};
  _.forEach(Object.keys(object).sort(), function(key) {
    result[key] = object[key];
  });
  return result;
}
```
Example:
```
var a = {2015: '2015', 2011: '2011', 2013: '2013'};
console.log(a); // {2015: '2015', 2011: '2011', 2013: '2013'}
var b = objectWithKeySorted(a);
console.log(b); // {2011: '2011', 2013: '2013', 2015: '2015'}
```

jdalton:
Your example may not be the best as modern engines will auto iterate index keys in ascending order. A map (dictionary) really shouldn't be concerned with order. What are you using it for?

stanleyxu2005:
I have an array of elements. Every element has a property `timestamp`. One demand is to return all these elements without any duplicates and they should be sorted by timestamp. As object's keys will never duplicate, so I tried to return an `object` (key=timestamp) instead of an array. I hope `object` can sort all its keys as nature. But it does not work in that way.

jdalton:
You should check out ES6 Map (shimmable) combo'ed with Babel (transpiler) if needed.
It's keys are uniq (objects can even be keys):
```js
_.sortBy([...map], 0);
```

stanleyxu2005:
I know this method, but the result is an array. But I want object....

jdalton:
What's your keys look like?

stanleyxu2005:
Please check the code in the initial question. 

You inspired me. Actually I should do like this.
```
var a = {2015: '2015', 2011: '2011', 2013: '2013'};
_.forEach(Object.keys(a).sort(), function(key) {
  console.log(a[key]);
});
```

jdalton:
As I mentioned before if that's really your keys there is no sorted needed because that's how modern engines [iterate by](http://www.ecma-international.org/ecma-262/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys).
```js
var a = {2015: '2015', 2011: '2011', 2013: '2013'};
a[2010] = '2010';
Object.keys(a).forEach(function(key) {
  console.log(a[key]);
});
// logs 2010, 2011, 2013, 2015
```

[...]

PR #2741, opened 2016-10-17
added optional "keys" argument for ordered key retrieval from objects

sfinktah:
from my wonderings (wandering?) in https://github.com/lodash/lodash/issues/1459#issuecomment-253996722

..., on the topic of indeterminate Object ordering, it makes me wonder why:

    _.values(o, ['3a', 4])

Isn't **a thing**...

    /* @example
     *
     * o = {"3a": 1, "3b": 2, "4": 3, "4a": 4, "5": 5}
     * _.values(o)
     * // [3, 5, 1, 2, 4]
     *
     * _.values(o, ['3a', '3b', '4', '4a', '5'])
     * // [1, 2, 3, 4, 5]
     *
     * _.values(o, '3a', '3b', '4', '4a', '5')
     * // [1, 2, 3, 4, 5]
     */
    function values(object) {
      return object
        ? (arguments.length > 1)
          ? baseValues(object, flatten(arguments).slice(1))
          : baseValues(object, keys(object))
        : [];
    }


jdalton:
Hi @sfinktah!

Thanks for the PR. I'm not looking to expand the functionality of `_.values` beyond that offered by [Object.values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values) at this time.

sfinktah:
Obviously that's your choice, although I would point out effectively all I have done, is is expose the existing functionality of `baseValues` 2nd parameter, which is otherwise inaccessible. 

Is there some other method in the vastness of lodash that allows order-guaranteed value retrieve from an object?

I get an extremely uncomfortable feeling whenever I write

    [x, y] = _.pick(elem.getBoundingClientRect(), 'left, 'top');

and a little voice in the back of my head says: "Sure, that will work, but you know the order isn't actually guaranteed."

jdalton:
The `_.pick` method will return an object composed of the keys picked. If you're looking for something that offers what this PR offers, an array of key values, you can use [_.at](https://lodash.com/docs/#at).

sfinktah:
`_.at` would be horrifically inefficient, esp. if you wanted to pass in an array of property names.

In the past I have used `_.values(_.pick())`, but it's rather a double-whammy of "order not guaranteed" and a fistful of typing to boot.

Could you not simply expose `baseValues`  as `_baseValues` or somesuch?

jdalton:
Nope, [_.at](https://lodash.com/docs/#at) is the way to go. It even supports accessing deep property values.

sfinktah:
Cool.   I have another one that I find useful, possibly not a fit for lodash, but let me run in by you.

It's in the `_.range` `_.inRange` family, I guess I would call it `_.transpose` or somesuch.  I've been using it a lot lately.  

Basically:

        /**
         * transpose a number (or [nested] array of numbers) between two ranges.
         * @example 
         *     function random(low, high) = {
         *         return transpose(Math.random(), [0, 1], [low, high];
         *
         * @param {number|array<number>} transposee
         * @param {array<number>} r1 source range
         * @param {array<number>} r2 target range
         * @returns {number|array<number>} value(s) transposed to new range
         */
        function transpose(transposee, r1, r2) {
            if (_.isArray(value)) {
                return value.map(function(v, k) {
                    return transpose.call(null, v, r1, r2);
                });
            }
            return ( value - r1[0] )
                 * ( r2[1] - r2[0] )
                 / ( r1[1] - r1[0] )
                 +   r2[0];
        }

I just thought of the random() example then, and I have this awful feeling that if I go and look at the lodash source, I'm going to find another private function :p

PR #2742, opened 2016-10-18
Ensure that the passthrough to `Array.prototype.sort` does a stable sort

### LINKED ISSUE #1553: Enhancement: _.sort a stable sort algorithm
Difference environments implement  different Array#sort algorithms, some of which are stable, some unstable  and some which are stable for certain array lengths. I would like a implementation that is known stable full stop. I have one coded in my personal library that I would be willing to convert for lodash if you are interested.

murrayju:
[closes #1553]

I really wanted to be able to `_.chain()` in a `.sort()` with a custom comparator function, but couldn't stand that it wasn't a stable sort in V8 (and others). This patch does a native sort, but also compares the starting indexes when values are otherwise equal, resulting in a stable sort.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -16946,9 +16946,9 @@
     });
 
     // Add `Array` methods to `lodash.prototype`.
-    arrayEach(['pop', 'push', 'shift', 'sort', 'splice', 'unshift'], function(methodName) {
+    arrayEach(['pop', 'push', 'shift', 'splice', 'unshift'], function(methodName) {
       var func = arrayProto[methodName],
-          chainName = /^(?:push|sort|unshift)$/.test(methodName) ? 'tap' : 'thru',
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @murrayju!

Thanks for the PR :heart:.
For a more robust sort we have [_.orderBy](https://lodash.com/docs#orderBy) and [_.sortBy](https://lodash.com/docs#sortBy). With those I'd like to keep the `.sort()` method as a proxy to the real `Array#sort`.

murrayju:
Hey @jdalton,

It seems to me that `_.orderBy` and `_.sortBy` are limited in that you can't provide a custom comparator for two values. It seems that you are instead forced to produce a value for each iteratee that can be compared with the `<` and `>` operators. Is there some sort of workaround that I'm missing?

murrayju:
Alternatively, we could keep `.sort()` as a direct proxy to `Array#sort`, and add a `.stableSort()` method instead.

PR #2763, opened 2016-10-27
add two array function

aboutqx:
add 2 array function to find and omit object with key-value in array


aboutqx:
Should I add some unit test?

PR #2766, opened 2016-10-27
updated n_ related note: n_ is also tested with Node.js 6.2.1

borisdiakur:
see https://github.com/borisdiakur/n_/blob/master/.travis.yml#L7


jdalton:
Hi @borisdiakur!

I'm not suggesting `n_` isn't tested in Node v6. I'm saying I recommend it's use in Node < 6 because in Node 6 you can assign to `_` without the problems of earlier versions.

PR #2779, opened 2016-11-01
Update _.chunk to better leverage immutability and ES6 constructs

mfix22:
This version of chunk does not require the `guard` parameter and better utilizes ES6 constructs to chunk arrays. I think this style is easier to reason with (and looks prettier 👍). 
However this does require ES6, so that is why I bumped the node engine up to >=v6.0.0. 

If you want to make a new branch for rewriting many of the functions in ES6/7, I can begin updating other functions as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "license": "MIT",
   "private": true,
   "main": "lodash.js",
-  "engines": { "node": ">=4.0.0" },
+  "engines": { "node": ">=6.0.0" },
   "scripts": {
     "build": "npm run build:main && npm run build:fp",
     "build:fp": "node lib/fp/build-dist.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "license": "MIT",
   "private": true,
   "main": "lodash.js",
-  "engines": { "node": ">=4.0.0" },
+  "engines": { "node": ">=6.0.0" },
```
StreetStrider: :laughing:

jdalton:
Hi @mfix22!

Thanks for this but we'll keep the more broadly supported syntax for Lodash v4.
I'll revisit [for v5](https://github.com/lodash/lodash/wiki/Roadmap) when we go ES2015+ (w/ Babel).

mfix22:
Thanks @jdalton

PR #2793, opened 2016-11-06
Fix freeProcess.binding(util) usage

jhen0409:
Currently I'm using WebWorker on Electron, and it have `process` object on global (it just mock `process.env`, no `process.binding`), it will pause at useless places if we open `Pause On Caught Excpetions` on devtools:


jdalton:
@jhen0409 Thanks! It's edge case but I can dig it!

Issue #2801, opened 2016-11-10
_.defaultsDeep incorrectly works with array values.


seth2810:
When I apply `_.defaultsDeep` for objects which have properties represented as array I got strange behavior: 

```
_.defaultsDeep({config: ['a/b']}, {config: ['b/a', 'c/d']}) // {config: ["a/b", "c/d"]} 
```

I expect that `config` property value doesn't change because it already has value. Could somebody tell me why it's possible?

seth2810:
After some research I found this line https://github.com/lodash/lodash/blob/master/lodash.js#L6530 and as I know `_.isObject` returns `true` for arrays.

jdalton:
Moved to #2802.

PR #2802, opened 2016-11-10
Prevent _.defaultsDeep from merging array values.

seth2810:
https://github.com/lodash/lodash/issues/2801


jdalton:
Merging arrays is tricky. We settled on the established deep extend behavior of jQuery. To allow folks to customize we provided things like [_.mergeWith](https://lodash.com/docs/#mergeWith), which is what `_.defaultsDeep` uses. For your needs youmay be better off using `_.mergeWith`.

PR #2837, opened 2016-11-19
typescript typings for lodash-es

stephenlautier:
Currently in order to use TypeScript with `lodash-es` its a problem since they are not available.

This would allow the TS users to simply download this package + typings from original source `@types/lodash` and they get typings out of the box for lodash-es.


jdalton:
Thanks @stephenlautier!

But I'm not interested in maintaining typings for Lodash.

PR #2849, opened 2016-11-24
 make the documentation about _.isEqual clearer

### LINKED ISSUE #2848: Documentation: make the documentation about _.isEqual clearer
The documentation of `_.isEqual` [states that](https://github.com/lodash/lodash/blob/4.17.2/lodash.js#L11509-L11510):

> Functions and DOM nodes are **not** supported.

Which made me think that giving such objects to the function raises an error, that they don't behave in a coherent way when given to the function, or at least that comparing them is discouraged. Is that on purpose ?

If not, this sentence could be rewritten as:

> Functions and DOM nodes are simply compared with the `===` operator.

lovasoa:
Fixes #2848

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -11521,9 +11521,8 @@
      * **Note:** This method supports comparing arrays, array buffers, booleans,
      * date objects, error objects, maps, numbers, `Object` objects, regexes,
      * sets, strings, symbols, and typed arrays. `Object` objects are compared
-     * by their own, not inherited, enumerable properties. 
-     * Functions and DOM nodes are simply compared with the `===` operator.
-
+     * by their own, not inherited, enumerable properties. Functions and DOM
+     * nodes are simply compared with the `===` operator.
      *
      * @static
      * @memberOf _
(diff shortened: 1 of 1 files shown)


lodash.js
```diff
@@ -11521,8 +11521,9 @@
      * **Note:** This method supports comparing arrays, array buffers, booleans,
      * date objects, error objects, maps, numbers, `Object` objects, regexes,
      * sets, strings, symbols, and typed arrays. `Object` objects are compared
-     * by their own, not inherited, enumerable properties. Functions and DOM
-     * nodes are **not** supported.
+     * by their own, not inherited, enumerable properties. 
+     * Functions and DOM nodes are simply compared with the `===` operator.
+
```
jdalton: Rogue new line.

lodash.js
```diff
@@ -11521,8 +11521,9 @@
      * **Note:** This method supports comparing arrays, array buffers, booleans,
      * date objects, error objects, maps, numbers, `Object` objects, regexes,
      * sets, strings, symbols, and typed arrays. `Object` objects are compared
-     * by their own, not inherited, enumerable properties. Functions and DOM
-     * nodes are **not** supported.
+     * by their own, not inherited, enumerable properties. 
+     * Functions and DOM nodes are simply compared with the `===` operator.
```
jdalton: No need to break it up. Use an 80 char wide ruler to guide line breaks.

jdalton:
Thanks!

Issue #399, opened 2013-11-19
Deactivate ES6 delimiter for _template() ?


jdespatis:
I'm using `yeoman` for templating a `build.xml` for `ant`. This file already uses a bunch of variables which have all the format `${...}`

Therefore I need to template the code `<% ...%>` in my `build.xml` but I also need to leave untouched the `${...}` in it, to not interfere with ant code

I could add a bunch of symbols to espace all `${...}` but it would complex the code

The quick and clean solution would be just to tell (dynamically) lodash to not support ES5, is it possible?

Thanks for your feedback

rovens:
@jdalton You rock.

beders:
${} should have never been added in the first place. Please make optional.
This affects both yeoman as well as slush generators dealing with ES6 templates

jdalton:
@beders It's been that way for ~3 years and is configurable.

beders:
but not mentioned in the documentation of _.template or _.templateSettings after 3 years.

In _.templateSettings:
'By default, the template delimiters used by lodash are like those in embedded Ruby (ERB). '

There's an example in _.template:
'// Use the ES delimiter as an alternative to the default "interpolate" delimiter.'
which is ambiguous at best and a place I would have looked at.

It could save a lot of time if the actual templateSettings documentation would spell out the RegExp's used.

Even when looking at lodashs source it isn't immediately clear what is being matched.

Happy to open an issue for that.

jdalton:
I'm open to a doc note (nothing fancy) if you want to add that.

PR #2857, opened 2016-12-01
Mention ES6 delimiters in templateSettings docs

davewasmer:
After a fair amount of confusion and hunting, I finally realized (via #399) that the template function defaults to ERB _and_ ES6 template string delimiters by default. The `templateSettings` docs explicitly mention the ERB style delimiters, but the ES6 style is mentioned only in a code sample comment on the template function itself.


jdalton:
I can dig it.

PR #2862, opened 2016-12-03
Added new functionality: filterWhile

7coder7:
A generalized version of takeWhile and dropWhile

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,4 +1,4 @@
-# <a href="https://lodash.com/">lodash</a> <span>v4.17.2</span>
+# <a href="https://lodash.com/">lodash</a> <span>v4.17.3-pre</span>
 
 <!-- div class="toc-container" -->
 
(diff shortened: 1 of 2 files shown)


megawac:
I think this is better handled by composing the two methods, either through a `chain`, where shortcut fusion will apply or through the standard interface `_.filter(_.takeWhile(...), ...)`. Implementing methods like this increase the cognitive load of the API (whats different between `filterWhile` than the combination of take and filter?) and also is a slippery slope as similar methods will end up being proposed.

7coder7:
@megawac yes I thought about that approach. The concern of traversing twice is what led me to this approach. Never looked it in through the perspective of load time though

megawac:
I wasn't referring to actual script load/parse time, I meant that increasing the number of methods in a library increases how long it takes for someone to learn the API and the amount of concepts they need to keep track of

jdalton:
Hi @7coder7!

This looks like [_.takeWhile](https://lodash.com/docs#takeWhile).

PR #2884, opened 2016-12-12
Random unique function

blackhawk389:
Creates an array of random and non-duplicate elements
Requires size of the array
Returns array of random and non-duplicate element of size provided


jfmengels:
Hi @blackhawk389! Thanks for the PR, but we prefer to discuss new methods in issues before-hand. In the case of this method, you can already do it using [`_.uniqueId`](https://lodash.com/docs/4.17.2#uniqueId)
```js
_.times(N).map(_.uniqueId)
```

PR #2890, opened 2016-12-16
Update README.md

NTrukhanovets:
Please, update docs about guaranteed `_.forEach` iteration order.


jdalton:
Hi @NTrukhanovets!

Thanks for the PR. I think in this case preserved order is the norm so there's little benefit in spelling it out. It just adds noise.

PR #2891, opened 2016-12-17
perf(isObject): Avoid caching loose comparison to null

bigopon:
Hi, from this [benchmark](https://gist.run/?id=680c1d0d904fad05549ab93fa033897d)

Using strict null check and avoiding caching can help increase performance about 25 - 30%


jdalton:
Thanks @bigopon, but I'm not concerned with micro opts like this.

PR #2925, opened 2017-01-11
minor createFlow updates

bjarneo:
- Removed not used variable (wrapper)
- Removed func since it's not needed
- Strictly check if not is function 
- Add a couple of new lines to increase readability


jdalton:
Thank you @bjarneo!

PR #2932, opened 2017-01-13
Add related function links for each function

bertyhell:
Would this be helpful?
We can add more of these if this is approved.

This could really help people discover functions in Lodash, that they didn't know about before.

### DIFF
--- a/keyBy.js
+++ b/keyBy.js
@@ -12,7 +12,7 @@ import reduce from './reduce.js';
  * @param {Array|Object} collection The collection to iterate over.
  * @param {Function} iteratee The iteratee to transform keys.
  * @returns {Object} Returns the composed aggregate object.
- * @related [groupBy](#groupBy), [partition](#partition)
+ * @see [groupBy](#groupBy), [partition](#partition)
  * @example
  *
  * const array = [
--- a/keyBy.js
+++ b/keyBy.js
@@ -12,7 +12,7 @@ import reduce from './reduce.js';
  * @param {Array|Object} collection The collection to iterate over.
  * @param {Function} iteratee The iteratee to transform keys.
  * @returns {Object} Returns the composed aggregate object.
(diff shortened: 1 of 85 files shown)


jdalton:
Hi @bertyhell!

We use the `@see` tag. If you'd like to add more that would be great.

keyBy.js
```diff
@@ -12,6 +12,7 @@ import reduce from './reduce.js';
  * @param {Array|Object} collection The collection to iterate over.
  * @param {Function} iteratee The iteratee to transform keys.
  * @returns {Object} Returns the composed aggregate object.
+ * @see [groupBy](#groupBy), [partition](#partition)
  * @example
```
jdalton: It can just be `@see groupBy, partition` no need to add markdown links as that's something for our doc generation to tackle.

jdalton:
It looks like our CLA bot is down (may be recovering from the recent github outage) but this level of contribution should be fine without it.

bertyhell:
ok, i added a bunch more @see statements.

@jdalton I'm not sure how i can reference the reverse function in sortBy since there doesn't seem to be a sortBy file. (same with the omit function). Where do those live?

jdalton:
The `sortBy` method is being dropped for `orderBy`.

bertyhell:
OK, I added "reverse" to orderBy.js

Let me know if there are any mistakes/remarks

jdalton:
Wow, thanks!

bertyhell:
These don't show up on the documentation site yet:
https://lodash.com/docs/4.17.4#camelCase

Is that because we're still on 4.17.4 or are the @see tags not included into the documentation?

jdalton:
Hi @bertyhell!

> Is that because we're still on 4.17.4 or are the @see tags not included into the documentation?

Yes and yes. This was added after v4.17.4 and we currently and I don' think we currently [transform them into HTML](https://github.com/jdalton/docdown).

PR #2940, opened 2017-01-18
shorter, cleaner code

guypeer8:


jdalton:
Hi @guypeer8!

Thanks for this. I'd like to see more, but this specific change is not something I'm looking for. I'm still on the fence on using the ES5 built-ins since we're a lib and are already abstracting things away the loop is minimal and not too unsightly.

PR #2948, opened 2017-01-24
Fix package.json to include some missing fields and keywords.

kramsamujh:
Hi, I added a few fields that I noticed were missing from the package.json file. I've also added some keywords that I felt described the project for it to be picked up by npmjs.


jdalton:
Thanks @kramsamujh!

But our npm `package.json` is in our [npm branch](https://github.com/lodash/lodash/tree/npm).

PR #2971, opened 2017-02-03
bugfixed in zip & concat funcs

ecmadao:
Find two small bugs in zip and concat funcs.


jdalton:
Thanks @ecmadao!

I imagine more will shake out once I reattach the unit tests.

PR #2984, opened 2017-02-08
avoid switch if possible

aozgaa:
Simplifies a branching operation slightly.


jdalton:
Hi @aozgaa!

In this case changing it to `undefined` won't affect it.
I use an empty string simply for type consistency.

PR #2987, opened 2017-02-09
Use parentheses around arrow function argument having a body with cur…

falsyvalues:
It's a style matter, but reasonable to use parentheses when you use function body.
So use `var => fn(var)` or `(var) => {  fn(var) }`, what do you think?

### DIFF
--- a/cond.js
+++ b/cond.js
@@ -31,7 +31,7 @@ import arrayMap from './.internal/arrayMap.js'
 function cond(pairs) {
   const length = pairs == null ? 0 : pairs.length
 
-  pairs = !length ? [] : arrayMap(pairs, pair => {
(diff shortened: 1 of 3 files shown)


RusinovAnton:
FYI your both examples have different outcome
`var => fn(var)` - returns result of `fn(var)`
`(var) => { fn(var) }` - doesn't return anything, just calling `fn(var)`
These two are equal
`var => fn(var)` or `(var) => { return fn(var) }`

falsyvalues:
I know outcome, it was just example of coding style :+1:

jdalton:
Hi @falsyvalues!

I'm OK with keeping functions with single params without parens.

falsyvalues:
Sure thing :+1:

PR #2991, opened 2017-02-09
Template broken after remove semicolons action

falsyvalues:


jdalton:
Ah nice find. Thank you!

PR #2993, opened 2017-02-09
Complete `identity` removal module.

falsyvalues:


jdalton:
Thank you @falsyvalues!

StreetStrider:
Man, I already miss them `identity` & `constant`. 😞

jdalton:
❤️ arrow functions

billyjanitsch:
@jdalton curious if you've done any benchmarking around the cost of instantiating the identity function every time it's used, as opposed to passing a constant reference. Obviously the unit cost is low but it might be the most expensive thing some of these tiny utils are doing, and could add up when they're used as an iteratee. Or are you not really interested in this kind of perf consideration?

jdalton:
Hi @billyjanitsch!

Ya I thought of that but I'll wait for someone to bring it up in a real perf scenario before digging too deep into it.

PR #2996, opened 2017-02-11
Remove duplicate check symbol type

wikibootup:
[isSymbol.js#L20](https://github.com/lodash/lodash/blob/master/isSymbol.js#L20)

When `isKey` function checks types, `isSymbol` function checks type of value.


jdalton:
The `typeof` check is super light and avoids the heavier call. So I try to use that when possible. But you're right I have the type check in the `isSymbol` function too.

PR #3001, opened 2017-02-13
Fixin imports in invoke.

falsyvalues:


jdalton:
RE #3000: And there was! Thanks again!

PR #3002, opened 2017-02-13
Fixin imports in invert methods.

falsyvalues:


jdalton:
Thanks again! I have a WIP patch for `toString` I need to double check too!

PR #3008, opened 2017-02-15
Fixin `arrayPush` imports.

falsyvalues:
@jdalton 54abb96 removes `in` methods, what about corresponding internals like `getAllKeysIn`?


jdalton:
Nice catch!

> what about corresponding internals like `getAllKeysIn`?

I think one or two core `In` methods will remain. Basically if you can get the keys all other methods can be created from that.

falsyvalues:
That's reasonable for sure :+1:

PR #3013, opened 2017-02-15
Create filterObjects function.

passabilities:
Purpose of this function is to filter keys and values out of an object rather than the default `filter` function that only loops through the values.


jdalton:
Hi @passabilities!

I think you're looking for [_.pickBy](lodash.com/docs#pickBy).

passabilities:
:+1:  haha I knew there had to be a function that did this..
Thanks

PR #3024, opened 2017-02-22
Resolving a Promise without any argument breaks in some browsers

mateoperez:
### Background

Promise.resolve without any argument throws an error with

```
TypeError: Expected at least one argument
```

This happens in lg (smarttv) and webmaf browsers

### Proposed Solution

Resolve with ``undefined``so the check passes and still keep the previous behavior.


jdalton:
Hi @mateoperez!

Since it's a bug in a built-in I'll defer to built-in shims/polyfills to fix the issue.

PR #3029, opened 2017-02-25
Fixed small typo in debounce docs

MarkPollmann:
Just a small thing I noticed, I think it reads better this way.


jdalton:
@Mark Pollmann Thank you!

PR #3033, opened 2017-02-27
Create titleCase.js

cope:
Creating titleCase function based on this wikipedia article: https://en.wikipedia.org/wiki/Capitalization#Title_case


jdalton:
Thanks @cope!

I think this is basically [_.startCase](https://lodash.com/startCase).

PR #3034, opened 2017-02-27
Create sentenceCase.js

cope:
Creating sentenceCase function based on this wikipedia article: https://en.wikipedia.org/wiki/Capitalization#Sentence_case


jdalton:
This is a dup of [_.capitalize](https://lodash.com/docs#capitalize).

benheng:
How about a _.sentenceCase that splits the words like _.lowerCase, _.upperCase, or _.startCase do?

Issue #3051, opened 2017-03-10
Throttle invokes leadingEdge immediately after trailingEdge 


cpnota:
`throttle` with default options invokes the function more frequently than expected. It seems very unlikely that this is intended behavior, and it's inconsistent with underscore behavior.

Reproductions steps:

1. Call a throttled function twice
1. `leadingEdge` is invoked
1. Wait for the throttle period (`wait`)
1. `trailingEdge` is invoked
1. Call the throttled function again immediately afterwards
1. `leadingEdge` is invoked, even though `trailingEdge` was just invoked

The root cause is fairly obvious, and has nothing to do with setTimeout inconsistencies. `shouldInvoke` looks like:

```javascript
  function shouldInvoke(time) {
    var timeSinceLastCall = time - lastCallTime,
        timeSinceLastInvoke = time - lastInvokeTime;

    // Either this is the first call, activity has stopped and we're at the
    // trailing edge, the system time has gone backwards and we're treating
    // it as the trailing edge, or we've hit the `maxWait` limit.
    return (lastCallTime === undefined || (timeSinceLastCall >= wait) ||
      (timeSinceLastCall < 0) || (maxing && timeSinceLastInvoke >= maxWait));
  }
```

`timeSinceLastCall` is the time since `debounced` was last invoked. Therefore, in the above scenario, on step 5 `shouldInvoke` returns `true`, and `leadingEdge` is called.

Discovered in a unit test that looks something like:

```javascript
import { expect } from 'chai';
import sinon from 'sinon';
import throttle from 'lodash.throttle';

describe('lodash.throttle', () => {

	let clock;
	let invokeCount;
	let lastInvoked;

	const func = () => {
		invokeCount++;
		lastInvoked = Date.now();
	}

	const waitFor = ms => new Promise(resolve => {
		setTimeout(resolve, ms);
		clock.tick(ms);
	});


	beforeEach(() => {
		clock = sinon.useFakeTimers();
		invokeCount = 0;
		lastInvoked = undefined;
	});

	afterEach(() => clock.restore());

	it('invokes func at most once every 10 seconds', () => {
		const throttlePeriod = 10;
		const throttled = throttle(func, throttlePeriod);

		expect(invokeCount).to.equal(0);

		throttled();
		throttled();

		expect(invokeCount).to.equal(1);

		return waitFor(throttlePeriod).then(() => {
			expect(invokeCount).to.equal(2);

			throttled();

			expect(invokeCount,
[...]

PR #3053, opened 2017-03-11
Fix trailing edge issue in throttle

cpnota:
PR to address issue https://github.com/lodash/lodash/issues/3051

Note: Given the state of flux of the repo, I was unable to run any scripts or tests, except for some local tests. There are also not any tests here, since they don't seem to be in this repo? Please advise.

Eventually I determined that the issue boils down to a fundamental incompatibility between the expectations for `debounce` and the expectations for `throttle`. Following the trailing edge, `debounce` does not (and should not) care about the the amount of time elapsed before the next leading edge. `throttle` does. There seem to be three possible approaches to resolving this:

1. Decouple the `throttle` and `debounce` implementations again. No thanks.
1. Reclaim `maxWait` in the name of House Throttle. In the original debounce-based throttle implementation, setting `maxWait` reset the `lastCall` time to the time that the trailing edge created by `maxWait` was invoked (see [here](https://github.com/lodash/lodash/commit/1933a766310997518ab09f591cf364bd71d30607#diff-001d0647fb00f8336795faccdec19a31R4885). This is ostensibly not correct, which is probably why the behavior was changed at some point.
1. Add a new option that `throttle` can utilize. This is the approach I went with.

I added an option called `cooldown` which lets you specify the minimum amount of time that must elapse between a trailing edge and a leading edge. Calls to `debounced` that occur in that interval are deferred until the end of that cooldown period (effectively creating a new trailing edge).

### DIFF
--- a/debounce.js
+++ b/debounce.js
@@ -117,7 +117,7 @@ function debounce(func, wait, options) {
       return cooldownRemaining;
     }
 
-    return maxing ? nativeMin(remaining, maxWaitRemaining, cooldownRemaining) : remaining
+    return maxing ? nativeMin(remaining, maxWaitRemaining) : remaining
   }
 
   function shouldInvoke(time) {
--- a/debounce.js
+++ b/debounce.js
@@ -31,6 +31,9 @@ const nativeMin = Math.min
  * @param {Function} func The function to debounce.
  * @param {number} [wait=0] The number of milliseconds to delay.
  * @param {Object} [options={}] The options object.
+ * @param {number} [options.cooldown]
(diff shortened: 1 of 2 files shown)


cpnota:
btw I signed the CLA, but the check hasn't updated. @jsf-clabot ?

jdalton:
@cpnota The CLA should be signed with the name and email associated with your git config. You can update your git config info then `git commit --amend --reset-author` on your commits and force push.

jdalton:
Closing as abandoned.

PR #3084, opened 2017-03-29
use reduce to deal the loop thing

tangweikun:


jdalton:
Hi @tangweikun!

Thanks, but no thanks. I'm up for using some built-ins but it has to be in cases where a user couldn't provide a sparse array.

PR #3086, opened 2017-03-30
Remove constant function from examples

102:
`methodOf` example is pretty ugly now (`i => () => i` part), but I've never seen this method in real life and can't create a nicer example


jdalton:
Yiss! Thank you @102!

PR #3088, opened 2017-03-30
Fix #3087: Update _.pad* examples to show pad length < string.

### LINKED ISSUE #3087: Documentation on _.pad* should give hint about pad lengths shorter than original string.
Looking just at the documentation, it is unclear what this should return: `_.pad("abc",2)`.  Without a hint, each developer needs to run the experiment.

merriam:
One example now shows that specifying a padding length of less than the
length of the string returns the entire original string.

### DIFF
--- a/pad.js
+++ b/pad.js
@@ -19,7 +19,7 @@ import stringSize from './.internal/stringSize.js'
  * pad('abc', 8, '_-')
  * // => '_-abc_-_'
  *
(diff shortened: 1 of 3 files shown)


jdalton:
Thank you @merriam!

merriam:
Wow!  Thanks for the fast turn-around!  You are awesome!

PR #3090, opened 2017-03-31
Update get.js documentation including better examples

merriam:
Documentation now clearly shows this function also allows accessing a path that may not be valid.  The examples now show a more clearly invalid path and that an invalid path returns `undefined` if no default value is provided.

### DIFF
--- a/get.js
+++ b/get.js
@@ -1,8 +1,9 @@
 import baseGet from './.internal/baseGet.js'
 
 /**
- * Gets the value at `path` of `object` or `undefined` if the path cannot be evaluated.  If 
- * the resolved value is `undefined`, the `defaultValue` is returned in its place.
+ * Gets the value at `path` of `object` or gets `undefined` if the 
+ * path cannot be evaluated.  If the resolved value is `undefined`, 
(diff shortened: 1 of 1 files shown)


jdalton:
I dig the code example but not the explicit call out for `invalid`. It's covered in the mention of values that resolve to `undefined`. As a path that doesn't exist will resolve to `undefined`.

merriam:
I respectfully disagree that it covered, and I see that the wording could be better.

The use case of this function is to get a value instead of an exception when getting `myObject.a[0].b.c` and myObject is really undefined because some call upstream failed.  It begs the question to think that a path that cannot resolve without an exception is undefined.   

My guess is you object to the word 'invalid'.  How about being really explicit in the description?

``` Gets the value at `path` of `object`, or get `undefined` if the path cannot be evaluated.  If the resolved value is `undefined`, the `defaultValue` is returned in its place.```

This gets rid of comments on the default parameter, avoids 'invalid' and is explicit.  Added the commits.

jdalton:
I decided against this PR. Thanks though.

PR #3123, opened 2017-04-24
Fix cloneDeep with circularly dependent Sets/Maps

### LINKED ISSUE #3122: cloneDeep throws RangeError for Sets with circular references
If you have an Object containing a Set with a reference to the Object and you `_.cloneDeep` it, you get `RangeError: Maximum call stack size exceeded`. In the same situation, but with an Array instead of a Set, there's no problem. Here's how you can reproduce it:

```js
var objSet = { x: new Set() };
objSet.x.add(objSet);
_.cloneDeep(objSet);
// RangeError: Maximum call stack size exceeded

var objArr = { x: [] };
objArr.x.push(objArr);
_.cloneDeep(objArr);
// Properly deep-cloned object
```

Seems like circular dependencies aren't properly handled when they're in Sets? From #2137 I saw that support for deep-cloning Sets, Maps, and Symbols was added at 24cef1a. Is this behavior expected? If not, I'd be happy to attempt a fix. Seems like I would need to patch the `baseClone` function?

ajgreenb:
I've submitted a patch below. Let me know how it should be improved! I couldn't find the tests (probably because of the work on v5?), but I wrote a passing test on the 4.17.4 release. I'll paste it below.

Fixes #3122


ajgreenb:
I had this under the `'clone methods'` module.

```js
QUnit.test('`_.cloneDeep` should deep clone maps and sets with circular references', function(assert) {
  assert.expect(2);

  if (Set) {
    var object = { x: new Set };
    object.x.add(object);
    var clone = _.cloneDeep(object);
    assert.ok(lodashStable.isEqual(object, clone));
  } else {
    skipAssert(assert);
  }

  if (Map) {
    var object = { x: new Map };
    object.x.set('y', object);
    var clone = _.cloneDeep(object);
    assert.ok(lodashStable.isEqual(object, clone));
  } else {
    skipAssert(assert);
  }

});
```

jdalton:
Awesome!!!!

phoet:
i'm experiencing this issue as well with lodash 4.17.10 in our production setup:

```
  "name": "RangeError",
  "message": "Maximum call stack size exceeded",
```

unfortunately, even with sourcemaps i can't really pinpoint where this happens exactly and with what kind of data exactly. i suspect it's when `cloneDeep` is called with data that is fetched from a firebase database. since this is happening randomly, i don't have a good understanding of it. i guess i could try logging whatever is beeing deep-merged so that i might be able to see what's causing the issue.

jdalton:
Hi @phoet!

If you can manage a repro that would be great! This fix is in 4.17.10 so that means it could be elsewhere. We do recursively call `baseClone` though which means we could hit max call stacks. This is a known limitation but I also figured it would be a rare occurrence _(it has been)_. If firebase data is JSON-like you could probably just perform `JSON.stringify()` and `JSON.parse()`.

phoet:
@jdalton thanks for the hint and pointing to a workaournd. will see if i can find the data that is causing this.

PR #3125, opened 2017-04-25
Created callOnceInInterval.js

rgabs:
It calls the passed function once in passed interval.


jdalton:
Hi @rgabs!

We don't usually add features straight from a PR. See [our contributing text](https://github.com/lodash/lodash/blob/master/.github/CONTRIBUTING.md#feature-requests).

PR #3142, opened 2017-05-04
Reference reject in filter docs.

fodma1:
I was looking for reject functionality, but couldn't find it under filter docs.


jdalton:
Thanks @fodma1!

I'm not sure if I'm carrying `reject` over to v5 so will hold off on this for now.

Issue #2195, opened 2016-04-01
Equivalence Classes / Binning?


c-dante:
I didn't see a way for lodash to nicely partition a collection into bins. Sometimes, in JavaScript, you want to group items by some property that doesn't translate to a key very well. For example, grouping a collection into equivalence classes.

For my own code, I ended up doing something like this:
```js
var genericGroup = (keyFn, valueFn, eqFn, collection) => collection.reduce((acc, item) => {
	var key =  keyFn(item);
	var val = valueFn(item);
	var bin = _.find((x) => eqFn(x.key, key), acc);
	if (bin) {
		bin.items.push(item);
	} else {
		acc.push({ key: key, items: [val] });
	}
	return acc;
}, []).map((x) => x.items);
```

Is there interest for this kind of function in lodash? Any ideas of a nicer way to do this? The idea is that your keying function doesn't have to map to a string -- it can accept an equivalence on keys to bin values by. You could potentially output tuples of `[key, items[]][]`, or similar. I imagine this would be a `_.binBy`. 

I might be totally missing the forest from the trees here.

---
A use case: I have a collection of sets that I would like to say are equivalent if they contain the same "important" items, and collect together groups of these "important" sets.

```js
// Assume unnamed objects are some complex JSON value
var myCollection = [
	[ A, B, C ],
	[ D, C, E ],
	[ A, E, C]
];

var keyFn = _.filter((x) => x === A || x === C); // Important nodes are A, C
var valueFn = _.identity;
var eqFn = (left, right) => left.length === right.length && _.isEmpty(_.difference(left, right)); // Set equality
var interestingGroups = genericGroup(keyFn, valueFn, eqFn, myCollection);
/*
[
	[ [ A, B, C], [ A, E, C ] ],
	[ [ D, C, E ] ]
]
*/
```

jdalton:
Thanks @c-dante! We'll keep an eye on user demand.

StreetStrider:
Looks like a generalized version of `partition`. :)

c-dante:
@StreetStrider that's the idea :) take `true/false` and go to... `eqFn`?

StreetStrider:
@c-dante, ye, I think some `fn` which would return «key» to distinguish partitions is required here. It is kind a «resolver».

emilefokkema:
[...]

PR #3153, opened 2017-05-10
A 'families' function on Collections

emilefokkema:
This is related to https://github.com/lodash/lodash/issues/2195.

I have often needed to partition a collection of things based on whether two elements in the collection belong to the same group. Say, for instance, you have a collection of things that are linked to each other one-by-one, so A is connected to C, C is connected to E and B is connected to D. I would like a function that, given A, B, C, D and E, returns two groups: one containing A, C and E and one containing B and D.

In other words, it would be very cool if there were like a ‘families’ function accepting two parameters:
- a collection, and
- a predicate of two parameters which, if it returns true, means that those two elements belong to the same group.

This would be a generalization of 'partition', in other words: 'partition', when called with predicate 'p', would call 'families' with predicate (x,y) => p(x) == p(y).

I don't know if 'families' is a good name. I was inspired by Wittgenstein's notion of family resemblance: https://en.wikipedia.org/wiki/Family_resemblance

### DIFF
--- a/families.js
+++ b/families.js
@@ -1,0 +1,52 @@
+import pullAt from './pullAt.js'
+import flatMap from './flatMap.js'
+import findIndex from './findIndex.js'
+import some from './some.js'
+
+/**
+ * Creates an array of elements split into groups. For any elements x and y of the collection,
+ * if predicate(x,y) returns truthy, x and y will end up in the same group. It is assumed that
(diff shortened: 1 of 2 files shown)


jdalton:
Hi @emilefokkema!

The issue #2195 isn't open to PRs at this time.

PR #3166, opened 2017-05-22
Update compact.js, optimize

odykyi:


jdalton:
Hi @odykyi!

Thanks for taking the time to do a PR. In this case however, the existing code is just fine (minifies to hardly and thing) and is faster.

PR #3168, opened 2017-05-23
Updated size.js documentation

Juzzz:
Added documentation to the size function to show undefined and null results


jdalton:
Hi @Juzzz!

Thanks, but no thanks. The method is a "Collection" method so the examples are collections.

PR #3214, opened 2017-06-20
default prefix to empty string

jisaacks:
This starting returning NaN for me. It looks like it is trying to add undefined + number when there is no prefix.


jdalton:
Thanks @jisaacks!

PR #3224, opened 2017-06-27
convert mapKey.forEach to for loop

aretecode:


jdalton:
Thanks @aretecode!

I want to hold off on rolling things to loops for now. I was just stepping into using more es5 built-ins.

PR #3230, opened 2017-06-27
feat: Added logic not function

Qquanwei:
logic log

```
 not(true) // => false
 not(false) // => true
 not(truth) // => false
```

### DIFF
--- a/not.js
+++ b/not.js
@@ -1,0 +1,30 @@
+/*
+ *
+ * @param {*} value The value to not.
+ * @return {boolean} Return `false` if value is truth
+ * @example
+ *
+ * not(false)
+ * // => true
+ *
+ * not(true)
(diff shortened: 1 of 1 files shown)


jdalton:
Closed until there's a bit more demand.

Qquanwei:
@jdalton 

In following case , I want to use `not `

//require lodash/fp

filter prop not equals
```
 filter(flowRight(not, eq(18), prop('age') ), collections)
```
not identity
```
filter(flowRight(not, identity), collections)
```
i think this will be helpful for us 

do you think about this?

robludwig:
you can make your own logical not using `negate` and `identity`: 

```
const not = _.negate(_.identity);
not(true) //false
not(false) //true
not(!!3) //false
```

PR #3231, opened 2017-06-27
feat: added logic not

Qquanwei:
logic not

```
 not(true) // => false
 not(false) // => true
 not(truth) // => false
```

```
 flowRight(not, identity)(obj)
```


jdalton:
See #3230.

PR #3264, opened 2017-07-16
_.update() docs grammar fix

nullvariable:
"that accepts" -> "that it accepts"


jdalton:
Hi @nullvariable! It looks like an extra leading space was added.

jdalton:
Thanks!

PR #3300, opened 2017-08-04
Compare strictly pickBy param with null

fabienheureux:
Best practices thing 
Maybe there was a good reason to use `==` here, that I would be glad to hear 👍 (honestly 😉)


jdalton:
Hi @fabienheureux!

The loose `==` is used to catch `null` and `undefined`.

PR #3308, opened 2017-08-08
Replacing a regex with a quick character check

tlhunter:
* Checking the 0th position of the string should be quicker than a regular expression

### DIFF
--- a/.internal/stringToPath.js
+++ b/.internal/stringToPath.js
@@ -1,7 +1,6 @@
 import memoizeCapped from './memoizeCapped.js'
 
 const reEscapeChar = /\\(\\)?/g
-const reLeadingDot = /^\./
 const rePropName = RegExp(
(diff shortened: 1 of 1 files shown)


.internal/stringToPath.js
```diff
@@ -25,7 +24,7 @@ const rePropName = RegExp(
  */
 const stringToPath = memoizeCapped((string) => {
   const result = []
-  if (reLeadingDot.test(string)) {
+  if (string.charAt(0) === '.') {
     result.push('')
```
jdalton: Comparing against the char code would be faster.
Something lile
```js
const codeOfDot = '.'.charCodeAt(0)
```
Then later
```js
string.charCodeAt(0) === codeOfDot
```

jdalton:
Closing as inactive.

PR #3309, opened 2017-08-08
Clearing memoize cache if size >= 500

tlhunter:
* If we somehow skipped from 499 to 501 the cache would never clear


jdalton:
Hi @tlhunter!

Cache size isn't something that can be skipped. It's a product of how many entries are in the cache.

PR #3310, opened 2017-08-09
clarify documentation about sortedUniq

potmo:
clarify that sortedUniq only works on sorted arrays

### DIFF
--- a/sortedUniq.js
+++ b/sortedUniq.js
@@ -1,8 +1,10 @@
 import baseSortedUniq from './.internal/baseSortedUniq.js'
 
 /**
- * This method is like `uniq` except that it's designed and optimized
+ * This method is like `uniq` except that it only works
  * for sorted arrays.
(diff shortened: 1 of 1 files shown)


gnapse:
Just to be clear, technically `sortedUniq` work if the array is so that it guarantees that all items with the same value are consecutive, even if the array is not sorted. Is that correct?

In the case of `sortedUniqBy`, it works if the array is so that all items for which the iteratee function yields a certain value, appear consecutively in the array.

I don't mean to imply with this that the documentation should go to this level of details. It's just that I'm using `sortedUniqBy` in a situation where I've got an array of objects, all of which have `name` and `id`: I want to sort by name, and then remove duplicates by id.

```javascript
// using the fp versions cause the argument order is preferable
sortedUniqBy('id', sortBy('name', array))
```

After the array is sorted by name, all items with the same id are guaranteed to appear consecutively, but not necessarily in order by `id`.

PR #3313, opened 2017-08-09
add is defined() function

jbourjeli:
Many times I find myself writing statements like

```
if (!_.isUndefined(value)) {
  // do something
}
```
which can be confusing. It would be more readable to write something like
```
if (_.isDefined(value)) {
  // do something
}
```


jdalton:
Hi @jbourjeli!

We hold off on feature PRs until there's enough interest. Search the issue tracker and upvote it.

PR #3316, opened 2017-08-14
Correct license references to be CC0 instead of MIT

dvdckl:
lodash is licensed as CC0, but the package.json and README.md say MIT.  This changes the package.json SPDX license identifier and README.md to match the actual license.


jdalton:
Hi @dvdckl!

Lodash is MIT licensed.

dvdckl:
@jdalton my bad, I see now that the CC0 clause applies only to sample code, not the library itself.

PR #3338, opened 2017-08-26
Add error checking for iteratee

gabel0287:


jdalton:
Hi @gabel0287!

Thanks for the PR, but I'm OK with the existing behavior.

PR #3348, opened 2017-08-31
Fix kebabCase.js

yeroshkin:
Fix `kebabCase` to handle `№` symbol correctly

### DIFF
--- a/kebabCase.js
+++ b/kebabCase.js
@@ -21,7 +21,7 @@ import words from './words.js'
  * // => 'foo-bar'
  */
 const kebabCase = (string) => (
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @yeroshkin!

Thanks for creating a PR. String methods like this only support deburring [Latin-1 Supplement](https://en.wikipedia.org/wiki/Latin-1_Supplement_(Unicode_block)#Character_table) and [Latin Extended-A](https://en.wikipedia.org/wiki/Latin_Extended-A) letters. I'd like to avoid a bunch of one off conversions.

PR #3351, opened 2017-08-31
feat(defaultTo): allow multiple default values

trs:
This allows `defaultTo` to accept multiple arguments `defaultTo(value, default1, default2, ..., defaultN)` returning the first value to pass the existing `defaultTo` check.

This should remain a non-breaking change and should be backward compatible with existing functionality.

### DIFF
--- a/defaultTo.js
+++ b/defaultTo.js
@@ -1,12 +1,12 @@
 /**
  * Checks `value` to determine whether a default value should be returned in
- * its place. The `defaultValue` is returned if `value` is `NaN`, `null`,
- * or `undefined`.
+ * its place. The first `defaultValues` argument that is not `NaN`, `null`,
+ * or `undefined` is returned.
  *
  * @since 4.14.0
  * @category Util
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @trs!

This looks like something that would be a better fit composed with Array `reduce`.

trs:
@jdalton The problem with `reduce` is that it can't be broken out of early, so if you have 10 arguments and the 2nd is valid, it will still have to loop through the other 8.

crabl:
@trs why not just use `_.find` with `_.defaultTo`?

```
function defaultToMultiple(value, ...defaultValues) {
  return _.find(defaultValues, v => _.defaultTo(value, v));
}
```

trs:
@crabl It wasn't so much an issue with how to do this, it had been determined we can use `_.find` with `_.defaultTo` previously.

This PR was more to do with not having to have helper functions on top of lodash's helper functions. QOL improvement.

Issue #2758, opened 2016-10-26
Make _.xor() work with n arrays


aronwoost:
```js
// expected
_.xor([1, 2], [1, 2]);
// []

// actual
_.xor([1, 2], [1, 2]);
// []

// expected
_.xor([1, 2], [1, 2], [1, 2]);
// []

// actual
_.xor([1, 2], [1, 2], [1, 2]);
// [1, 2]
```
`_.difference()` I can not use because I need to compare every-with-every array. Any idea?

jdalton:
Nice find @aronwoost!

PR #3355, opened 2017-09-03
Fix baseXor implementation

stefanmaric:
baseXor and its dependents were changed on commit
729d1a57aa77a9c82c8292dcfa0054e4b90060b1, based on issue #2758.

But the previous implementation was correct, which respected the
properties of the algebra.

The symmetric difference of two sets is the union of both of their
relative complements, and it is commutative:

`A Δ B = (A ∖ B) ∪ (B ∖ A)`

`A Δ B = B Δ A`

In that respect, both implementations were equivalents, the problem
appeared when you tired to xor multiple sets, since the new
implementation didn't respect the associative property:

`A Δ (B Δ C) = (A Δ B) Δ C`

```javascript
const a = [1, 2, 4, 5]
const b = [2, 3, 5, 6]
const c = [4, 5, 6, 7]

_.xor( _.xor(a, b), c )
// [1, 3, 5, 7] (correct)

_.xor( a, _.xor(b, c) )
// [1, 3, 5, 7] (correct)

_.xor( a, b, c )
// [1, 3, 7] (wrong)
```

With this fix, the associativity is respected:

```javascript
_.xor( a, b, c )
// [1, 3, 5, 7] (correct)

```

See:

https://en.wikipedia.org/wiki/Symmetric_difference
https://en.wikipedia.org/wiki/Exclusive_or
https://math.stackexchange.com/questions/84184/relation-between-xor-and-symmetric-difference

### DIFF
--- a/.internal/baseXor.js
+++ b/.internal/baseXor.js
@@ -18,19 +18,15 @@ function baseXor(arrays, iteratee, comparator) {
     return length ? baseUniq(arrays[0]) : []
   }
   let index = -1
-  const result = new Array(length)
+  let result = []
 
   while (++index < length) {
-    const array = arrays[index]
-    let othIndex = -1
-
-    while (++othIndex < length) {
-      if (othIndex != index) {
-        result[index] = baseDifference(result[index] || array, arrays[othIndex], iteratee, comparator)
-      }
-    }
+    result = baseFlatten([
+      baseDifference(result, arrays[index], iteratee, comparator),
+      baseDifference(arrays[index], result, iteratee, comparator)
+    ], 1)
   }
-  return baseUniq(baseFlatten(result, 1), iteratee, comparator)
+  return baseUniq(result, iteratee, comparator)
 }
 
 export default baseXor


stefanmaric:
This issue affects versions of lodash:

* `4.16.5`
* `4.16.6`
* `4.17.0`
* `4.17.1`
* `4.17.2`
* `4.17.3`
* `4.17.4`

It would be worth fixing it for the the v4 series, but I don't see any `v4` branch.

jdalton:
Thanks @stefanmaric!

I'm OK with our version of `xor`. I think it makes more sense for common usage and can't see the example case, `_.xor( a, b, c )`, with `5` in the result as being what the user intended.

stefanmaric:
@jdalton I understand the point; finding the mutually exclusive elements of a collection of sets seems a more recurrent scenario than finding their actual symmetric difference.

But in the beginning `xor` was _mathematically correct_ and only after multiple years one person complained about it – it took way less time for someone (myself) to complain about the opposite.

Wouldn't it better to use the right semantics and provide the user with the tools for them to achieve what they want? It may be a new utility or we might educate them as well, via the docs. What `xor` currently does can be achieved with:

```javascript
const exclusion = (...arrays) =>
  _.flatMap(arrays, x => _.difference(x, ..._.without(arrays, x)))
```

Just saying that it may be worth it to stand on the shoulders of mathematical foundations and that it shouldn't be hard to workaround users' expectations.

PR #3365, opened 2017-09-11
Fixed a little typo

InsaneNaman:
If i am not wrong it should be 
`$ npm i -g lodash` instead of `$ npm i -g npm`

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ In a browser:
 
 Using npm:
 ```shell
-$ npm i -g npm
+$ npm i -g lodash
 $ npm i --save lodash
 ```
 


jdalton:
Hi @InsaneNaman!

Thanks for taking the time to create a PR. In this case the existing documentation is by design. I included it so that folks would get into the habit of updating their npm version *(this was before `npm` prompted folks to update on its own)*.

InsaneNaman:
Ohh i get it :sweat_smile:. :+1: I was in dilemma that whether that command is for global install or npm update. Thanks and sorry for wrong PR. @jdalton

PR #3388, opened 2017-09-22
Add pending() function to debounce and throttle to fix #3387

### LINKED ISSUE #3387: Add `pending` property to throttled functions
Is there some way of knowing if there are pending invocations of throttled functions? Take this example of a throttled function that sends data in the browser to be saved on the server:

```javascript
const throttledSave = _.throttle(() => {
  server.save(getLatestLocalData());
}, 2000, {
  leading: false,
  trailing: true,
});
```

Because this throttled function can be called many times a second, elsewhere I'd like to be able to track whether there are pending invocations of the throttled function so that I can do useful things like inform users about the status of syncing with the server. If there were a boolean or integer `pending` property on throttled functions, I could do:

```javascript
displayedStatus = throttledSave.pending ? "Saving…" : "Saved";
```

Thoughts? Is there some other workaround I should consider?

chriscalo:
See issue #3387.

### DIFF
--- a/debounce.js
+++ b/debounce.js
@@ -51,6 +51,9 @@ import isObject from './isObject.js'
  *
  * // Cancel the trailing debounced invocation.
  * jQuery(window).on('popstate', debounced.cancel)
+ * 
+ * // Check whether there are currently pending invocations.
+ * debounced.pending()
  */
 function debounce(func, wait, options) {
   let lastArgs,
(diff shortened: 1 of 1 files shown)


jdalton:
Thank you @chriscalo!

chriscalo:
Thank you, @jdalton! 🙏

PR #3417, opened 2017-10-09
function checking.

karanjariwala:
Even if func is object it will pass the if condition. So i guess we need to explicitly check whether func is present and  it is a function.


jdalton:
Hi @karanjariwala!

With v5 the repo is in a bit of a scattered state but usually we'll defer the function check to the helper. In this case, `createWrap`. Lodash tries to do function type checks early to avoid a deferred error on invocation of the created function.

Issue #1436, opened 2015-09-03
Feature request: repeat arrays.


AlexFrazer:
Syntax in python will allow something like as follows:

```
>>> [1,2,3] * 4
[1,2,3,1,2,3,1,2,3,1,2,3]
```

Would this be something good for Lodash?

jdalton:
Maybe. I use a non-methodized form of this in our unit tests with `_.times`. While `_.times` is not as streamlined it can accomplish what you want.

I'll keep an eye on this to see if there's enough demand to warrant addition to core.

mk-pmb:
I was surprised to learn that lodash cannot yet repeat arrays. If `.times` can shim it, docs for `.repeat` should mention it, and should pre-emptively explain why not arrays.

I had to split out the relevant code in order to [fix it](https://github.com/mk-pmb/lodash-repeat-arrays-fix1436-pmb-js/commits/master) because I can't easily work on files that huge.

As I've gone each step (add array support, add buffer support) in an extra commit you can easily see the impact on assumed cost of operations:
* Using `.concat` to append strings or arrays costs that dot lookup + invocation in each step, and
* supporting the Buffer helper function costs an extra function call in each step. We could however save the `.concat` lookup if we made a special add function for the string case.

Hope it helps, and remember to add the docs spoiler! :-)

jdalton:
Hi @mk-pmb!

> If .times can shim it, docs for .repeat should mention it, and should pre-emptively explain why not arrays.

The docs have [_.repeat](https://lodash.com/docs#repeat) under the "Strings" category with documentation examples with strings. I'm not willing to go further than that in explaining the method as it would contribute to doc clutter.

mk-pmb:
> The docs have _.repeat under the "Strings" category

Oh ok, I missed that part because I followed the docs link from the lodash.repeat npm package. At least it's just a few clicks from "what, lodash can't do it?" to this feature request thread. :-)

jdalton:
I'll look into scrolling the toc entry into view as well when navigating to specific methods.

mk-pmb:
Wouldn't have helped in my case, but good idea anyways. (1. I hadn't granted JS execution for just reading docs, 2. I hadn't even looked at the toc because I felt entitled to expect based on the function name that of course it's not limited to just one data type.)

jdalton:
You should really enable JS on a JS library site. You're missing out on things like, doc repls, offline support, and enhanced search.

homerjam:
[...]

PR #3436, opened 2017-10-19
Add cycle function.

RikJurriaans:
In most FP languages you have some way to cycle an array. Lodash unfortunately had no easy way of doing this, except abusing the existing function times and merge.

### DIFF
--- a/cycle.js
+++ b/cycle.js
@@ -1,0 +1,32 @@
+/** Used as references for various `Number` constants. */
+const MAX_SAFE_INTEGER = 9007199254740991
+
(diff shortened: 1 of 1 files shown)


jdalton:
Dup of #1436.

tatethurston:
Cycle https://docs.python.org/2/library/itertools.html#itertools.cycle is slightly different than what #1436 is requesting. With cycle, an infinite data structure is returned, while #1436 is requesting to duplicate the elements of an array a fixed number of times.

PR #3445, opened 2017-10-23
fix(Node): ignore log files that end with random numbers

matthewdordal:
This is a pretty minor update but I sometimes see npm and yarn log files that are generated with the format `npm-debug.log.<randomNumbers>`. I'm not actually sure what causes this but it seems to happen to me when I'm in a location where my wi-fi reception is spotty.

This would prevent a file such as `npm-debug.log.123456789` from being added to the commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 .DS_Store
-*.log
+*.log*
 doc/*.html
 node_modules


jdalton:
Thanks @matthewdordal!

PR #3448, opened 2017-10-25
Fix typo in `escapeRegExp`

younesfkihi:


jdalton:
Thank you @younesfkihi !

falsyvalues:
@jdalton I don't think this is correct `\\` is an escape for `\` character and we match against `\` not `\\`.

jdalton:
@falsyvalues The typo is because you can't write `"\"` in JS because it's escaping the `"` so to make it more JS-y they wanted it written as `"\\"`.

falsyvalues:
@jdalton Still if you think about how regex work (and it is about regex) it's misleading and it's a comment which will be published on docs site. Anyway its up to you, just my few :dollar: :smile:

younesfkihi:
@falsyvalues @jdalton I added the fix because the [docs site](https://lodash.com/docs/4.17.4#escapeRegExp) shows `"$", "", "."` instead of `"$", "\", "."`. How can  `"\"` be shown on the site without escaping `\`?

jdalton:
@falsyvalues  Ah okay, I see your point. If you'd like to revert that's cool.

PR #3465, opened 2017-11-01
Workaround memory leak issue caused by slice method

mykola-mokhnach:
Recently we got several complains from our users about process crash due to OutOfMemory error. We've figured out, that the leak happens when one truncates a long string and passes it to the logger. More details here https://github.com/appium/appium/issues/9543

And [this Chromium bug report](https://bugs.chromium.org/p/v8/issues/detail?id=2869) contains more detailed description of what is happening under the hood. 

So, since we use `_.truncate` method to perform strings truncation, it might be a good idea to incorporate the fix/workaround here as well, so other module users don't suffer from this problem anymore.

### DIFF
--- a/truncate.js
+++ b/truncate.js
@@ -74,7 +74,7 @@ function truncate(string, options) {
   }
   let result = strSymbols
     ? castSlice(strSymbols, 0, end).join('')
-    : (' ' + string.slice(0, end)).substr(1) // https://bugs.chromium.org/p/v8/issues/detail?id=2869
+    : ` ${string.slice(0, end)}`.substr(1) // https://bugs.chromium.org/p/v8/issues/detail?id=2869
 
   if (separator === undefined) {
     return result + omission
--- a/truncate.js
+++ b/truncate.js
@@ -74,7 +74,7 @@ function truncate(string, options) {
   }
   let result = strSymbols
     ? castSlice(strSymbols, 0, end).join('')
-    : string.slice(0, end)
+    : (' ' + string.slice(0, end)).substr(1) // https://bugs.chromium.org/p/v8/issues/detail?id=2869
 
   if (separator === undefined) {
     return result + omission
--- a/truncate.js
+++ b/truncate.js
@@ -72,9 +72,14 @@ function truncate(string, options) {
   if (end < 1) {
     return omission
   }
-  let result = strSymbols
-    ? castSlice(strSymbols, 0, end).join('')
-    : ` ${string.slice(0, end)}`.substr(1) // https://bugs.chromium.org/p/v8/issues/detail?id=2869
+  let result
+  if (strSymbols) {
+    result = castSlice(strSymbols, 0, end).join('')
+  } else {
+    result = string.slice(0, end)
+    // Try to force a flatten: https://github.com/davidmarkclements/flatstr
+    Number(result)
+  }
 
   if (separator === undefined) {
     return result + omission


jdalton:
Thanks @mykola-mokhnach!

Super gross workaround though.
@bmeurer have you [seen this](https://bugs.chromium.org/p/v8/issues/detail?id=2869)?

jdalton:
I'd like to avoid going down this far of a hack if possible. Ideally, it'd be great if v8 fixed the bug.

mykola-mokhnach:
The bug has been reported 4 years ago. I have doubts something is gonna change there soon :(

jdalton:
If a fix isn't doable, @bmeurer could you suggest an easier to consume hack?

bmeurer:
We're aware of the problem. I'll bring that up again. I don't like the idea of spreading this *work-around* throughout the ecosystem, and I'm not sure the work-around buys you that much since `substr(1)` might as well create a `SliceString` again depending on the size of the slice result, plus it adds a performance cost to `_.truncate`.

However even if we address this problem in Chrome, you'll still run into with other's like Safari.

cc @hashseed

bmeurer:
Created a new tracking bug [v8:7025](crbug.com/v8/7025).

hashseed:
I had a fix 6 years ago, which flattens sliced strings on GC. However it was ultimately abandoned because it was not guaranteed to save memory. I'll propose something else.

hashseed:
What I've seen in V8's tests to create flat strings is to run a regexp on it, e.g. `/./.test(string)`. That's probably more performant than the proposed change here.

mykola-mokhnach:
@hashseed Cool, can you please link an example, so I could fix the PR?

bmeurer:
@mcollina suggested to use the `Number` constructor to force flatten a string, i.e. just sprinkle `Number(string)` and pray that TurboFan won't optimize it away :-)

hashseed:
There you go: https://github.com/v8/v8/blob/master/test/mjsunit/string-slices.js#L73

mcollina:
@davidmarkclements wrote this long explanation of the issue: https://github.com/davidmarkclements/flatstr.

We benchmarked the numbers and in our case passing it through `Number(str)` was slightly faster than passing it through a regexp. However this was a long time ago, so it's probably worth checking again.

mykola-mokhnach:
Does it look better now?

mcollina:
@mykola-mokhnach does it solve the problem? You should check on the supported Node release lines, because this is an implementation-level detail.

truncate.js
```diff
@@ -72,9 +72,14 @@ function truncate(string, options) {
   if (end < 1) {
     return omission
```
buoyantair: Can you tell me what this omission variable does

mykola-mokhnach: it's just period character ("..."), which is added at the end of the resulting string after/if it is truncated

buoyantair: Oh okay, thanks

mykola-mokhnach:
Many thanks for your support, although I believe you guys are much better in fixing/workarounding such stuff, than me :)
For now we are happy with our local patch, which is confirmed by users.

mcollina:
@mykola-mokhnach what is your local patch? can you add a link?

mykola-mokhnach:
It's still the same ugly stuff %)

https://github.com/appium/appium-support/pull/56/files

PR #3470, opened 2017-11-03
Update isPlainObject.js

robbix1206:
Simplify isPlainObject by checking has written in function header.
> Checks if `value` is a plain object, that is, an object created by the `Object` constructor or one with a `[[Prototype]]` of `null`.

### DIFF
--- a/isPlainObject.js
+++ b/isPlainObject.js
@@ -37,16 +37,13 @@ const objectCtorString = funcToString.call(Object)
  * // => true
  */
 function isPlainObject(value) {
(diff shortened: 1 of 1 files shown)


robbix1206:
Currently this PR is not es5 compliant
I can make it compliant by simply adding `typeof value === ''object"` if needed

jdalton:
Hi @robbix1206!

Thanks for making a PR. Our approach needs to be a bit more robust to work with objects across frames/realms.

PR #3479, opened 2017-11-07
Tiny performance improvement by not compiling regular expression each time

nazar-pc:
New version is constantly faster both in Firefox Nightly and Chromium Nightly: https://jsfiddle.net/dzym8svy/

Some projects call this function very often and performance hit can be significant in that case.

### DIFF
--- a/lowerCase.js
+++ b/lowerCase.js
@@ -19,8 +19,9 @@ import words from './words.js'
  * lowerCase('__FOO_BAR__')
  * // => 'foo bar'
  */
+const reQuotes = /['\u2019]/g
 const lowerCase = (string) => (
-  words(`${string}`.replace(/['\u2019]/g, '')).reduce((result, word, index) => (
+  words(`${string}`.replace(reQuotes, '')).reduce((result, word, index) => (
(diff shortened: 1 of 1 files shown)


lowerCase.js
```diff
@@ -19,8 +19,9 @@ import words from './words.js'
  * lowerCase('__FOO_BAR__')
  * // => 'foo bar'
  */
+const rx = /['\u2019]/g
```
buoyantair: @nazar-pc Just curious, how does this effect the performance? 

Im assuming that defining the regex before using it will eliminate the need for js to create a reference to it dynamically...

nazar-pc: Regular expressions take time to compile and additional time to execute upon some data. If you call lowerCase in a loop, regular expression will be compiler again and again on each call.

With proposed change it will be compiler once, thus each iteration will only involve executing pre-compiled regular expression, which is much faster.

jdalton: Can you rename the var to `reQuotes`.

nazar-pc: Done (commit amended)

buoyantair: Ohhh cool thank you!

Issue #2705, opened 2016-10-05
Symbol.toStringTag may fool isPlainObject()


trevnorris:
If a normally plain object with `Symbol.toStringTag` set is considered a plain object, it's currently fooled by `isPlainObject()`. Example:
```js
var o = {};
o[Symbol.toStringTag] = 'Foo';
Object.prototype.toString.call(o) === '[object Foo]';
```

jdalton:
Hey @trevnorris!

We usually don't try to get around folks using `Symbol.toStringTag`.
For example folks can do:
```js
var o = { [Symbol.toStringTag]: 'Date' }
_.isDate(o);
// => true
```

Is there a use case were this is causing problems?

jdalton:
I'm leaning towards being OK with objects with custom `Symbol.toStringTag` being treated as non-plain. For example `arugments` objects, error objects, etc are not considered plain objects.

jdalton:
@trevnorris ping
> Is there a use case where this is causing problems?

trevnorris:
@jdalton Cool. That was the same approach I took. Thanks.

jdalton:
I later moved it's own helper `getRawTag` since I have to wrap it in a try-catch. Built-ins like typed arrays, data views, and others have read-only `Symbol.toStringTag` values which will throw in strict mode.

not-an-aardvark:
For what it's worth, this can still be fooled by adding a getter for `Symbol.toStringTag`:

```js
// (running latest master, e7ddcad)

_.isPlainObject({get [Symbol.toStringTag]() { return 'Foo'; }})
// => false
```

I don't think it's possible to account for all cases like this (e.g. a Proxy could spoof all of its observable properties to look like a non-plain object). But I think it's worth noting that cases like this exist.

jdalton:
@not-an-aardvark Yap, though as long as it's configurable I could muck with it I guess.

jdalton:
I don't think we're trying to be super robust here. I just want to make sure there's communication in case there is a Node issue since @trevnorris is reporting.

jdalton:
I have a sneaky suspicion there's some investigation or discussion going on related to this that I'm not aware of.

not-an-aardvark:
No, this isn't related to a Node issue as far as I know -- I just happened to be looking at the Lodash issue list.

jdalton:
Ok 😎 !

jdalton:
[...]

PR #3483, opened 2017-11-08
Faster/simpler isPlainObject check

timdorr:
This runs up the prototype chain to check equivalency. When run in a cross-realm environment (differing contexts, iframes, etc), this ensures we're checking the value's prototype matches its context-specific instance of Object. 

This is faster than calling `toString()` on the constructor. There's still the `baseGetTag()` call, which does its own `toString()`, but this swaps out one for cross-realm stuff. It's also a bit simpler to understand, I think.

### DIFF
--- a/isPlainObject.js
+++ b/isPlainObject.js
@@ -28,7 +28,7 @@ import isObjectLike from './isObjectLike.js'
  * // => true
  */
 function isPlainObject(value) {
-  if (!isObjectLike(value) || baseGetTag(value) != objectTag) {
+  if (!isObjectLike(value) || baseGetTag(value) != '[object Object]') {
     return false
   }
   if (Object.getPrototypeOf(value) === null) {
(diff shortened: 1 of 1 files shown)


timdorr:
Do I need to do anything to get the tests to run? I checked out the `3203` branch and ran them there successfully.

jdalton:
> Do I need to do anything to get the tests to run? I checked out the 3203 branch and ran them there successfully.

Awesome!

That's fine for now. I think we can even remove some of the `baseTag` checks too.

timdorr:
>  I think we can even remove some of the baseTag checks too.

I had some difficulty there. These two tests fail in that case:
```
should return `false` for non-Object objects
should return `false` for objects with a read-only `Symbol.toStringTag` property
```
There's some edge case handling in #2705, but that's a good issue to read through for a reminder of why this check exists. 

It does feel a little weird to be looking to toString() to answer the question. I'm not sure if there's a workaround, but with all these new reflective APIs to override the toString() output, it's definitely a complex problem.

jdalton:
Cool. I'll re-evaluate those checks for v5.

PR #3487, opened 2017-11-10
Remove `map` example from `parseInt` and `trim`

younesfkihi:
The examples where `parseInt` and `trim` are used with `map` are not valid anymore because the `guard` param was removed:
``` javascript
map(['6', '08', '10'], parseInt)
 // => [6, NaN, 2]
map(['  foo  ', '  bar  '], trim)
 // => ['  foo  ', '  bar  ']
```


jdalton:
Thank you @younesfkihi! Nice catch!

PR #3499, opened 2017-11-17
Fix operator

kybetter:
Fix inequality operator and equality operator to strict mode to avoid type convert.


jdalton:
Please remove the extra `=` from the string comparisons. The `typeof` operation always returns strings there is no coercion there.

kybetter:
Thanks to your directions.

PR #3509, opened 2017-11-21
make returning copy not over-lengthing the source

wangsongiam:
I might be wrong but it's more intuitive to me to return a copy of the source when the destination array out-length the source.

//    copyArray([]source, <[]destination>) -> []copy
original: copyArray([1],[1,1]) // -> [1,1]
now     : copyArray([1],[1,1]) // -> [1]


jdalton:
Hi @wangsongiam!

This is an internal module not meant for general consumption.

PR #3525, opened 2017-11-29
Adds _.sign

viveleroi:
Adds `_.sign`, which returns a number representing the sign of a number. This method matches the functionality of [`Math.sign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/sign) which is not available in IE.

Due to the changing repo for v5 I could not see a way to add unit tests but I have tested this logic: https://jsfiddle.net/a8dkzjfx/2/

### DIFF
--- a/sign.js
+++ b/sign.js
@@ -1,0 +1,39 @@
+import isNumber from './isNumber';
+
+/**
+ * Returns a number representing the sign of `value`.
+ *
+ * If `value` is a positive number, negative number, positive zero or negative zero,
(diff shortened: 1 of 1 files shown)


jdalton:
Thanks @viveleroi!

cope:
Just an _anal_ suggestion:
```
function sign(v) {
	return !isNumber(v) ? NaN : (v > 0 ? 1 : (v < 0 ? -1 : 0));
}
```

PR #3531, opened 2017-12-05
add getType func

QingNeng:
add getType func


jdalton:
Hi @QingNeng!

Thanks for the PR. I want to hold off on this feature for now. You can totally make it a mixin though.

PR #3552, opened 2017-12-18
Add Map support in _.get

eatonphil:
I would like to be able to use _.get to abstract getting attributes of objects and Map objects.


jdalton:
Hi @eatonphil,

I want to avoid toggling for maps or sets here since they could have their own category of methods.

eatonphil:
Do objects vs arrays not have their own categories of methods? That is, is an array closer to an object than a Map is? I want my API to allow Maps or objects because as a consumer looking at a concrete key, they are identical (except for syntax).

Issue #3560, opened 2017-12-22
Feature Request: _.throttle using requestAnimationFrame via options flag


mattlubner:
Hi there!

Similar to #1165 (hat tip to @MeoMix), but instead of a whole new specialized function, what if we just add support to `_.throttle` directly, behind an options flag (eg. `options.requestAnimationFrame=true` perhaps?).

`_.throttle` could then degrade gracefully to its default implementation (with a default 1000/60 wait interval) whenever `window.requestAnimationFrame` is unavailable (such as in NodeJS).

Anyone else in favor of this? I, for one, would love to have this sort of utility within Lodash itself! 😀

jacob-beltran:
I have submitted a pull request to add this feature (#3562). Comments and feedback much appreciated.

jdalton:
Moved to #3562.

PR #3562, opened 2017-12-23
rAF support for debounce/throtte (#3560)

jacob-beltran:
### Adds an additional option to debounce and throttle
if the option `useRAF` is `true` and `requestAnimationFrame()` is available on the `root` object (e.g. `window` in a browser environment), it will be used for timing instead of `setTimeout()`.

This allows for very smooth and performant throttling of rapid firing events like scroll, resize, etc.

Feedback, comments, etc are much appreciated :)

### DIFF
--- a/debounce.js
+++ b/debounce.js
@@ -1,4 +1,5 @@
 import isObject from './isObject.js'
+import root from './.internal/root.js'
 
 /**
  * Creates a debounced function that delays invoking `func` until after `wait`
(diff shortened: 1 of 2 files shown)


jdalton:
Could this be done when nothing is set or when the timeout is set to a `Math.floor` of `16` ms? That way no option needs to be provided.

mattlubner:
That's an interesting suggestion. It would change the current default behavior (when no options are passed) from "deferring execution to the next tick" to "deferring execution to the next frame" in browser environments. I'd prefer this.

To play devil's advocate, should we expose an option to opt-out of RAF, to get the existing behavior?

mattlubner:
Another consideration regarding using RAF by default:

> The number of callbacks is usually 60 times per second, but will generally match the display refresh rate in most web browsers as per W3C recommendation. **`requestAnimationFrame()` calls are paused in most browsers when running in background tabs or hidden `<iframe>`s in order to improve performance and battery life.**

jdalton:
I'm fine with that as a default (non-specified times) then. If folks specify an ms then they'll get the old way. The next release will be v5 so any back-compat breaks are OK.

jdalton:
Closing due to inactivity.

mattlubner:
Apologies for the delay – I've re-opened this as #3567 and reworked the option so it's instead inferred from `wait` being omitted.

PR #3564, opened 2017-12-25
fix a test for the set methods

ersimont:
I believe this was intended. A negative number is used here in all other "complex paths" tests, and it wouldn't make sense to create keys past this one if it doesn't match the paths used.


jdalton:
Closed due to lack of activity.

PR #3567, opened 2017-12-30
rAF as default for debounce/throtte (#3560)

mattlubner:
Uses `requestAnimationFrame` by default (when available on the `root` object), unless `wait` is passed as `0` or a truthy value.

This is in reply to #3562, and in turn, #3560.

Feedback is welcome and appreciated. 😀

### DIFF
--- a/debounce.js
+++ b/debounce.js
@@ -4,22 +4,24 @@ import root from './.internal/root.js'
 /**
  * Creates a debounced function that delays invoking `func` until after `wait`
  * milliseconds have elapsed since the last time the debounced function was
- * invoked. The debounced function comes with a `cancel` method to cancel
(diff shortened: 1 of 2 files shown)


jacob-beltran:
Thanks for picking the up @mattlubner 
My apologies for not getting back to this. It's a busy time of year.

jdalton:
Thank you @jacob-beltran and @mattlubner!

joshbuckley182:
This doesn't seem to be available in any release at the moment (checked in 4.17.10). Should it be?

jdalton:
Hi @joshbuckley182!

The 4.17.10 bump was just for Node 10 compatibility feature work won't show up until v5.

PR #3596, opened 2018-01-19
Changed to function as documented.

Crysinyan:


jdalton:
Fixes #3597

jdalton:
Thanks @Crysinyan!

PR #3604, opened 2018-01-23
Pass key option to partition

furstenheim:
In partition it is not possible to access the key of the object. This is a bit inconsistent with the rest of the methods (map, forEach...).

I've added the code and the example


jdalton:
Hi @furstenheim!

For now, I'd like to keep the argument signature as it is.

PR #3615, opened 2018-01-29
Implement `defaultToAny` method

eran-nussbaum:


ypconstante:
How about adding a coalesce alias?

StreetStrider:
@ypconstante `coalesce` is a name that spinning in my mind from point when I saw that ticket.

eran-nussbaum:
@ypconstante @ypconstante I totally agree to that name.
I added it as an alias. Do I need to do anything other than add an `@alias` in JSDoc?...

Do you think I should just rename `defaultToAny` to `coalesce` instead?
I think that having its name start with `defaultTo` is good since it is closely-related to `defaultTo` method and its name should imply it.

StreetStrider:
@eran-nussbaum I personally prefer one-word, concise and pretty names.

eran-nussbaum:
@jdalton what's your take?

jdalton:
I not a fan of aliases and I like how discoverable `defaultToAny` is.

eran-nussbaum:
Okay.
So, do I need to do any further change to this PR? Is my implementation okay?…

jdalton:
You need to remove the `@alias` meta data, you could also use the built-in `array.reduce` if you wanted.

falsyvalues:
`defaultToAny` says nothing without checking what is behind in opposite to `coalesce` which is well known for this kind of operations.

ypconstante:
Maybe it would be better if `defaultTo` accepted multiple values instead of having `defaultTo` and `defaultToAny`

eran-nussbaum:
@jdalton do I need to do any more work on this PR for you to accept (and merge) it?

jdalton:
Hi @eran-nussbaum!

Nope, things are good!

PR #3616, opened 2018-01-30
Use posttest loops in chunk.js

bulgerxie:
I think use posttest loops(_do-while_)  will reduce one loop.


jdalton:
Thanks @bulgerxie!

I'm okay with how the loop is currently.

PR #3618, opened 2018-01-31
Add example to slice method

sudarsangp:


jdalton:
Thanks @sudarsangp!

I don't think slice will make it into v5 so I'm going to pass on this and just remove it instead.

Issue #3627, opened 2018-02-05
Documentation of _.slice() negative indexes


timo-lindemann-victors:
The documentation of _.slice() does not mention that negative indexing is supported at all (but it is):

    _.slice([1, 2, 3, 4], -1) // [4]
    _.slice([1, 2, 3, 4], -2) // [3, 4]
    _.slice([1, 2, 3, 4], -2, -1) // [3]

Is this not mentioned in the documentation intentionally, does it mean it's not reliable, or is it simply an omission (doc bug)?

jdalton:
Hi @timo-lindemann-victors!

I think the shoutout to the `Array#slice` method, which supports negative indexes, covers it.

timo-lindemann-victors:
Thanks for your reply.

I severely disagree. You document the parameters shortly thereafter, and nothing of the sort is mentioned there; in fact, you only reference Array#slice because you want to tell people that "dense arrays" are returned (which isn't helpful at all for this; ask any random JS dev and chances are they don't even know what that *is*). 

I am replying to this because now I cannot use negative indexing because superiors think it's unreliable (because undocumented) behaviour. If stringent documentation is not important to *you*, I can live with that; they, as it seems, can't.

jdalton:
Ah, that's a bummer. For them, `Array#slice` may be your best bet.

jnfingerle:
@jdalton So, there's an omission in the docs. Your argument that it's covered anyway is shown by @timo-lindemann-victors to be invald. But you don't want to fix it - because.  I'm a bit troubled, that complete documentation seems to be of no concern to you.

jdalton:
Thanks for the PR @jnfingerle!

PR #3631, opened 2018-02-06
Update documentation in slice.js

jnfingerle:
Even though #3627 has been closed, negative indexes should be documented.


jdalton:
Thanks!

PR #3635, opened 2018-02-08
Add: typeof func !== 'function'

yxuchen:


jdalton:
Thanks @yxuchen!

While my personal code style has changed to `===` this code base is currently `==` for anything guaranteed to be the expected type. In this case `typeof` can only return a string so `==` and `===` are functionality equivalent.

PR #3637, opened 2018-02-08
Fixed check on undefined

VladimirSchneider:


jdalton:
Hi @VladimirSchneider!

In this case our `undefined` is local so it's safe in pre-ES5 environments.

PR #3644, opened 2018-02-13
add uniq Map for generated ID

PxyUp:
Good day, to the best of our knowledge, a unique element must have context. Let's say each person is unique, they can be numbered, like cats, dogs and the like

```
uniqueId("cat") // "cat1"
uniqueId("cat") // "cat2"
uniqueId("dog") // "dog1"
uniqueId("dog") // "dog2"
uniqueId() // "1"
uniqueId() // "2"
```

### DIFF
--- a/uniqueId.js
+++ b/uniqueId.js
@@ -1,5 +1,5 @@
 /** Used to generate unique IDs. */
-let idCounter = 0
+let idCounter = {}
 
 /**
  * Generates a unique ID. If `prefix` is given, the ID is appended to it.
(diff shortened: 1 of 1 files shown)


jdalton:
I like this @PxyUp!

Thank you!

StreetStrider:
That's awesome, but where's the test? @PxyUp

jdalton:
The test debt will need to be paid soon. I'm wrapping up `@std/esm` v1 so that means I'll be able to focus more on Lodash v5.

falsyvalues:
I think we should alter also description .

omeid:
Why does it matter for the `uniqueId` to be serial?

PxyUp:
@omeid because uniq id, can be only in context

falsyvalues:
@omeid @PxyUp Worth to mention that this is the cheapest implementation that guarantee unique id in context.

PR #3647, opened 2018-02-17
Add string example to isEmpty docs

Dean177:
While this is covered in the the description above, there exist examples for arrays and objects and I think this gives 'at a glance' confirmation that the function works as intended with strings.


jdalton:
Thanks @Dean177!

PR #3648, opened 2018-02-18
Implement `same` method

cope:
Implement `same` method

### DIFF
--- a/same
+++ b/same
@@ -1,0 +1,28 @@
+import difference from './.difference.js'
+import isEmpty from './isEmpty.js'
+
+/**
+ * Checks if `array` and `values` have all the same items. Duplicate items are ignored.
+ *
+ * @since 0.1.0
+ * @param {Array} array The array to inspect.
+ * @param {Array} values The values to compare.
+ * @returns {boolean} Returns `true` if `array` and `values` have all the same items, else `false`.
+ * @see difference
+ * @example
+ *
+ * same([2, 1, 3], [1, 2, 3])
+ * // => true
+ *
+ * same([2, 1, 3], [1, 2])
+ * // => false
+ *
+ * same([2, 4], [1, 2, 3])
+ * // => false
+ */
+function same(array, values) {
+  return _.isEmpty(_.difference(array, values))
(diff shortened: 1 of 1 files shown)


jdalton:
Thanks @cope!

zaclummys:
@cope why not using `Array#includes` rather than `difference`?

cope:
@zaclummys I don't have a preference, TBH. I just used the first thing that worked for me.

Do you mean something like `return _.includes(array, values) && _.includes(values, array);`?

If that works, I would agree it's better.

zaclummys:
@cope,

I think `includes` is faster and more effective than `difference`, because the first only checks for the existence of a value rather than creating a new array with new values.

Your snippet looks good! How about adding also `values === array` to check whether both arguments point to the same object?

cope:
I agree.

Would this be acceptable?
```
import includes from './.includes.js'
import forEach from './forEach.js'

function containsAll (a1, a2) {
    let all = true;
    forEach(a2, x => all = includes(a1, x));
    return all;
}

function same (a1, a2) {
    return a1 === a2 ? true : containsAll(a1, a2) && containsAll(a2, a1);
}
```

zaclummys:
@cope,

You don't need to iterate over all values, you can stop the iteration (`break`/`return`)  when a value isn't included. In my snippet, `baseSame` do this trick.

zaclummys:
@cope, looks good!

It just add to my snippet a pre-test `array1.length === array2.length` to help to avoid unnecessary effort.

cope:
Copied the length check from you, that very much makes sense!

cope:
@jdalton Should I somehow update the PR, or should we first wait to see if the popularity of the request is going?

StreetStrider:
I would like to see this method merged. It is obvious that we use arrays here as sets, but builtin array type is still much more convenient and appropriate than `Set` in most of conditions. Besides, `Set` lacks such method either (that's nonsence for such type!). So, basically I still support any methods which allow to work with arrays as sets.

simonmilz:
+1

PR #3650, opened 2018-02-19
use isObjectLike module instead of (typeof value == 'object' && value!== null) in another modules

alirezavalizade:
being in the DRY law!


jdalton:
Thank you @alirezavalizade!

PR #3663, opened 2018-02-25
Implement `getParameterByName` method

tanrax:


jdalton:
Hi @tanrax!

I rather folks shim the WHATWG `URL` implementation. It has all the methods you need and in a spec compliant way.

Issue #3678, opened 2018-03-06
[Suggestion] Improve lodash.difference


zaclummys:
If both arrays point to the same object, return a clean array before the comparison iteration.

They will always have the same values.

```js
import 'lodash' from 'lodash';

const arr = [1, 2, 3]

const result = lodash.difference(arr, arr)
```

jdalton:
@zaclummys Good idea!

Would you be up for creating a PR?

zaclummys:
Due to `...values`, even though all the arguments point to the same object:

- The rest arguments won't be won't be strictly equal to the first, because they are wrapped in a new array.

- To verify the equality between the rest arguments and the first, it would be
 necessary an iteration ala `Array#every`.

jdalton:
Moved to #3679.

PR #3679, opened 2018-03-06
[PR] Improve lodash.difference

zaclummys:
https://github.com/lodash/lodash/issues/3678

### DIFF
--- a/difference.js
+++ b/difference.js
@@ -29,7 +29,7 @@ function difference(array, ...values) {
   
   const patched = map(values, x => x === array ? [] : x);
   
-  return baseDifference(array, baseFlatten(values, 1, isArrayLikeObject, true));
+  return baseDifference(array, baseFlatten(patched, 1, isArrayLikeObject, true));
 }
 
 export default difference
--- a/difference.js
+++ b/difference.js
@@ -1,8 +1,7 @@
 import baseDifference from './.internal/baseDifference.js'
 import baseFlatten from './.internal/baseFlatten.js'
 import isArrayLikeObject from './isArrayLikeObject.js'
-import every from './every.js'
-import eq from './eq.js'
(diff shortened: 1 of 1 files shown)


jdalton:
Is this worth the cost?

zaclummys:
I just wrote a simple benchmark to show it:

```js
const arr1 = Array(1e5);
const arr2 = Array(1e5);
```
> current diff arr1 arr2: 39.08477000892162ms
> current diff arr1 arr2: 39.205ms
> 
> current diff arr1 arr1: 29.104979991912842ms
> current diff arr1 arr1: 29.330ms
> 

> improved diff arr1 arr2: 17.764797002077103ms
> improved diff arr1 arr2: 17.969ms
> 
> improved diff arr1 arr1: 0.6249690055847168ms
> improved diff arr1 arr1: 0.840ms

zaclummys:
@jdalton, I believe there's a significant improvement in performance. What's our next step?

pedrofurtado:
@jdalton is this PR ready for merge, right? It's a great improvement in performance.

jdalton:
It's an OK improvement in perf when you happen to have the same arrays and the arrays are large. I don't know about the other cases.

zaclummys:
If this PR isn't OK, we can explore a way to improve it.
For example, we can add a pre-test: `values[0] === array && every(values, ...)`

pedrofurtado:
@zaclummys I think to add this pre-test is good idea. As soon as possible, show here a brenchmark with different  arrays sizes. So,  we Will sure about how many improvement was made by different lengths of arrays.

jdalton:
Thanks for you alls patience here. I'm going to hold off on this PR for now. If there's enough interest or a perf critical scenario arises where this can be applied then I'll totally revisit it.

Issue #1384, opened 2015-08-06
combining before and after functions


oriSomething:
sometimes, i find myself need a function that combine `_.before` and `_.after`. meaning, something like `beforeAfter(beforeFn, afterFn, n)`which create a `function` where there is counter that before `n` it runs the `beforeFn` and after it passes `n` it runs `afterFn`

jdalton:
Hey @oriSomething! Can you kick around a mixin and come back with it. That will help solidify your ideas and give something for others to :+1: around. We'll then be able to better gauge demand for it.

oriSomething:
this is kind of what i use for now
```js
_.mixin({
  beforeAfterTemporaryName: function(before, after, when) {
    when = _.isFinite(when) ? when : 1;
    var counter = 0;

    return function() {
      if (counter < when) {
        counter += 1;
        return before.apply(this, arguments);
      }

      return after.apply(this, arguments);
    };
  }
});
```

PR #3681, opened 2018-03-11
Add `afterBefore`

preethamvishy:
This method creates a function that invokes a function,  `func`, once it's called `start` or more times until it's called `end` times.

### DIFF
--- a/afterBefore.js
+++ b/afterBefore.js
@@ -1,0 +1,40 @@
+/**
+ * A combination of `after` and `before`. This method creates a function that invokes
+ * `func` once it's called `start` or more times until it's called `end` times.
+ *
+ * @since 5.0.0
+ * @category Function
+ * @param {number} start The number of calls before `func` is invoked.
+ * @param {number} end The number of calls at which `func` is no longer invoked.
+ * @param {Function} func The function to restrict.
+ * @returns {Function} Returns the new restricted function.
+ * @example
+ *
+ * var fn = afterBefore(2, 4, (num) => {
+ *    console.log(num)
+ * })
+ * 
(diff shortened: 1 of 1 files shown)


jdalton:
Thanks @preethamvishy!

preethamvishy:
Sounds good. 
Looking through the previous issues, I found #1384. This PR may be relevant.

cope:
I _really_ have an issue with the name :D

Isn't afterBefore = now? 😵

preethamvishy:
Haha probably. But I believe _afterBefore_ resonates more with the problem it's meant to solve than _now_.
Another strong contender could _between_? 
_after_ and _before_ are well know utilities and I like how lodash utilities are named after exactly what they do and I thought _afterBefore_ would be easy to discover.
 
Open to name changes if it gains traction and can be merged.

StreetStrider:
> I like how lodash utilities are named after exactly what they do

That's not always so. Like, this name is ambiguous. I can't point out is `afterBefore` time-related, call count -related or numeric. It refers to some ordered values, though, but it hasn't enough cerainty.

`between` got strong resemblance  with numeric ranges.

preethamvishy:
Can the same case of ambiguity relating time, call count, numeric etc be applied up for `after` and `before`? Both are popular utilities and I was banking on the popularity of both. I'm definitely open to the community's suggestions for naming if this PR is to be taken forward.

StreetStrider:
@preethamvishy I think `before`  and `after` suffers from the same ambgiguity as other simple names because of LoDash (and previously, underscore) have got single namespace.

PR #3682, opened 2018-03-11
Add `afterEvery`

preethamvishy:
This method creates a function that invokes `func` after every `n` times the function is called.

For example:
```javascript
const notifications = ['request', 'complete', 'pull', 'request']
const group = afterEvery(2, () => console.log('You have 2 new notifications'))

forEach(notifications, message => asyncFetch({ 'message': message, 'fetch': group }))
// => Logs 'You have 2 new notifications' twice after the four async fetches have
// completed.
```

### DIFF
--- a/afterEvery.js
+++ b/afterEvery.js
@@ -1,0 +1,39 @@
+/**
+ * This method creates a function that invokes `func` after
+ * every `n` times the function is called.
+ *
+ * @since 5.0.0
+ * @category Function
+ * @param {number} n The number of calls after which `func` is invoked each time.
+ * @param {Function} func The function to restrict.
+ * @returns {Function} Returns the new restricted function.
+ * @example
+ *
+ * const notifications = ['request', 'complete', 'pull', 'request']
+ * const group = afterEvery(2, () => console.log('You have 2 new notifications'))
+ *
+ * forEach(notifications, message => asyncFetch({ 'message': message, 'fetch': group }))
+ * // => Logs 'You have 2 new notifications' twice after the four async fetches have 
+ * completed.
+ */
+function afterEvery(n, func) {
+  if (typeof func != 'function') {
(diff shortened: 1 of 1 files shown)


jdalton:
Thanks @preethamvishy!

zaclummys:
Could you give us more examples of situations in which this method would be useful?

preethamvishy:
I found myself in situations that required a logic similar to this `afterEvery` on two previous occassions and they were my main motivation for creating this, @zaclummys.

1. I had a long running process which had a few hundred lines of console output. It was a set of very verbose ansible playbooks developed by multiple people. I wanted to monitor the process but without the hundreds of colourful console logs. Long story short, I piped the output and displayed every Nth line.
2. I was working on an app and I wanted the user to be able to set notification settings to be notified after 'Every 5 incoming messages', 'Every 10 incoming messages' and 'Every 25 incoming messages.'

Some use cases that I can think of right now are:
1. If you want to display every Nth incoming data element in a stream
```javascript
var notify = afterEvery(5, (data)=> {
      console.log(data.message)
})
stream.on('data', (data) => {
  notify(data)
})
```

2. If one uses a randomizing logic: randomizing after every N calls, wherever possible, could be a more efficient strategy than for every call/iteration
```javascript
let a = Math.random()
let ran = afterEvery(4, () => {
      a = Math.random()
})
for(var i=0; i<20; i++)
{
    ran()
    console.log(a)
}
```

3. Along the same lines, if one wants to use a key from a pool of keys for making API calls for N number of calls and then pick another key from the pool, `afterEvery` can be used.

zaclummys:
@preethamvishy,

Thanks for your clarification! I like your PR!

preethamvishy:
Thanks @zaclummys. I hope the community can find it useful.

PR #3708, opened 2018-03-24
Change all HTML characters to entity names.

labixiaoshuai:
I don't know why there are both entity names and entity numbers, so I changed all of them to entity names.


jdalton:
Hi @labixiaoshuai!

This is a legacy bit left over from earlier versions. We used `&#39;` because `&apos;` wasn't supported in IE8.

PR #3729, opened 2018-04-04
reset.js

narendrasinghrathore:
Useful to reset an object properties with default or given values as options passing as parameter in function call.


jdalton:
Hi @narendrasinghrathore!

I think this is better suited as a mixin since resetting is such a scenario driven thing. For example some may want to `delete` properties.

narendrasinghrathore:
@jdalton Okay.

PR #3736, opened 2018-04-06
Fix inclusion of litteral "+" in rsMisc

misogl:
`rsDigits` contained a "+" character but was also used in negated character class.


jdalton:
Thanks @misogl!

PR #3748, opened 2018-04-12
feat(getOr): add a `null`-handling getter

satblip:
Adds `_.getOr` function to avoid having to do :

```
_.get(data, 'key') || 'ifNull'
```

to protect for `null` data.

### DIFF
--- a/getOr.js
+++ b/getOr.js
@@ -1,0 +1,33 @@
+import baseGet from './.internal/baseGet.js'
+
+/**
+ * Gets the value at `path` of `object`. If the resolved value is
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @satblip!

The [`_.get`](http://lodash.com/docs/#get) method allows passing a default value param.

satblip:
@jdalton yes, but that default, is only returned when the result is `undefined` idea here is to allow a handling for a `null` returned

jdalton:
Ah, sorry I missed that. I think a mixin is fine in that case.

Issue #3757, opened 2018-04-17
Clarify _.union example in docs


tyrw:
Currently in the docs, the example for `_.union` is

```
_.union([2], [1, 2]);
// => [2, 1]
```

It's not clear from this that `union` is taking all unique values in either array rather than all values contained in both arrays.  Suggest changing the example to

```
_.union([2, 3], [1, 2]);
// => [2, 3, 1]
```

Seems like this might not warrant a PR, but I can do one if preferred.

jdalton:
Sure, a PR would be great! Just modify the in source example.

tyrw:
PR here https://github.com/lodash/lodash/pull/3758

jdalton:
Moved to #3758

PR #3758, opened 2018-04-17
Update _.union documentation

tyrw:
Clarifying example as per https://github.com/lodash/lodash/issues/3757


jdalton:
Thanks @tyrw!

Issue #3763, opened 2018-04-22
The case for custom compare function for sorting


ryanhomer:
There seems to be a growing number of cases in which we need to sort based on custom rules as evidenced by the following posts:

- [Is it possible to pass a custom comparator to lodash's sortBy function?](https://stackoverflow.com/questions/43596210/is-it-possible-to-pass-a-custom-comparator-to-lodashs-sortby-function?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa)
- [sortBy comparator is called on arrays with a single item](https://github.com/lodash/lodash/issues/2090#issuecomment-273485006)
- [comparator for set/array functions (_.difference, _.union etc)](https://github.com/lodash/lodash/issues/246#issuecomment-159917353)
- [Sort Array Elements (string with numbers), natural sort](https://stackoverflow.com/questions/15478954/sort-array-elements-string-with-numbers-natural-sort/)

I'd also like to add that custom collation varies a lot based on language, which [`String.prototype.localeCompare()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) tries to solve. There are also various language romanizations that may need custom collation, such as [Hanyu Pinyin](http://pinyin.info/readings/zyg/rules.html) for which order depends on tone marks, e.g.: ā, á, ă, à, a (in that order).

None of this seems possible with the current implementation of `_.orderBy`. However, I've come up with a solution that requires minimal changes to `compareMultiple` and would allow `orderBy` to take an array of `order` strings (`['asc', `desc`]`) or compare functions like this:

    const data = [
      { f1: "z", f2: 40},
      { f1: "z", f2: 7},
      { f1: "z", f2: 1},
      { f1: "z", f2: 5},
      { f1: "a", f2: 10},
      { f1: "a", f2: 2},
      { f1: "A", f2: 3},
      { f1: "ä", f2: 1},
    ];

    _.orderBy(data, [
      item => _.toLower(item.f1),
      item => item.f2,
    ], [
      (a, b) => a.localeCompare(b, 'da'),
      (a, b) => a - b,
    ]);

The code:

    function compareMultiple(object, other, orders) {
      var index = -1,
          objCriteria = object.criteria,
          othCriteria = other.criteria,
          length = objCriteria.length,
          ordersLength = orders.length;

      while (++index < length) {
        var order = index < ordersLength ? orders[index] : null;
[...]

PR #3764, opened 2018-04-22
Allow compareMultiple/orderBy to accept compare functions

ryanhomer:
See #3763 for rationale.

- This may negate the need for locale-specific fixes in `compareAscending.js` such as #3602.
- Test snippet below to be added to other `orderBy` tests

```
    QUnit.test('should work with `orders` specified as compare functions', function(assert) {
      assert.expect(3);

      function compareStrings(a, b) { return a < b ? -1 : a > b ? 1 : 0 }
      function compareStringsDesc(a, b) { return -compareStrings(a, b) }
      function compareNumbers(a, b) { return a - b }

      var actual = _.orderBy(objects, 'a', compareStringsDesc);
      assert.deepEqual(actual, [objects[1], objects[3], objects[0], objects[2]]);

      actual = _.orderBy(objects, ['a'], [compareStringsDesc]);
      assert.deepEqual(actual, [objects[1], objects[3], objects[0], objects[2]]);

      actual = _.orderBy(objects, ['a', 'b'], [compareStringsDesc, compareNumbers]);
      assert.deepEqual(actual, [objects[3], objects[1], objects[2], objects[0]]);
    });
```

### DIFF
--- a/.internal/compareMultiple.js
+++ b/.internal/compareMultiple.js
@@ -11,7 +11,7 @@ import compareAscending from './compareAscending.js'
  * @private
  * @param {Object} object The object to compare.
  * @param {Object} other The other object to compare.
- * @param {boolean[]|string[]} orders The order to sort by for each property.
+ * @param {(string|function)[]} orders The order to sort by for each property.
  * @returns {number} Returns the sort order indicator for `object`.
  */
 function compareMultiple(object, other, orders) {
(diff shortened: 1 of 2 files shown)


kikonen:
Q: When one could expect release containing this comparator addition to be available?

jdalton:
Hi @kikonen!

This did not land in v4 so will have a possibility of showing up in v5
(a future major release, whenever that is).

PR #3777, opened 2018-05-05
isArrayLike false detection

swznd:
`isArrayLike` return `true` when Object has length properties

```
_.isArrayLike({ name: 'video.mp4', length: 26210 });
=> true
```

### DIFF
--- a/isArrayLike.js
+++ b/isArrayLike.js
@@ -24,7 +24,7 @@ import isLength from './isLength.js'
  * // => false
  */
 function isArrayLike(value) {
-  return value != null && typeof value != 'function' && isLength(value.length)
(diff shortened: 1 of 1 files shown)


swznd:
Hi @jdalton 

could you tell me why you marked as invalid? Because this bug produce error in `_.forEach`

```
_.forEach({ 'a': 1, 'b': 2, length: 12 }, function(value, key) { console.log(key, value) });

=> 0 undefined
1 undefined
2 undefined
3 undefined
4 undefined
5 undefined
6 undefined
7 undefined
8 undefined
9 undefined
10 undefined
11 undefined
```

jdalton:
Hi @swznd!

The behavior of [isArrayLike](https://lodash.com/docs/#isArrayLike) is intentional.

PR #3786, opened 2018-05-11
Added short circuit for typed arrays

twolfson:
We're using `cloneDeep` for large typed arrays (80k entries or more). We noticed that `cloneDeep` was taking a lot longer than expected for these data sets. After some sleuthing, it looks like the culprit is `baseClone` trying to iterate over the keys of the typed array and copy over its values

We've decided to put in a short circuit clause (not sure if I did it right -- there's probably a slicker way with the `tag` string) to prevent this from happening

We've seen a decrease from 300ms to 3-30ms for 80k entries on an iPad mini

```js
console.time('a');
lodash.cloneDeep(new Float32Array(80e3));
console.timeEnd('a');
```

In this PR:

- Added short circuit for typed arrays

**Notes:**

We'd be happy to write performance tests as well but couldn't seem to find a location to do so

### DIFF
--- a/.internal/baseClone.js
+++ b/.internal/baseClone.js
@@ -218,6 +218,10 @@ function baseClone(value, bitmask, customizer, key, object, stack) {
     return result
   }
 
+  if (isTypedArray(value)) {
+    return result;
+  }
+
   const keysFunc = isFull
     ? (isFlat ? getAllKeysIn : getAllKeys)
     : (isFlat ? keysIn : keys)


jdalton:
Hi @twolfson!

I don't think short circuiting is going to work since that would return an incomplete clone. It's probably a better idea to create a custom clone method more suited for your needs.

twolfson:
Thanks for the quick reply! Could you elaborate on what you mean by an incomplete clone? Typed arrays cannot have anything but numeric data so it's always primitive values being copied

rreusser:
Am I correct in understanding that the problem is this situation and that there's no reasonable way to filter out the numbers? Makes sense, but is too bad since I can only imagine typed arrays without additional properties for which the fast method is preferable probably accounts for probably >99% of use cases.

```javascript
> x = new Float32Array(3)
> x.foo = 'bar'

> Object.getOwnPropertyNames(x)
< ▶︎ (4) ["0", "1", "2", "foo"]

> Object.getOwnPropertyDescriptors(x)
< ▼ {0: {…}, 1: {…}, 2: {…}, foo: {…}}
    ▶︎ 0: {value: 0, writable: true, enumerable: true, configurable: false}
    ▶︎ 1: {value: 0, writable: true, enumerable: true, configurable: false}
    ▶︎ 2: {value: 0, writable: true, enumerable: true, configurable: false}
    ▶︎ foo: {value: "bar", writable: true, enumerable: true, configurable: true}
    ▶︎ __proto__: Object
```

PR #3787, opened 2018-05-11
Remove apply from memoized function call.

Mtnt:
It`s possible to call a func without changing its context.


jdalton:
Hi @Mtnt!

We still need the `this` binding of `apply`.

Mtnt:
@jdalton why?

Mtnt:
@jdalton could you explain why you still need `this` binding?

Mtnt:
@jdalton ok, thanks, I just clarify for myself impossibility of double binding =)

PR #3804, opened 2018-05-29
CastArray: Add example for existing array value

spen:
### Summary

While reading the examples for `castArray` I noticed that many cases are covered, but not the case in which the given value is already an array. The result seems obvious enough, but for the sake of clarity and a fuller set of example cases I thought I'd add this regardless.

### Testing

As this is just an example, no actual testing is necessary. Instead, verify that `_.castArray([1])` does in-fact return `[1]`


jdalton:
You could assign it to a value to show that it is `===`:
``js
const array = [1]
castArray(array) === array
// => true
```

spen:
Ah, now I realise that there's a similar example (the last one) that shows this, albeit in a less direct way. Shall I close in that case?

jdalton:
Ah yes, it is there 😋 

I'll go ahead and close then then. Thanks!

spen:
NP Thanks for taking a look anyway! 😄

PR #3809, opened 2018-06-01
Create fromEntries.js

ycmjason:
this acts as an alias for `_.fromPairs` since entries seem to be the official name for those pairs. (`Object.entries`)


jdalton:
For v5 we'll likely just rename it to the official name. If you'd like to modify the PR to do that, that would be rad.

ycmjason:
@jdalton should I keep `_.fromPairs` as an alias?

jdalton:
You can drop `fromPairs`.

jdalton:
Thanks @ycmjason!

PR #3848, opened 2018-06-27
Update difference.js

Shubham-koli:
There was one extra star. I don't think we need it.


jdalton:
Hi @Shubham-koli!

The extra star is just part of the existing code style. Not needed but there for consistency and to designate a method comment block.

PR #3853, opened 2018-06-28
Clarify doc for `range` usage with single arg

doug-a-brunner:
I found the explanation of `range` usage with a single argument to be a bit confusing. This rearranges slightly to be less ambiguous.


doug-a-brunner:
Closing because CLA check seems to be wedged.

PR #3854, opened 2018-06-28
Clarify doc for `range` usage with single arg

doug-a-brunner:
I found the explanation of range usage with a single argument to be a bit confusing. This rearranges slightly to reduce ambiguity.

Recreated PR to fix CLA-bot problem due to different emails in use.


jdalton:
Thanks @doug-a-brunner!

PR #3857, opened 2018-07-01
Document that get supports arrays

lukeapage:
_.get appears to support arrays fine, but it is undocumented.
It is useful if navigating through an object structure containing arrays.

### DIFF
--- a/get.js
+++ b/get.js
@@ -5,9 +5,9 @@ import baseGet from './.internal/baseGet.js'
  * `undefined`, the `defaultValue` is returned in its place.
  *
  * @since 3.7.0
- * @category Object
- * @param {Object} object The object to query.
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @lukeapage!

The `get` method is a generic object method.
It doesn't treat arrays any differently than any other object.

lukeapage:
Does that mean it can be used with arrays ? Or we should avoid using it with arrays because it might break in the future?

lukeapage:
And in terms of passing a number for the path.. is that supported or currently works by accident?

jdalton:
Arrays, functions, regexeps, dates are all objects so generic object methods will work on them but they will just be treated like any-old object.

PR #3861, opened 2018-07-10
Update `filterObject` docs to reference `pickBy`

cooperka:
This is clearly an area of confusion -- see [this question](https://stackoverflow.com/questions/30726830/how-to-filter-keys-of-an-object-with-lodash) as an example. Hopefully this helps.

### DIFF
--- a/filterObject.js
+++ b/filterObject.js
@@ -3,12 +3,14 @@
  * `predicate` returns truthy for. The predicate is invoked with three
  * arguments: (value, key, object).
  *
+ * **Note:** If you want an object in return, consider `pickBy`.
+ *
  * @since 5.0.0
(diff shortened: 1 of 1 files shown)


filterObject.js
```diff
@@ -3,12 +3,14 @@
  * `predicate` returns truthy for. The predicate is invoked with three
  * arguments: (value, key, object).
  *
+ * **Note:** If you want an object in return, consider `pickBy`.
+ *
```
jdalton: @cooperka Could you reword
> If you want an object in return, consider `pickBy`.

cooperka:
@jdalton done!

jdalton:
Thank you!

PR #3866, opened 2018-07-12
Readme color images

circa10a:
Add some color images generated by [carbon.now.sh](https://carbon.now.sh) to the README


jdalton:
Hi @circa10a!

Thank you for these. I :heart: carbon.now.sh!
I want to avoid this though because I find it's nice to be able to copy/paste usage examples and with an image we lose that.

PR #3876, opened 2018-07-22
Added reverse split functionality

cspencer0101:
/**
 * Splits `string` by `separator`, but in reverse
 *
 * **Note:** This method is based on
 * [`split`](https://github.com/lodash/lodash/blob/master/split.js).
 *
 * @since 4.0.0
 * @category String
 * @param {string} [string=''] The string to split.
 * @param {RegExp|string} separator The separator pattern to split by.
 * @param {number} [limit] The length to truncate results to from the end.
 * @returns {Array} Returns the string segments.
 * @example
 *
 * rsplit('a-b-c', '-', 2)
 * // => ['b', 'c']
 */

### DIFF
--- a/rsplit.js
+++ b/rsplit.js
@@ -1,0 +1,51 @@
+import castSlice from './.internal/castSlice.js'
+import hasUnicode from './.internal/hasUnicode.js'
+import isRegExp from './isRegExp.js'
+import stringToArray from './.internal/stringToArray.js'
+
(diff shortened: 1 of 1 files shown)


cspencer0101:
This is just creating symmetry with the lodash version of the split functionality.  As such, I have tried to leave as much of the split function's source code intact as possible.

jdalton:
Thanks @cspencer0101 !

PR #3885, opened 2018-07-25
add classOf return type of value

andyxmq:
return type of value


jdalton:
Hi @andyxmq!

I'd like to hold off for now. There may be an existing request that you can upvote with the :+1: reaction. I have avoided a public toString tag method like this to encourage folks to cherry-pick the type methods they need instead of pulling in a larger all-in-one method.

Issue #2284, opened 2016-04-28
Feature suggestion: _.mode and _.modeBy (statistics function)


jrencz:
Since some math functions have been added recently I thought It may be useful to add a mode function.

I implemented one as a mixin (based on http://stackoverflow.com/a/18878650):

```js
_.mixin({
  modeBy: (collection, iteratee) => _
    .chain(collection)
    .countBy(iteratee)
    .toPairs()
    .maxBy(_.last)
    .head()
    // in this implementation string "undefined" is returned if iteratee returning 
    // non existing property is given
    .thru(value => value === 'undefined' ? undefined : value)
    .value(),
})
```
but it has its downsides.

I suggest this function might be added to lodash core.

My use case:
From collection of objects I need to get the most frequently occurring value of some property.

Florian-R:
Dupe of #2184

jrencz:
@Florian-R thanks for spotting that.

Issue #2184, opened 2016-03-31
New math method - Mode


SHAHRUKH-KHAN:
I would like to add a new method: Mode, which would calculate the mode of a collection.

jdalton:
Thanks for the request @SHAHRUKH-KHAN!
We'll keep an eye on user demand. (be sure to upvote with the :+1: reaction).

Related to #2182.

jfmengels:
Not sure what a mode is, but for consistency you might want a `modeBy` function too.

SHAHRUKH-KHAN:
@jfmengels mode is the maximum number of occurrences of a number in an array.
eg -> _.mode([1,1,2,2,3,3,3,3]) should give you [3].

StreetStrider:
@SHAHRUKH-KHAN, in your example you got 2 of ones, 2 of twos and 4 of threes, so mode = `4`, why is it `[3]`?

SHAHRUKH-KHAN:
@StreetStrider sorry My Bad ! . The number which has the highest frequency is the `mode`

SHAHRUKH-KHAN:
@jfmengels We can have `_.modeBy` too.

arieljake:
could we also add median while we are at it?

SHAHRUKH-KHAN:
@arieljake Surely we can ! 
First let this one go through :)

jpleclerc:
Why not add a statistic module? It would contain things like variance, standard deviation, coefficient of variation, mean, mode, median, trimmed mean...

jfmengels:
Wondering whether the method should be called `mode` or something like `mostCommon`. A few reasons:
- `mode` might be the appropriate statistical/mathematical name, but it's not meaningful to me. For the same reason, I would never search for `mode` if I searched for this functionality (it's not even a problem with English, the same word is used for the same thing in my native language).
- Having `_.mode` will pretty much lead to having a `_.modeBy`, which should probably return the first item whose `iteratee` value is the "mode".
Example:
```js
var votes = [
  {nominee: 'future president'},
  {nominee: 'future president'},
  {nominee: 'future president'},
  {nominee: 'future president'},
  {nominee: 'not future president'},
  {nominee: 'not future president'},
  {nominee: 'not future president either'}
];
_.modeBy(votes, 'nominee');
// => {nominee: 'not future president'},
```
[...]

PR #3906, opened 2018-08-09
Added missing most commonly used statistical functions

abibell:
Added mode, median, variance and standardDeviation plus the "By" variants - #2284 #2184 

### Tests
Since the master branch tests are in flux as per https://github.com/lodash/lodash/blob/master/.github/CONTRIBUTING.md, I am unable to add tests. You can do it yourself or I can add when test running issue is resolved.

### DIFF
--- a/median.js
+++ b/median.js
@@ -1,0 +1,19 @@
+import baseMedian from './medianBy.js'
+
+/**
+ * Computes the median of the values in `array`.
+ *
+ * @since 4.0.0
+ * @category Math
+ * @param {Array} array The array to iterate over.
+ * @returns {number} Returns the median, sorts the list and gets the (n+1)/2 th number.
+ * @example
+ *
+ * median([4, 2, 8, 6])
+ * // => 4
+ */
+function median(array) {
+  return baseMedian(array, (value) => value)
+}
+
+export default median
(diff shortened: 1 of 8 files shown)


abibell:
I signed the CLA with my Github account. I can't understand why it still thinks it is not signed. I made sure the email was set too later.

jdalton:
Thanks for this @abibell!

This is great but I want to hold off on adding these methods for now. I think these would probably be better suited/supported in a lib specifically for these types of methods.

As for the CLI see [this](https://cla.js.foundation/lodash/lodash?pullRequest=3906)

abibell:
@jdalton 

- **CLA** Maybe there is a defect in CLA. I don't see any action buttons on here. I already signed it. capture Please advise
- **Library** I am trying to understand your reasoning. lodash currently supports some of the methods - sum, min, max and mean. What is this library you are referring to? By the looks of it, enough people have voted for a median method. Are you suggesting they don't belong in lodash or they don't belong in lodash's Math category? I am sorry I am trying to understand the specific tangible reasoning.

jdalton:
Ah boo. It should have shown what was causing the CLA fail. It usually comes down to your commit email/user not matching with the user of the commit.

Issue #3359, opened 2017-09-05
Vulnerable Regular Expression


cristianstaicu:
The following regular expression used in words.js is vulnerable to ReDoS:
```js
/[a-z][A-Z]\|[A-Z]{2,}[a-z]\|[0-9][a-zA-Z]\|[a-zA-Z][0-9]\|[^a-zA-Z0-9 ]/
```
This regex is called by multiple methods exposed to the users such as "camelCase". The slowdown is moderately low: for 50.000 characters around 2 seconds matching time. However, I would still suggest one of the following: 
- remove the regex,
- anchor the regex,
- limit the number of characters that can be matched by the repetition,
- limit the input size.

If needed, I can provide an actual example showing the slowdown.

cristianstaicu:
I agree that for the general case that is a very large value, but security is about corner cases. Let us consider that for the camel case export alone (https://www.npmjs.com/package/lodash.camelcase) there are 237 dependents. I am pretty sure it can produce unexpected results in some packages. For example this package which seems to be the most popular one:
https://www.npmjs.com/package/css-loader

It looks like it just applies the camel case package on random CSS files you can specify using URLs. I am pretty sure there are nastier things than this, such as parsing HTTP headers which an attacker can easily control. 

I would like to submit a PR, but you have to tell me if any of the following options are acceptable: 
1) transform the regex into :
```js
/[a-z][A-Z]\|[A-Z]{2,2000}[a-z]\|[0-9][a-zA-Z]\|[a-zA-Z][0-9]\|[^a-zA-Z0-9 ]/
```
This effectively limits the word size to 2000 characters

2) check that the string parameter in "words" function has at most 20,000 characters. 

If, as you say, most of the use cases involve fewer number of characters, this should impact none of these use cases, but my prediction is that soon you will get new bug reports from angry users that random errors pop up in their application. :) Any other ideas?

jdalton:
Limiting the size is fine if there's no other way. The regexp is a monster so if there's a way to improve that, I'd prefer that route.

gibson042:
> The regexp is a monster so if there's a way to improve that, I'd prefer that route.

Well, it seems to break down into the following alternatives:
* lowercase a-z, followed by uppercase
* two or more uppercase A-Z, followed by lowercase
* 0-9 followed by A-Z (any case)
* A-Z (any case) followed by 0-9
* any character that is not a space or 0-9 or A-Z (any case)

[...]

PR #3925, opened 2018-08-28
Prevent ReDoS

manuel-jasso:
To fix https://github.com/lodash/lodash/issues/3359, modified reHasUnicodeWord to remove an unnecessary comma which made the regex greedy, this is only a test regex and not a matching regex. Added unit tests, this now should run under 5 ms instead of over 1000 ms for huge 50k+ char words.


manuel-jasso:
So sorry, I know what I did wrong, I will fix my email and re-submit. Sorry again...

jdalton:
No worries. Thank you for being persistent!

PR #3932, opened 2018-08-31
Add extended and supplement combining marks ranges

### LINKED ISSUE #3930: Include extra combining ranges
rsComboRange  in UnicodeSize.js does not include the Combining Diacritical Marks Extended (1AB0–1AFF) and Combining Diacritical Marks Supplement (1DC0–1DFF) ranges – I'm guessing these should also be included?

dracos:
I wasn't sure if there was a reason the same variables were defined in multiple files, so I left them as is, and added in all relevant locations. Fixes #3930.

This adds tne other combining marks ranges, though please note that this regex does not fully follow the UAX 29 algorithm for counting grapheme clusters which is quite a bit more complex. But this is hopefully a little improvement :)


jdalton:
Thank you @dracos!

Issue #2142, opened 2016-03-22
Propose _.isJSON


Ben52:
```
function isJSON(str) {
   return !_.isError(_.attempt(JSON.parse.bind(null, str)));
}
```

Any interest?

jdalton:
Not at this time.

nodesocket:
Is:

````javascript
function isJSON(str) {
   return !_.isError(_.attempt(JSON.parse, str));
}
````

Still the best way to do this? Would the following native JS be more performant?

````javascript
function isJSON(str) {
    try {
        var obj = JSON.parse(str);
        if (obj && typeof obj === 'object' && obj !== null) {
            return true
        }
  } catch (err) {}
  return false;
}
````

StreetStrider:
@nodesocket `null` is valid JSON.

richardstrutt:
Cleanest form: ```isJSON: json => !!json && !Array.isArray(json) && typeof json === 'object',```

PR #3976, opened 2018-09-27
isJSON Method

smusali:
We know that `object` type may include several things but for confirming whether the object is JSON-like, having `isJSON` method is necessary.

There should not be any error regarding test and style but the importance of this matters, and I think by most `_` users this method is going to be loved.

Thanks in advance!

### DIFF
--- a/isJSON.js
+++ b/isJSON.js
@@ -1,0 +1,30 @@
+import isObjectLike from './isObjectLike.js'
+import isArrayLike from './isArrayLike.js'
+
+/**
+ * This method checks if `value` is JSON-formatted object.
+ *
+ * @since 4.18.0
+ * @category Lang
+ * @param {*} value The value to check.
+ * @returns {boolean} Returns `true` if `value` is a JSON object,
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @smusali!

Thanks for taking time to make the PR.
This is actually a dup of to #1481 and #2142.

smusali:
@jdalton , thanks a lot for taking a look and responding!

The name of the method may look the same but if you take a look at #1481 closely, you'll see that it's for validating JSON string not the JSON object - for that method, `isJSONString` might be the better name.

Mine checks whether it is actual JSON or not.

jdalton:
JSON can represent a variety of values (string, number, array, object, etc.) so this kind of method wouldn't handle that. As presented this method is more like [_.isPlainObject](https://lodash.com/docs/#isPlainObject).

smusali:
Well, I didn't know this - my bad; so sorry - now I can code in peace. Thanks a lot, @jdalton

PR #3977, opened 2018-09-27
Fix comments of sortBy's result.

daenamkim:
Hi,

While I was reading docs I found results of sortBy were little bit different.
So I fixed what I checked out.
If I was wrong just tell me. 😄


daenamkim:
@jdalton Thanks you! I got you. 👍

PR #3979, opened 2018-09-27
currying added to add.js

ramesharr:


jdalton:
Hi @ramesharr!

Thank you for taking the time to create a PR.
We have currying with `lodash/fp/add`.

PR #3982, opened 2018-10-01
--save no longer needed

JoeCortopassi:
`--save` is on by default as of [npm 5](https://blog.npmjs.org/post/161081169345/v500), so `npm install aphrodite` is functionally equivalent to `npm install --save aphrodite` now

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ In a browser:
 Using npm:
 ```shell
 $ npm i -g npm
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @JoeCortopassi!

I believe `nvm` for Node 6 still uses npm@3 so I want to keep the `--save` command around a bit longer.

Issue #3992, opened 2018-10-08
Convert string to snake case behavior.


scvhost:
Hi, I curious about the behavior of snakeCase in lodash.
In this case, if i want to convert the "itemV2" to snake case.
The result should be "item_v2" , but the result from using snakeCase method is "item_v_2".
So, do you have any idea about this ?

jdalton:
Hi @scvhost!

We've found splitting words up to not be so cut and dry. We now side with treating `itemV2` as 3 compound words.

jdalton:
Hi @dwjft!

You're better off using a one-off custom snake case module or  package.

PR #3993, opened 2018-10-08
add snakeCase example with digit

abdullahtariq1171:
add `snakeCase` example with digit

refers to https://github.com/lodash/lodash/issues/3992


jdalton:
Thanks @abdullahtariq1171!

PR #4005, opened 2018-10-13
add new method eachElementCount for array

ar000n:
takes an array and return an object
which says each array elements occurrence
it works for even array of objects


jdalton:
Hi @ar000n!

I think would be a better fit as a separate mixin/package.

PR #4006, opened 2018-10-13
Made a discord server

ndr3w221:
Made a discord server, also updated the readme.


jdalton:
Thanks @ndr3w221!

I want to focus community voice to a single place to get questions answered. For now that's gitter, but we can always re-evaluate if/when the service is no longer sufficient.

ndr3w221:
Discord is very helpful, ytdl-core also uses discord

PR #4010, opened 2018-10-15
fix(lib): remove proto from hash object

PxyUp:
I deleted prototypes for all objects created for hash maps, to remove a potential error


jdalton:
@PxyUp Thanks for creating this PR!

However, I think we'll keep the existing behavior for now. In the future I may move to creating objects with `{ __proto__: null }` but I'm not there just yet.

PR #4011, opened 2018-10-15
Fix _.omit mutating objects w/ Arrays

### LINKED ISSUE #4007: _.omit is mutating objects
About version: `4.17.10`

Noticed that `_.omit` is mutating objects when the reference includes a array position:

```javascript
  const a = {
    id: 12,
    name: 'Jimmy',
    phoneNumbers: [
      {id: 12, number: '123 456-7980', type: 'Home'},
      {id: 13, number: '123 654-0987', type: 'Mobile'},
    ]
  };

  const newA = _.omit(a, ['id', 'phoneNumbers.0.id', 'phoneNumbers.1.id']);
  console.log(`a.phoneNumbers is now`, a.phoneNumbers[0].id);
  console.log(`Does this look like a problem? `, a.phoneNumbers[0].id == null);

  const b = {
    id: 12,
    name: 'Jimmy',
    phoneNumbers: [
      {id: 12, number: '123 456-7980', type: 'Home'},
      {id: 13, number: '123 654-0987', type: `'Mobile'},`
    ]
  };

  const newB = _.omit(_.cloneDeep(b), ['id', 'phoneNumbers.0.id', 'phoneNumbers.1.id']);
  console.log(`b.phoneNumbers is still`, b.phoneNumbers[0].id);
  console.log(`Would a cloneDeep 'solve' the problem? `, b.phoneNumbers[0].id != null);
```

Why is this a problem? Well, I use omit to skip the ids when running functional and integration tests. For my scenario, creating a function like:
```javascript
  function _omit<T>(value: T, omitKeys: string[]) {
    const clonedValue = _.cloneDeep(value);
    return _.omit(clonedValue, omitKeys);
  }
```
Solves my problem. 

However, I was still surprised by this fact and decided to create an issue.

ccjmne:
Fixes #4007

The culprit was the private function `customOmitClone`, which would be 
used to customize `_.cloneDeep` by specifically requesting all non-plain 
Objects to not be cloned.

Turns out that Arrays don't pass the `_.isPlainObject` test.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -5594,7 +5594,7 @@
      * @returns {*} Returns the uncloned value or `undefined` to defer cloning to `_.cloneDeep`.
      */
     function customOmitClone(value) {
-      return isPlainObject(value) ? undefined : value;
+      return isPlainObject(value) || isArray(value) ? undefined : value;
     }
 
     /**
--- a/test/test.js
+++ b/test/test.js
@@ -16516,12 +16516,12 @@
     });
 
     QUnit.test('should not mutate `object`', function(assert) {
-      assert.expect(4);
+      assert.expect(5);
 
-      lodashStable.each(['a', ['a'], 'a.b', ['a.b']], function(path) {
-        var object = { 'a': { 'b': 2 } };
+      lodashStable.each(['a', ['a'], 'a.b', ['a.b'], 'c[0]'], function(path) {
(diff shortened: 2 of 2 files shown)


ccjmne:
Hello @jdalton 👋 

I'm pretty new to the community process and might have missed something (even though I did read through the contribution guidelines).

I know you're dropping `_.omit` soon, but I wanted to try out contributing and that's the only issue currently open!

Please let me know if there's anything I should do differently 🙂

PR #4027, opened 2018-10-27
add minus example to ceil function

alireza-mh:
Add minus integer number to ceil function.

### DIFF
--- a/ceil.js
+++ b/ceil.js
@@ -1,7 +1,7 @@
 import createRound from './.internal/createRound.js'
 
 /**
- * Computes `number` rounded up to `precision`.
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @alireza-mh!

Instead of an additional value example would you mind tweaking this to describe `ceil` as a similar to [`Math.ceil`](https://mdn.io/ceil). That way it provides more context for usage.

alireza-mh:
Hi @jdalton,
i updated the PR.
by adding that example i was intended to avoid some mathematical errors.
i added `rounded up` definition by borrowing it from MDN web doc.

jdalton:
Thank you @alireza-mh!

PR #4029, opened 2018-10-28
Compress JSON Object Method

smusali:
This is the deep levelled JSON Object compression which is similar to `filterObject` but applies the function to all (the values of `PlainObject` values, the elements of `Array` values, and regular values). Please, let me know the necessity, since I have seen the necessity in my works. It is still undone and adding newer things time-by-time.

Thanks a lot in advance for your effort to review!


jdalton:
Hi @smusali!

I think this is better suited as a mix-in for now.

PR #4031, opened 2018-10-28
Simplifies the function flow

wjsc:
My attempt was to simplify the function using functional programming. 

Besides, this code avoid the use of variables and methods such as call and apply.

And by last, it reuses the castArray function.

### DIFF
--- a/flow.js
+++ b/flow.js
@@ -1,3 +1,5 @@
+import castArray from './castArray.js'
+
 /**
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @wjsc!

In this case, since this wrapper sits in front of the other function calls in flow I want to keep things barebones loops.

PR #4035, opened 2018-10-31
Add lodash-es description for ES6 scripts

alireza-mh:


jdalton:
Hi @alireza-mh!

Could you re-work this against the `npm` branch and add a note about using `lodash-es` for it.

alireza-mh:
Hi @jdalton .
thanks for your reply. 
Could you be more specific about `lodash-es` note and working on `npm` branch?
should i add examples for `lodash-es` or should i note that it support webpack tree shaking or ... ?
should i replace latest changes of `master` branch README.md with `npm` README.md as well?
also what is the benefit of working on `npm`?

alireza-mh:
Hi @jdalton .
thanks for your reply. 
Could you be more specific about `lodash-es` note and working on `npm` branch?
should i add examples for `lodash-es` or should i note that it support webpack tree shaking or ... ?
should i replace latest changes of `master` branch README.md with `npm` README.md as well?
also what is the benefit of working on `npm`?

jdalton:
Maybe instead of the ESM syntax snippet you could just add a line that says something like:

> Looking for Lodash modules written in ES6 or smaller bundle sizes? Check out [`lodash-es`](http://npmjs.com/lodash-es).

alireza-mh:
i guess this only helps if users use latest versions of webpack.
but it will make sense in the near future.
i will remove imports snippets and add your sentence.
the reason behind writing example was observing  `lodash` library(full library) on projects just for a few functions.

jdalton:
Thanks @alireza-mh!

PR #4054, opened 2018-11-09
add PascalCase

alireza-mh:


jdalton:
Hi @alireza-mh!

This is a dup of #942.

PR #4076, opened 2018-11-25
Call upperFirst directly, instead of capitalize, in camelCase

### LINKED ISSUE #4075: camelCase calls toLowerCase twice
The camelCase function code is calling toLowerCase on each word:
https://github.com/lodash/lodash/blob/136d5253e724838ca701fb54db09f18b618c127e/camelCase.js#L23-L28

The capitalize function also calls toLowerCase though, so this is being performed twice on each word (except the first):
https://github.com/lodash/lodash/blob/136d5253e724838ca701fb54db09f18b618c127e/capitalize.js#L16

Since string manipulations can be pretty costly, I think this is worth fixing. 

This is my first contribution to this project, so I apologize if I've missed anything.

blikblum:
This avoids double call to lowerCase. Fixes #4075

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -14133,7 +14133,7 @@
      */
     var camelCase = createCompounder(function(result, word, index) {
       word = word.toLowerCase();
-      return result + (index ? capitalize(word) : word);
+      return result + (index ? upperFirst(word) : word);
     });
 
     /**


jdalton:
Thank you @blikblum!

PR #4077, opened 2018-11-25
[master] Call upperFirst directly, instead of capitalize, in camelCase (#4075)

blikblum:
Same as #4076 , but for master branch


jdalton:
Thanks @blikblum!

PR #4078, opened 2018-11-25
Call clearTimeout on debounce.flush (#4074)

### LINKED ISSUE #4074: `clearTimeout`  when `debounce.flush()` is called
When you flush a debounce function, the timer is not cleared.

If the timer is active (`timerId !== undefined`), the function `trailingEdge`is call:

https://github.com/lodash/lodash/blob/0843bd46ef805dd03c0c8d804630804f3ba0ca3c/lodash.js#L10378-L10388

Shouldn't be better with this?

``` javascript
function trailingEdge(time) {
  if (timerId !== undefined) {
    clearTimeout(timerId);
    timerId = undefined;
  }
  // Only invoke if we have `lastArgs` which means `func` has been
  // debounced at least once.
  if (trailing && lastArgs) {
    return invokeFunc(time);
  }
  lastArgs = lastThis = undefined;
  return result;
}
```

blikblum:
Fixes #4074 but unlike proposed in the issue, calls clearTimeout outside of trailingEdge to avoid extra conditional check and unnecessary clearTimeout call, i.e., when timer is expired

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -10390,7 +10390,11 @@
       }
 
       function flush() {
-        return timerId === undefined ? result : trailingEdge(now());
+        if (timerId !== undefined) {
+          clearTimeout(timerId);
+          return trailingEdge(now());
+        }
+        return result;
       }
 
       function debounced() {


jdalton:
Thank you again @blikblum!

PR #4079, opened 2018-11-25
Remove support for node v0.10 in arrayLikeKeys

blikblum:
Removes support to node v0.10 in arrayLikeKeys following v5 roadmap


jdalton:
Thank you @blikblum!

PR #4085, opened 2018-11-28
fix: limit the precision when the precision is negative

asilinwei:
Signed-off-by: asilinwei <1021069119@qq.com>
There is a problem about _.floor():
`_.floor(12, Infinity)  // => 12`
But, when the precision is -Infinity:
`_.floor(12, -Infinity) // => NaN`
It is not very suitable because it is only limited the precision when the precision is positive.
It also should be limited the precision when the precision is negative.


jdalton:
Thanks @asilinwei!

PR #4087, opened 2018-11-29
createRound 12x faster using pure arithmetic only

NeekSandhu:
Run the script below in Chrome devtools to test the performance difference

```javascript
function createRoundNew(methodName) {
  const func = Math[methodName]
  return (number, precision) => {
    precision = precision == null ? 0 : (precision >= 0 ? Math.min(precision, 292) : Math.max(precision, -292))
    if (precision) {
      precision = Number.isInteger(precision) ? precision : Math.floor(precision)
      let modifier = 1
      if (precision > 0) {
        for (let i = 0; i < precision; i++) {
          number *= 10
          modifier *= 10
        }
      } else {
        modifier = Math.pow(10, precision)
        number *= modifier
      }

      return func(number) / modifier
    }
    return func(number)
  }
}

function createRoundOld(methodName) {
  const func = Math[methodName]
  return (number, precision) => {
    precision = precision == null ? 0 : (precision >= 0 ? Math.min(precision, 292) : Math.max(precision, -292))
    if (precision) {
      // Shift with exponential notation to avoid floating-point issues.
      // See [MDN](https://mdn.io/round#Examples) for more details.
      let pair = `${number}e`.split('e')
      const value = func(`${pair[0]}e${+pair[1] + precision}`)

      pair = `${value}e`.split('e')
      return +`${pair[0]}e${+pair[1] - precision}`
    }
    return func(number)
  }
}

const roundOld = createRoundOld('round')
const roundNew = createRoundNew('round')


// Setup Sample
var sample = []
for(let i = 0; i < 100000; i++) {
	sample[i] = performance.now() * Math.random()
}

// Test if output of both methods match
for(let i = 0; i < sample.length; i++) {
	console.assert(roundOld(sample[i], 3) === roundNew(sample[i], 3))
}


// Perfomance testing
let t0 = performance.now()
for(let i = 0; i < sample.length; i++) {
	roundOld(sample[i], 3)
}
console.log('Old method took', performance.now() - t0)

t0 = performance.now()
for(let i = 0; i < sample.length; i++) {
	roundNew(sample[i], 3)
}
console.log('New method took', performance.now() - t0)
```

### DIFF
--- a/.internal/createRound.js
+++ b/.internal/createRound.js
@@ -10,13 +10,8 @@ function createRound(methodName) {
   return (number, precision) => {
     precision = precision == null ? 0 : (precision >= 0 ? Math.min(precision, 292) : Math.max(precision, -292))
     if (precision) {
-      // Shift with exponential notation to avoid floating-point issues.
-      // See [MDN](https://mdn.io/round#Examples) for more details.
-      let pair = `${number}e`.split('e')
-      const value = func(`${pair[0]}e${+pair[1] + precision}`)
-
-      pair = `${value}e`.split('e')
-      return +`${pair[0]}e${+pair[1] - precision}`
+	    const modifier = Math.pow(10, precision)
+	    return func(number * modifier) / modifier
     }
     return func(number)
   }
--- a/.internal/createRound.js
+++ b/.internal/createRound.js
@@ -10,8 +10,19 @@ function createRound(methodName) {
   return (number, precision) => {
     precision = precision == null ? 0 : (precision >= 0 ? Math.min(precision, 292) : Math.max(precision, -292))
     if (precision) {
-	    const modifier = Math.pow(10, precision)
-	    return func(number * modifier) / modifier
+      precision = Number.isInteger(precision) ? precision : Math.floor(precision)
+      let modifier = 1
+      if (precision > 0) {
+        for (let i = 0; i < precision; i++) {
+          number *= 10
+          modifier *= 10
+        }
+      } else {
+        modifier = Math.pow(10, precision)
+        number *= modifier
+      }
+
+      return func(number) / modifier
     }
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @NeekSandhu!

Have you run this mod against the unit tests? You can do so using the `4.17.12-pre` branch.

NeekSandhu:
All tests passing for V8!

PS: As a general feedback, it was extremely hard for me to run the tests, `master` branch doesn't have build/test scripts, and `4.17.12-pre` doesn't have "fresh" source files to build from. One unified system in two separate places. Was quite hard IMO

NeekSandhu:
Apologies I couldn't run the tests inside SpiderMonkey (FF) and Chakra (Edge). The repo structure is too complex for me to wrap my head around.

BTW The performance is still faster than the previous implementation.

NeekSandhu:
😲 I did not have any idea how deep this rabbit hole goes!

I guess lodash is use-case agnostic and must cover every use case, even when the probability of encountering such is less than 0.1%.

And just now I realized (visiting the issue @gibson042 linked), that lodash previously used the same logic I'm proposing now. So I guess, this implementation is very limited in how many use-cases it can "safely" cover. So there's that.

Until some miracle happens. I'm retracting the PR 🙂

jdalton:
@NeekSandhu 
> I guess lodash is use-case agnostic and must cover every use case, even when the probability of encountering such is less than 0.1%. [...] So I guess, this implementation is very limited in how many use-cases it can "safely" cover. 

Yep. I was thinking this might be the case. Thank you for creating the PR anyways!

gibson042:
I'd love to see this protected by unit tests, but each implementation appears to have its own edge cases. Any thoughts on how to catch it?

PR #4091, opened 2018-11-30
Remove checks for the Map existence

blikblum:
Removes the checks for Map existence and related fallback code


jdalton:
Thanks @blikblum!

Issue #4094, opened 2018-12-01
Use Map instead of MapCache in memoize


blikblum:
By replacing MapCache by Map a considered amount of code (implementations of MapCache, Hash and ListCache) can be saved for those using only memoize

Even if MapCache is [still faster](https://github.com/lodash/lodash/issues/3117) than Map (may not be true anymore due to [recent](https://v8.dev/blog/hash-code) browser optimizations) probably is not relevant is real world usage. So IMO the trade off is worth.

Alternatively MapCache can be exposed to those that need this specific kind of optimization

jdalton:
Go for it @blikblum!

PR #4095, opened 2018-12-01
Replace MapCache by Map in memoize (#4094)

blikblum:
Replaces MapCache by Map in memoize as stated in #4094

For the record i patched locally v4 and all tests passes


jdalton:
Thank you @blikblum!

PR #4100, opened 2018-12-04
 Re-add keysIn module fixing baseClone and baseMerge (#4099) 

### LINKED ISSUE #4099: baseMerge and baseClone does not works because of missing keysIn
baseMerge and baseClone imports keysIn which does not exists anymore

For the record baseKeysIn was removed [here](https://github.com/lodash/lodash/commit/58e484f389762750a13b0d8f952eb80235a31649)

should baseKeysIn be restored or do the same as [here](https://github.com/lodash/lodash/commit/58e484f389762750a13b0d8f952eb80235a31649#diff-5c9720962ba84f94900532fc93850ba5L12)?

blikblum:
Fixes #4099 

Since keysIn is passed along as a variable, i needed to create a local function instead of using key-in in place

### DIFF
--- a/keysIn.js
+++ b/keysIn.js
@@ -1,0 +1,32 @@
+/**
+ * Creates an array of the own and inherited enumerable property names of `object`.
+ *
+ *
+ * @static
+ * @memberOf _
+ * @since 3.0.0
+ * @category Object
+ * @param {Object} object The object to query.
+ * @returns {Array} Returns the array of property names.
+ * @example
+ *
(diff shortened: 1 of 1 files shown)


jdalton:
Welp, you might as well make it its own module.

blikblum:
A public or an internal one?

jdalton:
Let's go external.

blikblum:
Done

jdalton:
Thank you!

PR #4101, opened 2018-12-04
Fixes compilation of merge and mergeWith

blikblum:
Restored createAssigner since both merge and mergeWidth is basically composing with it


jdalton:
Thanks @blikblum!

PR #4103, opened 2018-12-05
Fix findLast missing dependency

blikblum:
It fixes findLast by inlining the logic of createFind


jdalton:
Thanks again @blikblum!

PR #4105, opened 2018-12-06
Fix baseOrderBy missing dependency

blikblum:
It does so by inlining baseMap


blikblum:
BTW: differently from stated in the docs / comments, only function iteratees are supported:

`const criteria = iteratees.map((iteratee) => iteratee(value))`

jdalton:
👏 👏

PR #4112, opened 2018-12-10
Use baseGetTag instead of getTag where possible

blikblum:
Use baseGetTag instead of getTag where the fallbacks done in getTag are not relevant

This reduces the affected modules size significantly

### DIFF
--- a/.internal/baseIsEqualDeep.js
+++ b/.internal/baseIsEqualDeep.js
@@ -2,7 +2,7 @@ import Stack from './Stack.js'
 import equalArrays from './equalArrays.js'
 import equalByTag from './equalByTag.js'
 import equalObjects from './equalObjects.js'
-import getTag from './getTag.js'
+import baseGetTag from './baseGetTag.js'
 import isBuffer from '../isBuffer.js'
 import isTypedArray from '../isTypedArray.js'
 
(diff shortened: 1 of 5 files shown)


blikblum:
> Thanks @blikblum! (I really appreciate you taking the time to do these PRs).

You're welcome

> 
> 
> As a follow-up we need to remove the `masked` check workarounds from `baseGetTag` as it's causing potential ECMA262 pain for trying to normalize some things.

Can you elaborate more?

I investigated the possibility of removing masked workaround but it supports the case where toStringTag is mutated in an object:

```
const obj = {}
obj[Symbol.toStringTag] = 'test'
toString.call(obj) // -> [object test] instead of [object Object]
```

Complete test at https://codepen.io/blikblum/pen/NePaog

jdalton:
The `baseGetTag` module can be simplified to just:

```js
 function baseGetTag(value) { 
   if (value == null) { 
     return value === undefined ? '[object Undefined]' : '[object Null]' 
   } 
   return toString.call(value) 
 }
```

blikblum:
> The `baseGetTag` module can be simplified to just:

I can do that. But how about the case i mentioned above? 

A warn in documentation about the danger of mutating Symbol.toStringTag would be enough?

jdalton:
I'm fine with it being left untackled. In this case if someone is really wanting their value to be branded as a given type then they want it to be treated as such. A note wouldn't hurt though.

PR #4113, opened 2018-12-11
Remove check for Symbol.toStringTag field mutations in baseGetTag (#4112)

blikblum:
As per discussion in #4112

### DIFF
--- a/.internal/baseGetTag.js
+++ b/.internal/baseGetTag.js
@@ -1,7 +1,4 @@
-const objectProto = Object.prototype
-const hasOwnProperty = objectProto.hasOwnProperty
-const toString = objectProto.toString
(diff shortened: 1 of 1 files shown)


blikblum:
> As a follow up. I'm looking to replace all whateverFunc.call and whateverFunc.apply calls with Reflect.apply calls

I see some (small) drawbacks of such changes:
* Worse minification
* Instantiation of an extra array when replacing call methods

jdalton:
I'm fine with the extra array. The win here is consistent use of the `Reflect` API (basically if it can be done with it we'll be doing it). Translates well to eventual `Proxy` use too. Minification with Gzip will gobble that up (so not as big a deal). Gzip loves repeated code bits.

PR #4115, opened 2018-12-11
Replace getTag implementation by the one from baseGetTag 

blikblum:
It remove the workarounds to Map, Promise etc since they are not necessary anymore. IE11 with core-js correctly supports those tags and node v4 is unmaintained

Here's a [pen](https://codepen.io/blikblum/full/xmbWKN) with the tests that can be loaded with IE11


jdalton:
Very nice! Thank you @blikblum!

PR #4116, opened 2018-12-11
Use native filter instead of custom one where possible

blikblum:
This flattens the dependency tree making the size cost of some modules smaller

BTW: not related to the changes but xorBy and xorWith seems buggy (not tested) since is including the iteratee in the filtered array


jdalton:
This is great @blikblum! Thank you!

Issue #3903, opened 2018-08-08
attempt + default value


sreevisakh:
a function to like attempt which will catch error and return a default value if there is an error. lot of try catch statements can be reduced to single lines

eg:
### without
```js
try{
  value = JSON.parse(str);
}
catch(e){
  //dont care
  value = [];
}
```
### with
```js
value = attempt(JSON.parse.bind(null, str), [])
//or
value = attempt(JSON.parse, [str], [])

```

jdalton:
Hi @sreevisakh!

I've wanted something like that myself though I usually handle it with one off `tryXyz` methods now.
That said I could see this as something like an `_.attemptOr` method.

We'll keep an eye on [the popularity of the request](https://github.com/lodash/lodash/blob/master/.github/CONTRIBUTING.md#feature-requests).

biiiipy:
I'm actually **very** surprised that there is not a simple function for this. I assumed that lodash 100% has this

Tonkonozhenko:
Hi
It was also strange for me that it's not supported by default. Our use case is parsing a string to JSON with fallback to the default

wojtiku:
+1, this would be great to have

kyrylo-soulandwolf:
+1

PR #4123, opened 2018-12-16
Add _.attemptOr method

joselcvarela:
In flavor of #3903
Adds `attemptOr` method. Pretty similar to `attempt`, but instead of returning error in case of throw, returns default value defined as second argument


jdalton:
Hi @joselcvarela!

Thanks for creating the PR. If/when we tackle #3903 we'll use this PR as the basis.

PR #4140, opened 2019-01-07
sortBy: Only evaluate iteratees when needed

### LINKED ISSUE #4067: `sortBy` evaluates all iteratees even on non-ties
Upon evaluating

    lodash.sortBy([2, 1], (x) => x, (x) => void console.log(x))

I was a bit surprised to see that two log statements are in fact printed
to the console, even though there are no ties on the first iteratee and
so the second iteratee need never be evaluated.

Upon looking at the implementation, it becomes clearer why this is the
case: you precompute all the `criteria` for each element so that they
only have to be computed once. But this makes use cases in which there
is a potentially expensive “fallback iteratee”, like

    sortBy(
        objects,
        (x) => x.id,  // almost all objects have distinct IDs
        (x) => JSON.stringify(x)  // expensive, but should happen only rarely!
    );

much more costly.

(All this holds the same for `orderBy`, of course.)

One could imagine an implementation that computes each criterion lazily,
only as it is needed to break a tie. From reading the docs, this would
appear to be a backward-compatible change; you don’t currently specify
under which circumstances the iteratees are invoked (which is good!).

Would you be interested in making or accepting such a change?

wchargin:
Summary:
Resolves #4067. Performance-wise, this incurs some extra administrative
overhead on all invocations. In exchange, we can see order-of-magnitude
improvements in cases where there are expensive “fallback iteratees”
that are rarely hit (e.g., a `JSON.stringify`).

Running the `sortBy` benchmarks included in `perf/perf.js` gives the
following results on my laptop (Intel i5-4300U with Chrome 67 on Linux):

  - “`_.sortBy` with `callback`”:
      - before this patch: 59,249 ops/sec ±2.65% (56 runs sampled)
      - after this patch: 54,504 ops/sec ±2.91% (56 runs sampled)
      - (new / old) ratio: 0.92 ± 0.05 (0.87–0.97)

  - “`_.sortBy` with `property` name”:
      - before this patch: 98,956 ops/sec ±2.37% (58 runs sampled)
      - after this patch: 93,793 ops/sec ±1.46% (58 runs sampled)
      - (new / old) ratio: 0.94 ± 0.04 (0.91–0.98)

…and these on a workstation (Intel W-2135 with Chrome 71 on Linux):

  - “`_.sortBy` with `callback`”:
      - before this patch: 66,071 ops/sec ±2.39% (61 runs sampled)
      - after this patch: 58,707 ops/sec ±2.82% (58 runs sampled)
      - (new / old) ratio: 0.89 ± 0.05 (0.84–0.94)

  - “`_.sortBy` with `property` name”:
      - before this patch: 132,945 ops/sec ±2.71% (56 runs sampled)
      - after this patch: 129,162 ops/sec ±3.30% (61 runs sampled)
      - (new / old) ratio: 0.97 ± 0.06 (0.92–1.03)

(Best-of-3 timing used in each case.)

We can see that there is usually a statistically significant _decrease_
in the overall performance in these existing benchmarks.

This commit does not add a new benchmark to observe the best-case
behavior because the Underscore.js API does not support multiple
iteratees in `sortBy`, which would be needed for a fair comparison.

Test Plan:
The newly added test cases fail before this commit, but pass after it.
All existing tests pass (`npm run validate` passes).

wchargin-branch: sortBy-lazy

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -4669,7 +4669,8 @@
      */
     function forceCriterion(criterion) {
       if (!criterion.computed) {
-        criterion.value = criterion.iteratee(criterion.element);
+        const iteratee = criterion.iteratee;
+        criterion.value = iteratee(criterion.element);
         criterion.computed = true;
       }
       return criterion.value;
--- a/lodash.js
+++ b/lodash.js
@@ -3724,7 +3724,7 @@
 
       var result = baseMap(collection, function(value, key, collection) {
         var criteria = arrayMap(iteratees, function(iteratee) {
-          return iteratee(value);
+          return { 'computed': false, 'iteratee': iteratee, 'element': value };
         });
         return { 'criteria': criteria, 'index': ++index, 'value': value };
       });
@@ -4637,7 +4637,9 @@
           ordersLength = orders.length;
 
       while (++index < length) {
-        var result = compareAscending(objCriteria[index], othCriteria[index]);
+        var result = compareAscending(
+            forceCriterion(objCriteria[index]),
+            forceCriterion(othCriteria[index]));
         if (result) {
           if (index >= ordersLength) {
             return result;
@@ -4654,6 +4656,23 @@
       // This also ensures a stable sort in V8 and other engines.
       // See https://bugs.chromium.org/p/v8/issues/detail?id=90 for more details.
       return object.index - other.index;
+    }
+
+    /**
+     * Used by `compareMultiple` to evaluate a criterion, either reusing its
+     * computed result or computing it for the first time and then caching the
+     * result.
+     *
+     * @private
+     * @param {Object} criterion The criterion to evaluate
+     * @return {*} The comparison key
+     */
+    function forceCriterion(criterion) {
+      if (!criterion.computed) {
+        criterion.value = criterion.iteratee(criterion.element);
+        criterion.computed = true;
+      }
+      return criterion.value;
     }
 
     /**
--- a/test/test.js
+++ b/test/test.js
@@ -20650,6 +20650,16 @@
       assert.deepEqual(actual, [1, 2, 3, 4, undefined]);
     });
 
+    QUnit.test('should invoke iteratees with unbound `this`', function(assert) {
+      assert.expect(1);
+      var input = [1, 2];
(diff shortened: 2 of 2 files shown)


wchargin:
@jdalton: Performance details in the description. There’s a reasonable
argument for not merging this or for making the behavior optional: it’s
up to you.

lodash.js
```diff
@@ -4656,6 +4658,23 @@
...
+     * computed result or computing it for the first time and then caching the
+     * result.
+     *
+     * @private
+     * @param {Object} criterion The criterion to evaluate
+     * @return {*} The comparison key
+     */
+    function forceCriterion(criterion) {
+      if (!criterion.computed) {
+        criterion.value = criterion.iteratee(criterion.element);
```
jdalton: Can you pluck the `criterion.iteratee` so that it's `this` is not associated with `criterion`:
```js
const { iteratee } = criterion

criterion.value = iteratee(criterion.element)
```

lodash.js
```diff
@@ -3724,7 +3724,7 @@
 
       var result = baseMap(collection, function(value, key, collection) {
         var criteria = arrayMap(iteratees, function(iteratee) {
-          return iteratee(value);
+          return { 'computed': false, 'iteratee': iteratee, 'element': value };
```
jdalton: You can use this syntax:
```js
{ 'computed': false, iteratee, 'element': value }

jdalton: It's ok, we'll switch to `terser` instead if `uglify`.

wchargin: I’m not sure what you’re asking—would you like me to switch this to use
the punning syntax, such that `npm test` fails at this commit, or should
I leave it as is? (Or wait for an update to the base branch 4.17.12-pre
to change the parser?)

wchargin:
Curious why this was closed?

jdalton:
Not sure. It wasn't an explicit close. Might be related to some branch activity going on. Fee free to resubmit the PR. The GitHub UI is not giving me the ability to Reopen it.

jdalton:
Ah because I removed the `4.17.12-pre ` branch.

PR #4145, opened 2019-01-11
Change type check and the way of sending params.

moonformeli:
1. Changed the line 56, to get the higher performance.
2. Changed the line 64-65 to the way of ES6. Since this file is using the features from ES6, line 64-65 seems to be changed as well.


jdalton:
Thank you @moonformeli!

PR #4148, opened 2019-01-11
Use the exported functions

moonformeli:
Imported some functions that do the exactly the same things as the original codes.

* line 32 - It seems ```object``` must be an object or the error will be printed out. (According to the description of this function) For the better performance, check if ```object``` is an object at the very beginning of the codes.

* line 35 - The original code in ```while``` was below
```javascript
/* ... */
object = object[key]
```
Because of this part, ```object``` was not immutable, so ```cloneDeep``` was needed to create a 
new object for the immutability and the safety of the original variable.

* line 45, 53 - Replaced with the existing exported functions which do the same thing.

----

Further question
* Since ```isIndex()``` has ```!!length``` in its return statement, isn't ```!!length``` unnecessary in ```hasPathIn.js``` file's return statement? It seems redundant.

### DIFF
--- a/hasPathIn.js
+++ b/hasPathIn.js
@@ -3,6 +3,10 @@ import isArguments from './isArguments.js'
 import isIndex from './.internal/isIndex.js'
 import isLength from './isLength.js'
 import toKey from './.internal/toKey.js'
+import cloneDeep from './cloneDeep'
+import hasIn from './hasIn.js'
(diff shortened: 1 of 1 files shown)


jdalton:
Thanks @moonformeli!

I don't mind this minor bit of duplication and want to avoid `cloneDeep` use here.

PR #4149, opened 2019-01-12
Suggesting addition of new API, 'keyMany'. 

iiian:
Like 'keyBy', except instead of keying on a 'by-last' basis, keys to a collection where each element advertises as being in that key group. Iteratee takes an element and optional key, returns a string array of which keys it should be grouped into.


jdalton:
Thanks @iiian!

PR #4151, opened 2019-01-15
added toBoolean method

ylcnfrht:
Hi, I have added toBoolean method. that methot returns string boolean to boolean for example;
'true' -->    true
'True' -->    true
'TRUE' -->  true
'tRuE' -->   true
'false' -->   false
'False' -->   false
'FALSE' --> false
'fAlsE' -->  false
other case --> false
Can you check it?

### DIFF
--- a/toBoolean.js
+++ b/toBoolean.js
@@ -1,0 +1,50 @@
+import isString from './isString.js'
+import toLower from './toLower.js'
+
+/**
+ * Returns string boolean to boolean
+ *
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @ylcnfrht!

Thank you for taking the time to create this PR.
I think this kind of addition is better suited as a [`_.mixin`](http://lodash.com/docs/#mixin).

PR #4152, opened 2019-01-15
added toBoolean method again

ylcnfrht:
Hi @jdalton
Firstly I would like to ask why did you close my PR without my answer?
Secondly, I think you don't understand me. I mean I want to return a string to boolean like below;
'true' to true or 'false' to false
And you mentioned about _.mixin but I can't see a useful example for my case.
Can you give me an example if a method will work for my case ('true' to true or 'false' to false)?
Thank you again.

### DIFF
--- a/toBoolean.js
+++ b/toBoolean.js
@@ -1,0 +1,50 @@
+import isString from './isString.js'
+import toLower from './toLower.js'
+
+/**
+ * Returns string boolean to boolean
+ *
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @ylcnfrht!

Thanks for the PR but I will not be merging your request in. Instead we hold them for users to react with :+1:.

PR #4153, opened 2019-01-16
Use 3 equal signs

moonformeli:
Like other if statements, use 3 equal signs


jdalton:
Thank you @moonformeli!

PR #4154, opened 2019-01-16
Changed to arrow function

moonformeli:
MDN Link - [Array.isArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray)

As long as the function only wants to check if the type of param is either ```null``` or ```undefined``` or ```Array```, I think this arrow function will work well too

### DIFF
--- a/initial.js
+++ b/initial.js
@@ -12,9 +12,6 @@ import slice from './slice.js'
  * initial([1, 2, 3])
  * // => [1, 2]
  */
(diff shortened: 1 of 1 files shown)


jdalton:
Thanks for the PRs @moonformeli!

In this case I prefer the more readable, but longer syntax form, that's in place.

PR #4156, opened 2019-01-17
Fix _.isFunction for async generator functions

lukas-reineke:
async generator function have a string representation that `isFunction` did not check against.

While adding the test for this, I also noticed that all the tests for `isFunction` only tested if `Function('string of function to test')` is a function, not the actual function that was supposed to be tested.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -92,6 +92,7 @@
   var argsTag = '[object Arguments]',
       arrayTag = '[object Array]',
       asyncTag = '[object AsyncFunction]',
+      asyncGenTag = '[object AsyncGeneratorFunction]',
       boolTag = '[object Boolean]',
       dateTag = '[object Date]',
       domExcTag = '[object DOMException]',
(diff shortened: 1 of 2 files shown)


jdalton:
Thanks @lukas-reineke! I'd be fine removing `isFunction` entirely and deferring to `typeof` now. Browsers no longer mixup `typeof`.

a0viedo:
hi there @jdalton, could you elaborate on what you meant by browsers mixing up `typeof`? trying to get some historical context...`typeof` seems cleaner and easier to grasp

jdalton:
Sure @a0viedo!

There used to be problems with [older Safari versions](https://github.com/lodash/lodash/blob/3.10.1/lodash.js#L8711-L8713) and [IE11](https://github.com/lodash/lodash/blob/3.10.1/lodash.js#L366-L367).

ewojjowe:
Nooice! looks good.

jdalton:
I'm closing this in favor of a PR that simply removes `isFunction`.

PR #4163, opened 2019-01-20
avoid extensionless imports

43081j:
This means lodash will work out of the box for things like [deno](https://github.com/denoland/deno).


jdalton:
Thank you @43081j!

PR #4172, opened 2019-01-27
Add initial test files from lodash v4.

bnjmnt4n:
These files were created using a simplistic AST manipulator using `recast` to preserve formatting. There's bound to be a huge chunk of errors, but this serves as a good start. QUnit was replaced with Mocha, with ES2015 imports running via `esm`.

As far as possible, QUnit-specific syntax has been replaced with Mocha's `describe` and `it`, while the native Node.js `assert` module is used for assertions. Files in the `test` directory ending in `.test.js` will be treated as test files.

I've included all tests, including those for methods removed in the upcoming v5, for posterity's sake. Feel free to cherry-pick only relevant files. Functions to be tested are imported unless they were run using computed `lodash[fn]` style syntax, in which case they require updating. The files also conform to the old code style, which is drastically different from the current semicolon-free, ES2015-friendly code, so maybe a run of prettier/eslint/jscodeshift/*insert popular code formatter here* could help in the cleanup. Also, loads of legacy constructs like `isNpm` and `isModularize` can slowly be removed when porting


bnjmnt4n:
Seems like the formatting's a little wonky: possibly I forgot to set the tab width option on recast.🤦

jdalton:
Awesome @bnjmnt4n!

Thank you so much! If you need time to kick the tires more on recast let me know and I'll hold up for possible rebases to the PR to fix whatever formatting or nits you have.

bnjmnt4n:
Any feedback on any part of the generated code? Should be easy to fit in some general AST transformations if there's anything else. (Should I convert the `describe` and `it` function expressions to arrow functions?)

Is the `lodashStable` usage something that will continue being used? I'm probably going to change `import { lodashStable } from "./utils.js"` to `import lodashStable from "lodash'`, as well as transform the `assert` commands to ensure compatibility with Node.js. Probably going to look into removing some of the legacy `if (!isStrict/!isNpm/...) {` calls to inline the code as well.

falsyvalues:
I think it would be nice to split tests and keep them per corresponding source code (easier to navigate and maintain) and create things like `Strings.category.methods.js`.

bnjmnt4n:
> keep them per corresponding source code

Could you clarify what you mean by this? The generated tests were split into individual files based on the test names passed into `QUnit.module` — which can be a little naive, especially for tests where multiple (similar) methods are tested at once. Files such as the aforementioned `String.category.methods.js` is one prime example.

Of course, these generated files are just a starting point to cleanup the tests, so hopefully we can build a robust test system based on these (crudely put together) files.

bnjmnt4n:
> `lowerFirst.js` will have corresponding `lowerFirst.test.js` file.

Ah, that was the initial plan after the generated test files were cleaned up.

@jdalton I've pushed a new set of commits with some fixes - unless there's some major issue with the generated code - feel free to chop things up and fix the tests.

jdalton:
This is so nice @bnjmnt4n!

I noticed test files are `foo.js` and others are `foo.test.js`. What's the significance of the `.js` vs `.test.js` there? Is it an oversight?

jdalton:
Ah okay!

PR #4177, opened 2019-02-01
Implement `is4byte` method

toshisugimo:
Hi!

I implemented the `is4byte` method.


jdalton:
Hi @toshisugimo!

This seems a bit too specialized for Lodash core. However, it'd make a good [`_.mixin`](http://lodash.com/docs/#mixin).

PR #4182, opened 2019-02-10
added check for NaN

saurabhg22:


jdalton:
Hi @saurabhg22!

The current behavior is by design.

StreetStrider:
@saurabhg22 `NaN` is a number. You probably look for `isFinite`.

saurabhg22:
Thanks, but still don't understand how "Not A Number" is actually a number.

StreetStrider:
@saurabhg22 it's a special value, however it belongs to number domain. Just like `Infinity` and `-Infinity`. This is related to a standart on top of which JS' number is founded (look for IEEE 754).

PR #4186, opened 2019-02-11
Update defer.js. Change timeout at zero

Milesq:
With zero it also works


jdalton:
Hi @Milesq!

The timeout is `1` because historically some browsers had problems handling `0`. I'm okay keeping it at `1` for now.

PR #4189, opened 2019-02-12
Updating due to security vulnerability

Alec321:
Version 4.17.10 was compromised if I recall. Should force users to use the version that is not compromised.


Alec321:
Version bump

jdalton:
Closed by https://github.com/lodash/lodash/commit/7606ea3e25969076e77e237effda84f78ae533ca.

PR #4192, opened 2019-02-12
Create spyOn.js

Milesq:
Function that allows you to spy on another function and control what to return


StreetStrider:
Why not use Sinon for spies?

jdalton:
Hi @Milesq!

I'd like to defer to established spy/mock libs like [Sinon](https://sinonjs.org/) for this.

Milesq:
Of course you can use any library for unit tests. But 1) these libraries are quite scary and it is not profitable to link the whole library (especially since these libraries have a different purpose) 2) in Lodash a small function would be useful, for example to add hooks

PR #4199, opened 2019-02-18
Remove --save option as it isn't required anymore

stevemoser:
"As of npm 5.0.0, installed modules are added as a dependency by default, so the --save option is no longer needed. The other save options still exist and are listed in the documentation for npm install."

https://stackoverflow.com/a/19578808/142358

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ In a browser:
 Using npm:
 ```shell
 $ npm i -g npm
-$ npm i --save lodash
+$ npm i lodash
 ```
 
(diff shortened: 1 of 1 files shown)


MV88:
Hi, i was thinking that not everybody are using the latest npm. so i would add also a text like this: "add --save if you are using npm < 5.0.0".  what do you think about it?

stevemoser:
@MV88 Agreed and done.

jdalton:
Thank you @stevemoser!

PR #4203, opened 2019-02-20
Adding pickTruthyValues function to lodash library

tameemsafi:
# Description

This function will take an object and remove any empty/falsy values from it. It will also do a deep search of nested objects and arrays.


shengslogar:
Would prefer to see something along the lines of `pickByDeep` and `omitByDeep`. This method feels too rigid.

tameemsafi:
I created my own package on npm and re-wrote this function is typescript.

https://github.com/tameemsafi/pick-truthy-values

wmertens:
The tests are missing the `0` value

PR #4207, opened 2019-02-22
_.result returns default value when property is null

MateusGabi:
I'm proposing a little difference between the actual state. Today, the _.result function returns:
> If the resolved value is undefined, the default value is returned in its place.

My proposal is:
> If the resolved value is undefined or null, the default value is returned in its place.

### Example

```
var nullInnerObject = { any: null }

 assert.deepEqual(result(nullInnerObject, 'any', { foo: 'bizz' }), {
  foo: 'bizz'
});
```


jdalton:
Hi @MateusGabi!

I've found cases where `null` should be counted as a non-value but in those cases I think deferring to a one-off helper is preferred.

Issue #4228, opened 2019-03-08
use +value to convert binary/octal/hexadecimal string


gu-xionghong:
https://github.com/lodash/lodash/blob/e7221b96ec2073e578e948966f0fce8815815c36/toNumber.js#L58-L62
We can use `+value` to achieve this part:
```js
return +value
```

jdalton:
Hi @gu-xionghong!

Are you sure? Can you validate it against our tests?

gu-xionghong:
Yes, I validated it and it passed.  
You can check the [ecma-262 - ToNumber Applied to the String Type](https://www.ecma-international.org/ecma-262/7.0/index.html#sec-tonumber-applied-to-the-string-type) to know why it can work.
![image](https://user-images.githubusercontent.com/12096816/54022169-d984be80-41cc-11e9-9907-d5be0694f349.png)
![image](https://user-images.githubusercontent.com/12096816/54022184-e1446300-41cc-11e9-86d6-e01d3f340db2.png)

gu-xionghong:
Another question, how you guys run the unit tests? I need to create a new mocha-project to run the test. Thx for your answer.

jdalton:
Moved to #4230.

PR #4230, opened 2019-03-11
Simplifies toNumber and add tests

gu-xionghong:
Using `+value` to convert binary/octal/hexadecimal string and add some tests.
Issue https://github.com/lodash/lodash/issues/4228


jdalton:
Thank you @gu-xionghong!

PR #4233, opened 2019-03-12
Fix _.maxBy and _.minBy

gu-xionghong:
In the `master`, the  `_.maxBy` and `_.minBy` will work like this:
```js
const objects = [{ n: 1 }, { n: 3 }, { n: 2 }];

_.maxBy(objects, ({n}) => n)
// expect {n: 3}, but return {n: 2}

_.minBy(objects, ({n}) => n)
// expect {n: 1}, but return {n: 2}
```

### DIFF
--- a/maxBy.js
+++ b/maxBy.js
@@ -22,8 +22,8 @@ function maxBy(array, iteratee) {
   if (array == null) {
     return result
   }
+  let computed
   for (const value of array) {
-    let computed
     const current = iteratee(value)
 
     if (current != null && (computed === undefined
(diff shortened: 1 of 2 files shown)


jdalton:
Thanks @gu-xionghong!

falsyvalues:
Does tests pass before this change? Maybe we should add more if so.

gu-xionghong:
Hi, @falsyvalues , I found that after this [commit](https://github.com/lodash/lodash/commit/65654f8f9e4313b57b341c88017456950dd2d8ee#diff-09ff95768a2b402d1bad2463bc3257a8L30)  which removes the `baseIteratee`, the `master` branch can't be passed the tests. But the other branches always keep the different version, and they can be passed the tests.

PR #4234, opened 2019-03-12
Adding `compress` Method

smusali:
Among the other `_` methods, I have been looking for something like this but the closest one is the `_.compact` which is not exactly same. What I think is most developers will benefit from this new contribution.

Thanks in advance!

### DIFF
--- a/compress.js
+++ b/compress.js
@@ -1,0 +1,42 @@
+import forEach from './forEach.js'
+import isNull from './isNull.js'
+import isUndefined from './isUndefined.js'
+import isEmpty from './isEmpty.js'
+import isString from './isString.js'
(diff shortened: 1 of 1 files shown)


smusali:
@jdalton, I think, it's worth thinking about adding this into the list of methods but I'm not sure whether I prepared it correctly or not - should I need `test` as well? Thanks in advance!

jdalton:
Hi @smusali!

Thanks for opening this PR! I think this kind of method would be better as a standalone package or a [mixin](http://lodash.com/docs/#mixin).

smusali:
@jdalton, what do you mean by `a standalone package`? Can you clarify? Thanks in advance!

PR #4243, opened 2019-03-22
mergeWith: stack passed to customizer should always be defined

mhassan1:
Summary:
If the first values encountered in the `object` in mergeWith are not objects, `stack` is undefined when passed to the `customizer`. Once the first object-ish value is encountered, `stack` gets initialized, and all further calls to `customizer` include a defined `stack`. This PR makes `stack` always defined, even before the first object-ish value is encountered.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -3588,8 +3588,8 @@
         return;
       }
       baseFor(source, function(srcValue, key) {
+        stack || (stack = new Stack);
         if (isObject(srcValue)) {
-          stack || (stack = new Stack);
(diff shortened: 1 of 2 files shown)


jdalton:
Thank you @mhassan1!

Once you complete the CLA we should be good to go.

mhassan1:
I force-pushed with the right email address, but jsf-clabot didn't recognize that I signed the CLA. I've created a new PR: https://github.com/lodash/lodash/pull/4244

PR #4244, opened 2019-03-22
mergeWith: stack passed to customizer should always be defined

mhassan1:
Summary:
If the first values encountered in the `object` in mergeWith are not objects, `stack` is undefined when passed to the `customizer`. Once the first object-ish value is encountered, `stack` gets initialized, and all further calls to `customizer` include a defined `stack`. This PR makes `stack` always defined, even before the first object-ish value is encountered.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -3588,8 +3588,8 @@
         return;
       }
       baseFor(source, function(srcValue, key) {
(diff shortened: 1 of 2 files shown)


jdalton:
Thank you @mhassan1!

PR #4259, opened 2019-04-08
Fix findLast missing dependency

gu-xionghong:
Add `isArrayLike` dependency  in `findLast`.


jdalton:
Thank you @gu-xionghong!

PR #4261, opened 2019-04-08
Update README.md / Added collection with Bit

jonisar:
Shared Lodash with Bit (https://github.com/teambit/bit) so now there's a component collection for the library (https://bit.dev/lodash/lodash) - took about 10 minutes :).

There are several added benefits for this:
1. People can now visually explore the collection and play with each function online (example: https://bit.dev/lodash/lodash/is-match)
2. They can install each function as a package with NPM and Yarn (I know you already did that by refactoring, publishing each package etc)
3. People can use Bit to develop each component right from their own project and best fit it to their needs.

I think this adds to the engaging modular expirience of people working with Lodash. I'd be very happy to hand over ownership over the organization + components and help update it if you like. I hope you find it useful!

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,9 @@
 [Wiki](https://github.com/lodash/lodash/wiki "Changelog, Roadmap, etc.") |
 [Code of Conduct](https://js.foundation/community/code-of-conduct) |
 [Twitter](https://twitter.com/bestiejs) |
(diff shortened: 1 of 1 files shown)


jdalton:
Hi @jonisar!

Thanks for the PR but I'm not looking to add this at the moment.
Our docs have interactive REPLs built in.

PR #4267, opened 2019-04-15
Add optional resolveEmpty parameter to defaultTo

huntie:
This is an enhancement to Lodash that extends `defaultTo` with an optional `resolveEmpty` parameter. If supplied, `resolveEmpty` will be used to determine whether the passed `value` is empty and that `defaultValue` should be returned.

This would allow the following usage examples, amongst many others:

1. Default to a value when given a value considered empty by `lodash.isEmpty`:

```js
defaultTo('', defaultValue, isEmpty);
```

2. Default to a value after running `lodash.attempt` on a statement that may return an error using `lodash.isError`.

```js
defaultTo(attempt(require, filePath), defaultValue, isError);
```

### DIFF
--- a/defaultTo.js
+++ b/defaultTo.js
@@ -1,12 +1,17 @@
+import isNaN from './isNaN.js'
+import isNil from './isNil.js'
+import overSome from './overSome.js'
+
 /**
  * Checks `value` to determine whether a default value should be returned in
  * its place. The `defaultValue` is returned if `value` is `NaN`, `null`,
- * or `undefined`.
+ * or `undefined`, or if `resolveEmpty` returns `true`.
  *
  * @since 4.14.0
  * @category Util
  * @param {*} value The value to check.
  * @param {*} defaultValue The default value.
+ * @param {Function} resolveEmpty The function to determine if the value is empty.
  * @returns {*} Returns the resolved value.
  * @example
  *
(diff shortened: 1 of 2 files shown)


defaultTo.js
```diff
@@ -15,9 +20,15 @@
...
+ *
+ * defaultTo('', 'default', isEmpty)
+ * // => 'default'
+ *
+ * defaultTo(new Error(), 'default', isError)
+ * // => 'default'
  */
-function defaultTo(value, defaultValue) {
-  return (value == null || value !== value) ? defaultValue : value
+function defaultTo(value, defaultValue, resolveEmpty = overSome([isNaN, isNil])) {
```
huntie: As the [first example in the PR description](https://github.com/lodash/lodash/pull/4267#issue-270557882) illustrates, the main note I have on this implementation is that `resolveEmpty` is not equivalent to `lodash.isEmpty` or indeed any single existing Lodash function, and this was the existing behaviour of `defaultTo` beforehand.

Altering this of course would be breaking, so the default value for `resolveEmpty` behaves identically. I believe `overSome([isNaN, isNil])` is the most succinct way to redefine the `(value == null || value !== value)` test by using tested Lodash functions.

danijmoss:
Sounds good! Thanks!

jdalton:

Issue #4266, opened 2019-04-14
_.round(Infinity, 2) returns NaN


jefffriesen:
```
_.round(Infinity) // Infinity
_.round(Infinity, 2) // NaN
```
Logically, it doesn't makes sense to round Infinity and especially round it to 2 decimal places. However, this is not a contrived example. The value being passed into _.round() is often a finite number where I need to round to 2 decimal places. Occasionally, the value is Infinity, which is a valid value in my case and should return Infinity. However, it returns NaN. 

I can put guards in my code to not round if the incoming value is Infinity, but 
a) It's not ideal
b) There is the inconsistency with `_.round(Infinity)` which works as I would expect

Is this an error or an edge case that needs to be guarded against? Thanks

jdalton:
Hi @jefffriesen!

It's an edge case we didn't consider. I'm fine with making it consistent though. If you're up for a PR against `master` or `4.17.12-pre` branch that would be great!

jefffriesen:
Ok, sounds good. I have a deadline right now, but I can try to put a PR together in a few days if that's soon enough (I know you are amazingly diligent about closing tickets).

jdalton:
That'd be great. Thank you @jefffriesen!

jefffriesen:
I see there are already a couple PRs for this. Great. I just want to also note that _.ceil and _.floor have this same problem

jdalton:
Feel free to create your own PR @jefffriesen : )

jdalton:
Closed by #4272.

PR #4272, opened 2019-04-17
Round Infinity with a precision argument returns Infinity

jefffriesen:
This fixes https://github.com/lodash/lodash/issues/4266

Tested with Infinity, -Infinity, _.round(), _.floor() and _.ceil().

This is a second PR for the same issue. I'm closing the previous PR because I messed up somehow and there were other commits (not mine) in there.

This was branched off of 4.17.12-pre.

Happy to updates with any improvements needed.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -5406,7 +5406,7 @@
       return function(number, precision) {
         number = toNumber(number);
         precision = precision == null ? 0 : nativeMin(toInteger(precision), 292);
-        if (precision && Math.abs(number) !== Infinity) {
+        if (precision && Number.isFinite(number)) {
           // Shift with exponential notation to avoid floating-point issues.
           // See [MDN](https://mdn.io/round#Examples) for more details.
           var pair = (toString(number) + 'e').split('e'),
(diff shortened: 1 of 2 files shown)


lodash.js
```diff
@@ -5406,7 +5406,7 @@
       return function(number, precision) {
         number = toNumber(number);
         precision = precision == null ? 0 : nativeMin(toInteger(precision), 292);
-        if (precision) {
+        if (precision && Math.abs(number) !== Infinity) {
```
jdalton: What about checking if `precision and Number.isFinite(number)`

jefffriesen:
Yep. I thought about that after I submitted it. I wasn’t sure if it’s better to use the lodash version of isFinite or the native one. But you answered that. I’ll fix it in the morning.

jdalton:
Thank you @jefffriesen!

jdalton:
Could you add a test for `NaN` too (since it's not finite)

jefffriesen:
Good suggestion, thanks. I just added them. Let me know if there are more tests I should add.

jdalton:
Thank you @jefffriesen!

jefffriesen:
Thank you @jdalton  for your guidance and patience : )

PR #4279, opened 2019-04-23
Added function checking existence of a subsequence within a parent string. #4280

Cazaimi:
Also added tests.


jdalton:
@Cazaimi Thank you for creating the PR. If/when #4280 is taken on we'll use this PR as the initial commit implementation.

PR #4290, opened 2019-05-08
realNames failing to check .hasOwnProperty()

ekkis:
this patch fixes a problem that breaks lodash in the presence of object prototype functions.  the issue arises from lodash's failure to check the .hasOwnProperty() on keys within the realNames object, which pulls up these methods when it shouldn't

WARNING: there may be other places in the code failing this but so far this is the only instance I've discovered

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -17020,9 +17020,10 @@
     baseForOwn(LazyWrapper.prototype, function(func, methodName) {
       var lodashFunc = lodash[methodName];
       if (lodashFunc) {
-        var key = (lodashFunc.name + ''),
-            names = realNames[key] || (realNames[key] = []);
+        var key = lodashFunc.name + '';
+        if (!hasOwnProperty.call(realNames, key)) return;
 
+        var names = realNames[key] || (realNames[key] = []);
         names.push({ 'name': methodName, 'func': lodashFunc });
       }
     });


lodash.js
```diff
@@ -17020,9 +17020,10 @@
     baseForOwn(LazyWrapper.prototype, function(func, methodName) {
       var lodashFunc = lodash[methodName];
       if (lodashFunc) {
-        var key = (lodashFunc.name + ''),
-            names = realNames[key] || (realNames[key] = []);
+		var key = lodashFunc.name + '';
+		if (!realNames.hasOwnProperty(key)) return;
```
jdalton: ☝️ indentation nits. Also you can use the local `hasOwnProperty.call(realNames, key)`

ekkis: do I need to resubmit the PR?

jdalton: Nope. You can make changes `git add` the files then `git commit --amend` then `git push origin +realNames-ownProperty` (force push an update)

ekkis: done

jdalton:
Thank you @ekkis!

ekkis:
I need to produce a PR for Babel, which depends on lodash, to upgrade to the version that will contain this patch.  I see that my patch has been merged into the -pre.  any idea when this may happen, or way for me to be notified when this is merged onto master?

PR #4292, opened 2019-05-14
Add type: module to allow loading in nodejs with --experimental-modules

### LINKED ISSUE #4289: In loadash-es, package.json should include "type":"module"
In nodejs 12 with the "--experimental-modules", es (es6/esm) modules can be loaded.

But for that to work, for the file to be considered by nodejs as a esm module, by either end in .mjs or
because its package.json must specify "type: module".

Since lodash-es is specifically an es module, I suggest that its package.json contain "type: module". So that it can be loaded in nodejs also.

That should have no other side effects.

Documentation for nodejs here: https://nodejs.org/dist/latest-v12.x/docs/api/esm.html#esm_code_package_json_code_code_type_code_field

Why do I want that?

Well, I developp some es modules that need to work on both the server and the client, without any "bundling". So my module use lodash-es. Today, it can not be loaded in nodejs because nodejs consider it as a cjs (commonJS) module. This should do the trick.

jehon:
Fix #4289

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,5 +16,6 @@
     "John-David Dalton <john.david.dalton@gmail.com>",
     "Mathias Bynens <mathias@qiwi.be>"
   ],
+  "type": "module",
   "scripts": { "test": "echo \"See https://travis-ci.org/lodash-archive/lodash-cli for testing details.\"" }
 }


jdalton:
Thank you @jehon!

blikblum:
AFAICS this will break anywhere that relies on the check for global `module` like [isBuffer](https://github.com/lodash/lodash/blob/es/isBuffer.js#L14) since is not defined when using ES modules

BTW: currently, is already broken with bundlers that supports ES modules like rollup

damianobarbati:
@jdalton do you have an idea when this will make into release?

phistuck:
See https://github.com/lodash/lodash/issues/4359 - this is breaking `npm install` with some NPM versions.

jdalton:
@jehon I had to revert this for now due to ecosystem breakage.

jehon:
@jdalton Yeah, I see. I think the "node module team" didn't see that coming neither :-) Will need to be fixed there finally :-)

PR #4328, opened 2019-06-11
Add a loose condition to get

reesmanp:
This solves the need for doing:

```
get(object, path, defaultValue) || {}
```

### DIFF
--- a/get.js
+++ b/get.js
@@ -9,11 +9,12 @@ import baseGet from './.internal/baseGet.js'
  * @param {Object} object The object to query.
  * @param {Array|string} path The path of the property to get.
  * @param {*} [defaultValue] The value returned for `undefined` resolved values.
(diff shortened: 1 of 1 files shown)


falsyvalues:
Thanks @reesmanp!

We'll keep an eye on [the popularity of the request](https://github.com/lodash/lodash/blob/master/.github/CONTRIBUTING.md#feature-requests), but this change doesn't cover all values that coerce to `false.`

reesmanp:
@falsyvalues thank you for that link, I will put the request up there.
What do you mean by not converting all falsy values? This turns `undefined`, `false`, `null`, and `""` into false. Although I guess a better solution would be checking `typeof` instead, in case an object is where an array should be.

falsyvalues:
@reesmanp Often values `!= null` are considered as valid ones. Personally I think its better idea to pass `function` that will allow full customisation.

PR #4334, opened 2019-06-18
Fix typo in merge test

tkrugg:
It seems like `merge` does convert strings to arrays when merging arrays of `source`. 
But the test description implies currently the opposite. It's probably a typo and this PR fixes it.

```js
_.merge({ a: 'abcde' }, {a: []})
// returns {a: []} 	
```


jdalton:
Thanks @tkrugg!

PR #4342, opened 2019-06-26
fix reTypedTag error

lvzhenbang:
/^\[object (?:Float(?:32|64)|(?:Int|Uint)(?:8|16|32)|Uint8Clamped)Array\]$/ or / (?:Float(?:32|64)|(?:Int|Uint)(?:8|16|32)|Uint8Clamped)/ is ok.


jdalton:
Thank you @lvzhenbang!

PR #4361, opened 2019-07-11
chore: add tools to lint git commit msg against the conventions

legend80s:


falsyvalues:
Hi @legend80s!
Thanks, I think we are fine.

PR #4365, opened 2019-07-15
Update map.js

McZenith:
Hello! I am really interested in adding my quota to the community. I am a javascript newbie.


falsyvalues:
Thanks for the PRs @McZenith!

In this case we prefer the more readable, but longer syntax form.

PR #4367, opened 2019-07-16
Fix: clone certain regexps in Microsoft Edge

rluba:
Cloning a `RegExp` _without flags_ failed in Microsoft Edge (observed in Edge 42.17134.1.0 / EdgeHTML 17.17134).

The `RegExp` stringifies its flags as "undefined", causing `cloneRegExp` to pass the string `"undefined"` as second argument to the `RegExp` constructor (which blows up with a `SyntaxError`).

### DIFF
--- a/.internal/cloneRegExp.js
+++ b/.internal/cloneRegExp.js
@@ -9,7 +9,8 @@ const reFlags = /\w*$/
  * @returns {Object} Returns the cloned regexp.
  */
 function cloneRegExp(regexp) {
-  const result = new regexp.constructor(regexp.source, reFlags.exec(regexp))
+  const flags = reFlags.exec(regexp);
+  const result = new regexp.constructor(regexp.source, flags[0] === 'undefined' ? undefined : flags);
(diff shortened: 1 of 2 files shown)


jdalton:
Hi @rluba!

I cannot produce an error using `_.clone(/a/)` or `_.cloneDeep(/a/)` in Edge 17.
The result of `reFlags.exec()` is `[""]`

rluba:
I’m sorry, you’re correct. It seems like it’s caused by a shim.

If I run `/\w*$/.exec(/a/)` in Edge’s console _on new tab_, it returns `[""]`. But if I try it on a page that contains scripts we’ve transpiled with Babel, it returns `["undefined"]`.

I’ll have to dig deeper to find the culprit that’s causing this. Sorry for wasting your time, @jdalton. 😞

rluba:
In case anyone ever stumbles upon this: It was caused by [a bug in core-js 3.0.1](https://github.com/zloirock/core-js/issues/536) [that’s been fixed](https://github.com/zloirock/core-js/pull/537).

PR #4371, opened 2019-07-17
clone deep with proto

gkio:


jdalton:
Hi @gkio!

I think wiring up the prototype is interesting but not as it's own method. I wouldn't mind experimenting with things being tweaked for the normal clone/cloneDeep case though.

PR #4382, opened 2019-07-20
Removed unneeded ternary

tsogi:


jdalton:
This is needed to allow for `-0`.

PR #4396, opened 2019-08-07
remove recursion from baseClone.

prestonalexwebster:
Recursion implementation of baseClone fails on  very deep objects (because of CallStack limit):
for example:
```
const linkedList = {next: null, value: 3};
for(let i = 0; i < 10**6; ++i){
    linkedList = {next: linkedList, value: 5};
}
cloneDeep(linkedList); //Error
```

The commited version fixes this issue by replacing recursion by loop


jdalton:

PR #4405, opened 2019-08-15
Do not use property shorthand in countBy documentation

blikblum:
property shorthand support is removed in master / v5


jdalton:
Thank you @blikblum!

PR #4409, opened 2019-08-16
Add missing isObject import to isIterateeCall

blikblum:
isObject import is missing in isIterateeCall


jdalton:
Thank you @blikblum!

PR #4410, opened 2019-08-16
Coerce indexOf fromIndex parameter to integer using toInteger

blikblum:
Uses toInteger instead of plus sign to coerce fromIndex parameter in indexOf.

This fixes use cases like `indexOf([1, 2, 1, 2], 2, 2.1)`


blikblum:
Here is a test case from v4 covering this issue: https://github.com/lodash/lodash/blob/4.17.5-post/test/test.js#L6247

jdalton:
Thank you again and again @blikblum!

PR #4411, opened 2019-08-16
Enable indexOf tests

blikblum:
Renames findIndex-and-indexOf.js to indexOf.test.js (findIndex is removed) and adjust code accordingly


jdalton:
Thank you @blikblum!

PR #4412, opened 2019-08-16
Coerce drop and dropRight n param using toInteger

blikblum:
It fixes n param coercion along side with some edge cases in dropRight

Enable the respective tests

Notes
 * i removed the tests that checks iteratee support since this feature is removed and is not working anymore
 * the lazy sequence tests passes but seems is not testing the actual code since chaining support is removed. Should i remove these tests

### DIFF
--- a/drop.js
+++ b/drop.js
@@ -1,4 +1,5 @@
 import slice from './slice.js'
+import toInteger from './toInteger.js'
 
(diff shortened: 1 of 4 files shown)


jdalton:
Thank you @blikblum!

PR #4413, opened 2019-08-17
Coerce to integer and set default value for chunk size parameter

blikblum:
Fix chunk size parameter handling
Enable chunk tests, removing test for iteratee support


jdalton:
Thank you @blikblum!

Issue #4415, opened 2019-08-17
Lint files


phapdinh:
when i run `npm run style` in your repo I get 37 eslint error messages
![lodash-linting](https://user-images.githubusercontent.com/5825929/63214772-923f1a00-c0ea-11e9-98ac-8aafbd6c6710.png)

jdalton:
Moved to #4416.

PR #4416, opened 2019-08-17
lint files

phapdinh:
This fixes issue issue #4415


.eslintrc.js
```diff
@@ -83,7 +83,7 @@ module.exports = {
       'requireReturnForObjectLiteral': false
     }],
 
-    'arrow-parens': ['error', 'always'],
+    // 'arrow-parens': ['error', 'always'],
```
phapdinh: I commented out this rule because sometimes arrow functions if they have only one parameter don't need parens

jdalton: It's fine to enable. I want to be consistent and always use parens.

phapdinh: I have added back this rule

jdalton:
Thank you @phapdinh!

PR #4417, opened 2019-08-17
Implement asciiWords and unicodeWords using String.match instead of binding RegExp.exec

blikblum:
The current implementation using a bound RegExp.exec has a strange behavior. It alternates between expected result and null in each subsequent invocation: See https://codepen.io/blikblum/pen/oNvLGQv

It reverts the implementation to the same as v4
Also enable tests


jdalton:
Thank you @blikblum!

PR #4418, opened 2019-08-17
Enable create tests and adapt it for the new primitive prototype behavior

blikblum:
It enables the create tests and adapt it to match new behavior

Had to change how created objects are compared because node `assert.deepStrictEqual` takes into consideration to prototype

For falsey and primitive as prototype, just check if is a object

Let me know if there's a better way to handle these tests

### DIFF
--- a/test/create.test.js
+++ b/test/create.test.js
@@ -1,6 +1,6 @@
 import assert from 'assert';
 import lodashStable from 'lodash';
-import { falsey, stubObject, primitives, stubTrue } from './utils.js';
+import { falsey, primitives, stubTrue } from './utils.js';
 import create from '../create.js';
 import keys from '../keys.js';
 
@@ -33,7 +33,10 @@ describe('create', function() {
 
     assert.ok(actual instanceof Circle);
     assert.ok(actual instanceof Shape);
-    assert.deepStrictEqual(Circle.prototype, expected);
+    assert.deepStrictEqual(Object.keys(Circle.prototype), Object.keys(expected));
+    Object.keys(expected).forEach(property => {
+      assert.strictEqual(Circle.prototype[property], expected[property]);
+    });
   });
 
   it('should assign own properties', function() {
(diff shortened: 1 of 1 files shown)


test/create.test.js
```diff
@@ -55,23 +64,23 @@ describe('create', function() {
...
-
+  it('should accept a primitive `prototype`', function() {
     var actual = lodashStable.map(primitives, function(value, index) {
-      return lodashStable.isPlainObject(index ? create(value) : create());
+      return index ? create(value) : create();
     });
 
-    assert.deepStrictEqual(actual, expected);
+    actual.forEach(value => {
+      assert.ok(value && typeof value === 'object');
```
jdalton: You can use `lodashStable.isObject()` there.

test/create.test.js
```diff
@@ -43,7 +46,13 @@ describe('create', function() {
     }
     Foo.prototype.b = 2;
 
-    assert.deepStrictEqual(create({}, new Foo), { 'a': 1, 'c': 3 });
+    var actual = create({}, new Foo);
+    var expected = { 'a': 1, 'c': 3 };
+
+    assert.deepStrictEqual(Object.keys(actual), Object.keys(expected));
+    Object.keys(expected).forEach(property => {
```
jdalton: I see in a couple places we do `Object.keys(expected)` would it make sense to store the result in a var when possible?

test/create.test.js
```diff
@@ -33,7 +33,10 @@ describe('create', function() {
 
     assert.ok(actual instanceof Circle);
     assert.ok(actual instanceof Shape);
-    assert.deepStrictEqual(Circle.prototype, expected);
+    assert.deepStrictEqual(Object.keys(Circle.prototype), Object.keys(expected));
+    Object.keys(expected).forEach(property => {
```
jdalton: arrow functions should wrap the param in parens (style consistency nit)

blikblum: done

test/create.test.js
blikblum: done

jdalton:
Thanks @blikblum!

PR #4419, opened 2019-08-17
Lint tests

phapdinh:
Hi @jdalton, I thought I would help with code consistency between your test files and the other lodash files by adding the test/*.test.js to your linting in npm run style. Most of this was done by eslint with the --fix flag. I just added mocha and _ to the configurations. I ran npm run test:wip and it passes. It is up to you if you want the feature, but I thought I put it out there.


blikblum:
IMO, in this moment, it should not bother with test files because many of these files are not relevant anymore and will surely be removed in future

phapdinh:
I see the *.test.js files being updated and recent pull requests so I thought they were going to be used in the future, but I might be wrong

.eslintrc.js
```diff
@@ -3,7 +3,11 @@ module.exports = {
   'plugins': ['import'],
   'env': {
     'es6': true,
-    'node': true
+    'node': true,
+    'mocha': true
+  },
+  'globals': {
+    '_': 'readonly'
```
armano2, line 10: you should all those env variables coditionally only if file is `*.test.js`
```js
overrides: [
  // all test files
  {
      files: ['**/*.test.js'],
      env: {
        'mocha': true
      },
      'globals': {
         '_': 'readonly'
      ]
  }
]
```

phapdinh:
too many conflicts

PR #4429, opened 2019-08-21
merge: copy function values independent of source index 

blikblum:
This is port of a fix introduced in 4.7.11

Enabled tests. The sparse array tests fails but is related to changes in `keysIn`


jdalton:
Awesome @blikblum!

PR #4430, opened 2019-08-21
Ensure after n parameter is coerced from NaN to 0 

blikblum:
Uses a simple fallback to 0 unlike v4 which is using `toInteger`. Cover all tests


jdalton:
Thank you @blikblum!

PR #4436, opened 2019-08-23
Enable case methods tests

blikblum:
Enable tests for camelCase, startCase, upperCase, kebabCase, lowerCase and snakeCase methods

Remove test that checks for deburr to account for new behavior

Also updated roadmap in wiki


jdalton:
Thanks @blikblum!

PR #4440, opened 2019-08-23
Coerce findLastIndex and lastIndexOf fromIndex parameter to integer 

blikblum:
Makes all related tests pass


jdalton:
Thank you @blikblum!

PR #4441, opened 2019-08-24
Simplify toString

blikblum:
It uses native array map instead of custom one and assumes Symbol.prototype.toString exists

### DIFF
--- a/toString.js
+++ b/toString.js
@@ -1,4 +1,3 @@
-import map from './map.js'
 import isSymbol from './isSymbol.js'
 
 /** Used as references for various `Number` constants. */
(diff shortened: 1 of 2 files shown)


toString.js
```diff
@@ -36,10 +32,10 @@ function toString(value) {
   }
   if (Array.isArray(value)) {
     // Recursively convert values (susceptible to call stack limits).
-    return `${map(value, (other) => other == null ? other : toString(other))}`
+    return `${value.map((other) => other == null ? other : toString(other))}`
```
vivekjain202: i feel `()` for map parameters is redundant here, can we not use 

return `${value.map( other => other == null ? other : toString(other))}`  ?

blikblum: Its the project style. See https://github.com/lodash/lodash/pull/4418#discussion_r316724333

jdalton:
Thank you @blikblum!

PR #4442, opened 2019-08-24
Fix string methods to handle empty values

blikblum:
It fixes the tests for empty values input trying to avoid importing toString where possible to make the dependency tree as flat as possible (v4 uses toString every where)


jdalton:
Thank you @blikblum!

Issue #4438, opened 2019-08-23
_.orderBy does not accept iteratee path arrays


samuelcolburn:
I don't know if this is intended behavior, but  `_.orderBy` does not work with path arrays.

An example I created using the REPL in the docs:
```javascript
var users = [
  { id: '4', address: { zipCode: 4, streetName: 'Beta' } },
  { id: '3', address: { zipCode: 3, streetName: 'Alpha' } },
  { id: '1', address: { zipCode: 1, streetName: 'Alpha' } },
  { id: '2', address: { zipCode: 2, streetName: 'Alpha' } },
  { id: '5', address: { zipCode: 4, streetName: 'Alpha' } },
];
 
// Sort by `address.zipCode` in ascending order
// and by `address.streetName` in descending order.

// This works
 _.orderBy(
  users,
  ['address.zipCode', 'address.streetName'],
  ['asc', 'desc']
);

// This doesn't
_.orderBy(
  users,
  [['address','zipCode'], ['address.streetName']],
  ['asc', 'desc']
);

// => objects for [
//     ['1', [1, 'Alpha']],
//     ['2', [2, 'Alpha']],
//     ['3', [3, 'Alpha']],
//     ['4', [4, 'Beta']],
//     ['5', [4, 'Alpha']]
// ]
```

REPL link: [https://runkit.com/samuelcolburn/5d6023bcb4e038001412b42b](https://runkit.com/samuelcolburn/5d6023bcb4e038001412b42b
)
#### Expected Behavior
That a path array `['address', 'zipCode']`  is treated the same as a dot-delimited path string`'address.zipCode'`. 

#### Actual Behavior
`'address.zipCode'` works as expected. `['address', 'zipCode']` is not sorted at all.

My apologies if this is intended behavior or if all iteratees are treated this way, it was my assumption based on functions like `_.get()` that these two types of paths were treated equivalently.

jdalton:
Moved to #4453.

PR #4444, opened 2019-08-25
fix: allowing orderBy iteratees to be arrays like _.get

shubhamzanwar:
#4438

### DIFF
--- a/orderBy.js
+++ b/orderBy.js
@@ -41,7 +41,14 @@ function orderBy(collection, iteratees, orders) {
   }
   if (!Array.isArray(iteratees)) {
     iteratees = iteratees == null ? [] : [iteratees]
+    iteratees = iteratees.map((iteratee) => {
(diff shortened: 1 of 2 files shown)


orderBy.js
```diff
@@ -41,7 +41,14 @@ function orderBy(collection, iteratees, orders) {
   }
   if (!Array.isArray(iteratees)) {
     iteratees = iteratees == null ? [] : [iteratees]
+    iteratees = iteratees.map((iteratee) => {
+      if (Array.isArray(iteratee)) {
+        return iteratee.join('.')
+      }
```
jdalton: We don't want to use `join()` here as array paths can capture things not possible or more complex in dot form.

shubhamzanwar: Could you give me an example of such a case, I'll try improving this PR :)

jdalton:
Thank you @shubhamzanwar!
I'm closing this in favor of #4453.

PR #4445, opened 2019-08-26
Fix flow and flowRight parameter handling

blikblum:
Use native rest parameters / spread operator in flow and flowRight

Enable tests, removing tests related to removed features


blikblum:
> Can you also update the usage examples in the jsdoc comments above flow and flowRight?

Done

PR #4448, opened 2019-08-28
handled return default value for falsy result from baseGet - used isU…

alirezavalizade:
- handled return default value for falsy result from baseGet
- used isUndefined function

example:
```
const obj = {
  a: {
    b: {
      c: "Hi",
      d: null
    }
  }
};

get(obj, 'a.b.c')
output // => Hi

get(obj, 'a.b.d', '--default-value--')
currently it'll return null
output // => null

but with this pull request
get(obj, 'a.b.d', '--default-value--', true)
output // => '--default-value--'
````

### DIFF
--- a/get.js
+++ b/get.js
@@ -1,4 +1,6 @@
 import baseGet from './.internal/baseGet.js'
+import isNumber from './isNumber.js'
+import isUndefined from './isUndefined.js'
 
 /**
  * Gets the value at `path` of `object`. If the resolved value is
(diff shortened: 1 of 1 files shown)


alirezavalizade:
Hi @jdalton,
ok sounds good, but I think anybody that uses lodash doesn't want to be involve with these helpers, I think that's the benefit of lodash, easy to use and handy and about this function, personally I'm using it a lot in my projects and I faced with this small issue.
anyway, thanks

PR #4464, opened 2019-09-12
Added filler option for chunk method

sharmaadityaHQ:
Behaviour shown by addition is as follows:-
1.chunk(['a', 'b', 'c', 'd'], 2)
 // => [['a', 'b'], ['c', 'd']]

2.chunk(['a', 'b', 'c', 'd'], 3)
// => [['a', 'b', 'c'], ['d']]

3.chunk(['a', 'b', 'c', 'd', 3, 'e')
// => [['a', 'b', 'c'], ['d', 'e', 'e']]

4.chunk(['a', 'b', 'c', 'd', 3, null)
// => [['a', 'b', 'c'], ['d', null, null]]

### DIFF
--- a/chunk.js
+++ b/chunk.js
@@ -13,42 +13,39 @@ import toInteger from './toInteger.js'
  * @returns {Array} Returns the new array of chunks.
  * @example
  *
- * chunk({
- *    array: ['a','b','c','d'],
- *    size: 2
- * })
+ * chunk(['a','b','c','d'], 2)
  * // => [['a', 'b'], ['c', 'd']]
  *
- * chunk({
- *    array: ['a','b','c','d'],
- *    size: 3
- * })
+ * chunk(['a','b','c','d'], 3)
  * // => [['a', 'b', 'c'], ['d']]
  *
- * chunk({
- *    array: ['a','b','c','d'],
+ * chunk(['a','b','c','d'],{
  *    size: 3,
  *    filler: 'e'
  * })
  * // => [['a', 'b', 'c'], ['d', 'e', 'e']]
  *
- * chunk({
- *    array: ['a','b','c','d'],
+ * chunk(['a','b','c','d'],{
  *    size: 3,
  *    filler: null
  * })
  * // => [['a', 'b', 'c'], ['d', null, null]]
  *
- * chunk({
- *    array: ['a','b','c','d'],
+ * chunk(['a','b','c','d'],{
  *    size: 3,
  *    filler: undefined
  * })
  * // => [[ 'a', 'b', 'c' ], [ 'd', undefined, undefined ]]
  */
-function chunk(options) {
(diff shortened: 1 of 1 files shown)


jdalton:
I like this though I think we should move to an options object and not another param. This will also allow folks to specify `filler` with an `undefined`. You can juggle `size` if it's an object treat it as `options` else as `size`.

sharmaaditya570191:
Hi @jdalton. I have changed my code as you suggested. Please have a look.

MarcelloTheArcane:
@jdalton that would mean it _always_ fills, surely?

sharmaaditya570191:
Hey @MarcelloTheArcane as written by me in example it would not always fill.
Eg.chunk({
  array: ['a','b','c','d'],
  size: 3
})
// => [['a', 'b', 'c'], ['d']]

jdalton:
Oh @sharmaaditya570191 I meant an options object for the second param (`size`). The `array` argument would still be used.

sharmaaditya570191:
I have made necessary changes now. Please have a look @jdalton

sharmaaditya570191:
Please have a look @jdalton

chunk.js
```diff
@@ -13,24 +13,54 @@ import toInteger from './toInteger.js'
...
+ *
+ * chunk(['a','b','c','d'],{
+ *    size: 3,
+ *    filler: undefined
+ * })
+ * // => [[ 'a', 'b', 'c' ], [ 'd', undefined, undefined ]]
  */
-function chunk(array, size = 1) {
-  size = Math.max(toInteger(size), 0)
+function chunk(array, options) {
```
Hovakimyan: I think it's better to make function to get the size.
```
/**
 * @param options: number or Object
 */
function getMaxSize(options) {
    if (typeof options !== 'object') {
        return Math.max(toInteger(options), 0)
    } else {
        return Math.max(toInteger(options.size), 0)
    }
}
```
And then use it
```
const size = getMaxSize(options)
```

chunk.js
```diff
@@ -13,24 +13,54 @@ import toInteger from './toInteger.js'
...
   if (!length || size < 1) {
     return []
   }
   let index = 0
   let resIndex = 0
   const result = new Array(Math.ceil(length / size))
-
   while (index < length) {
-    result[resIndex++] = slice(array, index, (index += size))
+    const arrayChunk = slice(array, index, (index += size))
```
Hovakimyan: Maybe something like this will be a good solution too.
```
while (index < length) {
    result[resIndex++] = slice(array, index, (index += size))
}
if (result[resIndex - 1].length !== size) {
    for (let i = result[resIndex - 1].length; i < size; i++) {
        result[resIndex - 1].push(options.filler)
    }
}
```
In this case we will do check only after creating all chunks

jdalton:

PR #4465, opened 2019-09-13
Added fromObject

zaguiini:


jdalton:
Hi @zaguiini!

Thank you for the PR, but I think this kind of functionality is better suited as a one-off util module.

zaguiini:
Ok, sure! I published it as https://github.com/zaguiini/from-object

PR #4469, opened 2019-09-16
Deep clone Buffer should not use slice

16546513:
i think there is something wrong.


16546513:
> i think unit test need update too @16546513

emm, i have no idea where i can update.

xqin:
https://github.com/lodash/lodash/blob/master/test/clone-methods.js#L167

image

jdalton:
Closing due to inactivity.

PR #4480, opened 2019-09-25
:bug defaultValue when the get parameter is null

karzanOnline:
### Problem
When the parameter of the get method is null, the defaultValue setting is not entered.
This is not the normal process we expect

E.g
```
import { get } from 'lodash'

const params = {
  a: null
}

get(params, 'a', 'defaultValue')   // return null

```

### DIFF
--- a/get.js
+++ b/get.js
@@ -26,7 +26,7 @@ import baseGet from './.internal/baseGet.js'
  */
 function get(object, path, defaultValue) {
   const result = object == null ? undefined : baseGet(object, path)
-  return result === undefined ? defaultValue : result
+  return (result === undefined || result === null) ? defaultValue : result
 }
 
 export default get


MV88:
hi @karzanOnline i'm not a maintainer or a contributor, but from the [get documentation](https://lodash.com/docs/4.17.15#get) it specifies that only in case of undefined resolved values it will returns the default value. 
on the other hand null is different from undefined, it means ‘no value’ or nonexistence of any value, so I assume the actual behaviour is intended.

probably something like this could be useful for you
```js
const getNotNullOrUndefined (params, path, default) => {
    const res = get(params, path, default);
    return isNull(res) ? default : res;
}
```

falsyvalues:
Hi @karzanOnline!

@MV88 is right we fallback to default value only on `undefined`.

PR #4482, opened 2019-09-25
perfomance fixes, eslint fixes

anatoliy8493:


falsyvalues:
Hi @anatoliy8493!
Thank you, we are fine with current solution.

jdalton:
Thank you @falsyvalues!

Ya know, after all these years I think it's time we go `===` :p
It's easy for new contributors to reason about.

falsyvalues:
@jdalton Sure thing 👍

jdalton:
🙃

anatoliy8493:
@falsyvalues Hi! It might seem like a strong word, however if we use === instead of == we no need to JS Type Conversion

jdalton:
In this case our usage of `==` was when there was no type conversion so basically spec'd identical steps as `===`. There is no noticeable performance difference with it.

PR #4495, opened 2019-10-01
Add _.getWith function

sharmaaditya570191:
Takes customizer to customize values retrieved from _.get

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -13129,7 +13129,7 @@
     }
 
     /**
-     * This method is similat to '_.get' except that it accepts `customizer`
+     * This method is similar to '_.get' except that it accepts `customizer`
      * which is invoked on the value returned from '_.get' to transform it
      * before it is returned.
      * @static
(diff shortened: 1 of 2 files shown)


sharmaaditya570191:
Hi @jdalton . How should I resolve the conflicts?

lodash.js
```diff
@@ -13128,6 +13128,34 @@
       return result === undefined ? defaultValue : result;
     }
 
+    /**
+     * This method is similat to '_.get' except that it accepts `customizer`
```
georgeglessner: similat -> similar

sharmaaditya570191:
Hi @georgeglessner .Thanks for the review. I have corrected the typo.

jdalton:
Hi @sharmaaditya570191!

In this case the customizer is just mapping the return value which isn't really what our `with` methods do. In this case I think this fits better as a one-off helper module in your project/npm.

sharmaaditya570191:
Hi @jdalton . Where do I need to move this in order to add this in project/npm.

PR #4496, opened 2019-10-03
Short circuit sortedIndexBy methods for empty arrays

megawac:
I have a use case that involves building many smallish sorted arrays so I'm using `sortedIndexBy` as part of an insertion sort logic. As empty arrays are a common condition for me, I figured I should reach out and see if we could get them short circuited. Cheers.


jdalton:
Thanks @megawac!

You may also want to open this against the `4.17.15-post` tag so that it will get into the next minor/patch release.

megawac:
Sounds good

PR #4502, opened 2019-10-06
refactor: updated uniqueId util

Gherciu:
updated `uniqueId` util with less code ;)


falsyvalues:
Hi @Gherciu!
We prefer readability over number of used characters. Cheers ✌️

PR #4453, opened 2019-08-30
add order by behaviour with array similar to get method

Mohit21GoJs:
fix issue #[4438](https://github.com/lodash/lodash/issues/4438)

PR #4513, opened 2019-10-10
Ensure `orderBy` will accept iteratee path arrays #4438

falsyvalues:
While backporting https://github.com/lodash/lodash/pull/4453 for #4438 I noticed it just don't work, we didn't catch it because tests for `orderBy` were disabled. This contains necessary fixes (I will forward port it to master later on).


jdalton:
Thank you @falsyvalues!

Issue #4320, opened 2019-06-07
lodash.isEqual returns different result when order of arguments is switched


capaj:
As found out by @benjamn here: https://github.com/apollographql/apollo-client/pull/4924#issuecomment-499280139

here is how it runs in a console on lodash.com:
![image](https://user-images.githubusercontent.com/1305378/59087335-aacdcf80-8904-11e9-8cd2-267affbda1b0.png)

miscoder:
@falsyvalues I'll take this issue.

benjamn:
@miscoder I’m sure there are multiple solutions that would work, but [here’s one strategy](https://github.com/benjamn/wryware/blob/50e7d6181b0b937ad54b543118e823ad98e2bca9/packages/equality/src/equality.ts#L121-L129) that I settled on.

cukejianya:
@miscoder Are you still working on this issue?

cukejianya:
If it's alright with @miscoder, I would like to take on this issue. Not sure if he is still working on it.

ccjmne:
It looks like @benjamn pretty much worked out the hard parts already, and @miscoder doesn't seem to have done any progress in the 3+ months he's had the claim to this issue. I think you can go ahead, @cukejianya.
If you aren't interested anymore or don't have the time these days, just let us know 🙂.

cukejianya:
@ccjmne Sounds good.

jdalton:
Closed by #4515.

PR #4515, opened 2019-10-12
Fix lodash.isEqual for circular references #4320

cukejianya:
`lodash.isEqual` for circular reference in objects/arrays returned different results based on order of arguments. More detail can be found here:  #4320

image


jdalton:
Awesome @cukejianya!

Did the CLA bot not kick in for this?

jdalton:
NM i see it.

Issue #89, opened 2012-10-10
arr.insert(index, item)


coolaj86:
```
_.insert = function (arr, index, item) {
  arr.splice(index, 0, item);
};
```

I know it's hardly worth including, but I forget all the time that splice does insert - and there's a lot of google traffic for "javascript array insert", y'know.

JabbyPanda:
Good use-case for the using `_.mixin` API in our projects

continuata:
could we re-open this?
Array#splice is not a solution, as it mutates the original array instead of returning a new spliced one.
I thought one of the main points of lodash was to ensure functional versions of javascript methods are implemented.

something like this would work ...
```es6
  _.insert (arr, index, insertObject) {
    let before = _.slice(arr, 0, index)
    let after = _.slice(arr, index)
    return _.flatten([before, insertObject, after])
  }
```

Happy to open a PR for this, if it's considered a useful addition

albertogasparin:
+1 :+1: 
It would be nice to provide a *non mutable* helper for adding elements to an array. I'm suggesting:

``` js
_.insert(array, value, [index])
```

With `index` as an optional argument (falls back to `_.concat` if missing)

jdalton:
Thanks @albertogasparin!

We'll keep an eye on [the popularity of the request](https://github.com/lodash/lodash/blob/master/.github/CONTRIBUTING.md#feature-requests).

TurBas:
+1

prose100:
It would be nice for the `insert` function to accept a number or an array. 
So instead, `_.insert(array, insertion, [index])` where `insertion` is a number or an array.

It would be documented as follows:
insertion (number or Array): The value or values to insert. 

`_.insert([1,2,6,7], [3,4,5], 2)`
//[1,2,3,4,5,6,7]

`_.insert([1,2,4,5], 3, 2)`
//[1,2,3,4,5]

bompi88:
+1

freewind:
I'm searching 'how to insert an element to array with lodash' and reached this issue

elranu:
Hi
sorry but is there a method that while inserts, inserts in the order specified by a selected property?

lest say something like ` _.insert(arr, itemToInsert, ['age'], ['asc']);` ?

elranu:
I made a performance between insert many items in an array and then sort , and insert in a order way.  
https://jsbin.com/lezanab/1/edit?js,console

[...]

PR #4519, opened 2019-10-17
Add _.insert function for arrays

DoloMike:
Proposing the following function for inserting items into an array at a given index per [issue #89](https://github.com/lodash/lodash/issues/89#issuecomment-542967821)

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -7509,10 +7509,10 @@
      * @returns {Array} Returns a new array with the value inserted at the given index.
      * @example
      *
-     * _.insert(2, 'Hello', ['Hey', 'Hi', 'Howdy'])
+     * _.insert(['Hey', 'Hi', 'Howdy'], 2, 'Hello')
      * // => '['Hey', 'Hi', 'Hello', 'Howdy']'
      * 
-     * _.insert(-2, 'Hello', ['Hey', 'Hi', 'Howdy'])
+     * _.insert(['Hey', 'Hi', 'Howdy'], -2, 'Hello')
      * // => '["Hey", "Hello", "Hi", "Howdy"]'
      */
     function insert(array = [], index = 0, value) {
--- a/lodash.js
+++ b/lodash.js
@@ -7497,6 +7497,30 @@
     }
 
     /**
+     * Inserts `item` at the given `index` into the `array`.
+     *
+     * @static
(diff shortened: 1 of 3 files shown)


DoloMike:
@jdalton any thoughts or comments on this? Not a lot of discussion in the thread so far.

lodash.js
```diff
@@ -7496,6 +7496,29 @@
...
+     *
+     * _.insert(2, 'Hello', ['Hey', 'Hi', 'Howdy'])
+     * // => '['Hey', 'Hi', 'Hello', 'Howdy']'
+     * 
+     * _.insert(-2, 'Hello', ['Hey', 'Hi', 'Howdy'])
+     * // => '["Hey", "Hello", "Hi", "Howdy"]'
+     */
+    function insert(index = 0, item, array = []) {
+      return [...array.slice(0, index), item, ...array.slice(index)];
+    }
```
jdalton: We do data first in our generic methods so it would be `array`, `index`, `value`.

ZimGil: Shouldn't it be `array`, `value`, `index` for cases where `index` is omitted and `value` should be inserted to default `index` (0?)

DoloMike: Thanks. Sorry for taking so long but I've finally resolved this.

DoloMike:
@falsyvalues this has lingered quite a bit. Not sure if it's still wanted or not but maybe you'll see this. Thanks.

lodash.js
```diff
@@ -7496,6 +7496,29 @@
...
+     * @example
+     *
+     * _.insert(['Hey', 'Hi', 'Howdy'], 2, 'Hello')
+     * // => '['Hey', 'Hi', 'Hello', 'Howdy']'
+     * 
+     * _.insert(['Hey', 'Hi', 'Howdy'], -2, 'Hello')
+     * // => '["Hey", "Hello", "Hi", "Howdy"]'
+     */
+    function insert(array = [], index = 0, value) {
+      return [...array.slice(0, index), value, ...array.slice(index)];
```
doubleppereira, line 7519: I think a `array.splice(index, 0, value)` in the end would be more performant although we would need to store array in a variable since it's mutating the array..

jdalton:

PR #4524, opened 2019-10-23
feat: Get method add  defaultValue type check

iam-frankqiu:
When the default value is given in  Get method, if the type of ultimate result not strict equal the type of default value, then return the default value.


jdalton:
Thank you @iam-frankqiu, but I don't want to introduce this kind of check. Values can be many things and a default value is just one of them. I think maybe something like TypeScript may help here.

PR #4531, opened 2019-11-01
Remove extraneous conditional

camjc:
Since `function flow(funcs)` changed to `function flow(...funcs)` this ternary became unnecessary, since funcs will always have a length now.


jdalton:
Thanks @camjc!

Cassieminkus1 (approved):

PR #4550, opened 2019-11-19
simplify useless ternary expression

DirtyWhite:
sir,I am reading the source code.
I can't understand what this tenary expression for，I think it could be simplified.


jdalton:
Hi @DirtyWhite!

The logic is needed for the case of `-0`.

PR #4554, opened 2019-11-19
Updated the checks of isEmpty and isBoolean methods

RajuPedda:


isEmpty.js
```diff
@@ -27,11 +19,38 @@ const hasOwnProperty = Object.prototype.hasOwnProperty
...
+ * // => true
+ *
+ * isEmpty(false)
  * // => true
  *
  * isEmpty(1)
  * // => true
+ * 
+ * isEmpty({})
+ * // => true
```
Mifrill: nice! this is most important example to shows the cool point of `_.isEmpty()`, because of this: (javascript)

```javascript
  isEmpty({});               // false
```

```javascript
  _.isEmpty({});               // true
```

I think it's would be great to add this example to [documentation on sait](https://lodash.com/docs/4.17.15#isEmpty) too

RajuPedda: @jdalton Can you please review my changes?

RajuPedda:
@jdalton @Mifrill Can you please review my changes?

jdalton:
Hi @RajuPedda!

I'm not really looking to change either method at this time. Thank you though for this and other PRs!

PR #4557, opened 2019-11-20
Added extra check for undefined & null

RajuPedda:


jdalton:
Hi @RajuPedda!

Thank you for the PR. I'm okay with the existing behavior.

PR #4573, opened 2019-12-04
Update `isEmpty` method to support non-collection values judge

lingxiaoguang:
The current isEmpty method is only support collection values judge like Set,Map,Object,Arguments and do not support the other types like Boolean, Number, RegExp,Symbol etc.  Why not support the both. So I make this pull request.

### DIFF
--- a/isEmpty.js
+++ b/isEmpty.js
@@ -46,6 +46,9 @@ function isEmpty(value) {
   if (value == null) {
     return true
   }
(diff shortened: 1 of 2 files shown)


RajuPeddaM:
@lingxiaoguang There is a [PR](https://github.com/lodash/lodash/pull/4554) to support all these requirements and also functions ( which returns Boolean / Array / Object values). Please check @jdalton comment there.

lingxiaoguang:
@jdalton

PR #4574, opened 2019-12-04
Fixed faulty example for uniqBy and sortedUniqBy

stefansurkamp:
`Math.floor` will resolve to intergers, compare https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/floor


falsyvalues:
Hi @stefansurkamp!
I don't think this is correct. Please test PR.

PR #4580, opened 2019-12-08
Added Strict Equality Comparison.

divyanshu-rawat:
Replaced `Abstract Equality Comparison (==)` with `Strict Equality Comparison (===)`.


jdalton:
Hi @divyanshu-rawat!

I'm fine with the current comparison but wouldn't be opposed to changing it all-up.

PR #4581, opened 2019-12-08
Replaced Abstract with Strict Equality Comparison.

divyanshu-rawat:
Here, 

```
 * isEmpty(undefined)
 * // => true
```

and if we replace Abstract with Strict Equality Comparison.

then

```
 * isEmpty(undefined)
 * // => false
```

or are we treating `undefined` and `null` same here?

### DIFF
--- a/isEmpty.js
+++ b/isEmpty.js
@@ -43,7 +43,7 @@ const hasOwnProperty = Object.prototype.hasOwnProperty
  * // => false
  */
 function isEmpty(value) {
-  if (value == null) {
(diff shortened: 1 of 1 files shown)


falsyvalues:
Hi @divyanshu-rawat!
We do, thats why we don't replace these blindly.

RajuPeddaM:
This [PR](https://github.com/lodash/lodash/pull/4554) resolves the above strict equality case and other like Set, Map, Boolean/ Object constructor etc.

PR #4587, opened 2019-12-11
Update isPlainObject.js

ryanmtaylor:
Can somebody explain the point of iterating down the prototype chain and comparing?  This version fits the provided definition much better, more literally, and is much shorter.

### DIFF
--- a/isPlainObject.js
+++ b/isPlainObject.js
@@ -28,17 +28,7 @@ import isObjectLike from './isObjectLike.js'
  * // => true
  */
 function isPlainObject(value) {
-  if (!isObjectLike(value) || getTag(value) != '[object Object]') {
-    return false
-  }
-  if (Object.getPrototypeOf(value) === null) {
-    return true
-  }
-  let proto = value
-  while (Object.getPrototypeOf(proto) !== null) {
-    proto = Object.getPrototypeOf(proto)
-  }
-  return Object.getPrototypeOf(value) === proto
+  return !!value && (value.constructor === Object || Object.getPrototypeOf(value) === null)
 }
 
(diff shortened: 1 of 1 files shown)


rafde:
@ryanmtaylor https://github.com/lodash/lodash/commit/b36f21cbaff7c885223b44267c70e6711b754af6#diff-d518adcd51672639ee7c4b353eb8aaee may help with an explanation.

ryanmtaylor:
at the very least you shouldn't need this part:

```
  if (!isObjectLike(value) || getTag(value) != '[object Object]') {
    return false
  }
```

ryanmtaylor:
Is the suggestion that something like...
```
const ObjectPrototype = Object.getPrototypeOf({})
const ValuePrototype = Object.getPrototypeOf(value);
if(null === ValuePrototype || ObjectPrototype === ValuePrototype) {
  return true;
}
return false;
```
...could be invalid in certain contexts?

It still feels weird, since when can prototypes be recursive or exist twice in a prototype chain?  They can't (or the loop might go forever sometimes, right?) – so asking if this looped `proto === Object.getPrototypeOf(value)` (as we call this root-prototype object) is really equivalent to asking `Object.getPrototypeOf(Object.getPrototypeOf(value)) === null`.

I suppose it could be more like...
```
if(!value) return false;
const maybeRootPrototype = Object.getPrototypeOf(value);
if(maybeRootPrototype === null) return true;
return Object.getPrototypeOf(maybeRootPrototype) === null && maybeRootPrototype.constructor === value.constructor;
```
I mean when else can a prototype be null except for a "contexted" root-level `Object` and `Object.create(null)`?

Issue #4582, opened 2019-12-10
New HIGH vulnerability detected by audit in version 4.17.15


BorisNaguet:
Hello,

We haven't changed our dependencies recently, but there's apparently a new vulnerability detected by `npm audit`:

```
  High            lodash Package for Node.js lodash.js baseDifference()
                  Function Array Handling Memory Exhaustion DoS Weakness

  Package         lodash

  Patched in

  Dependency of   lodash

  Path            lodash

  More info       https://vulndb.cyberriskanalytics.com/vulnerabilities/218918
```

`npm list |grep lodash` effectively returns me 4.17.15

sgeffken:
Same here, a hotfix would be highly appreciated, whole CI broken down currently.

lokhaeri:
I had a look at the `baseDifference()` and it looks like in:

```
if (iteratee) {
    values = map(values, (value) => iteratee(value))
  }
  if (comparator) {
    includes = arrayIncludesWith
    isCommon = false
  }
  else if (values.length >= LARGE_ARRAY_SIZE) {
    includes = cacheHas
    isCommon = false
    values = new SetCache(values)
  }
```

in ` else if (values.length >= LARGE_ARRAY_SIZE) {`
there is no upper limit which can cause Memory Exhaustion
so the correct would be I assume something like this
```
if (comparator && values.length <= LARGE_ARRAY_SIZE) {
    includes = arrayIncludesWith
    isCommon = false
  }
  // not to allow to dump large amount of array into browser memory cache there should be an upper limit too
  else if (value.length < LARGE_ARRAY_MAX_SIZE) {
    includes = cacheHas
    isCommon = false
    values = new SetCache(values)
  }
```

msolwecki:
Fantastic that this was addressed so quickly, I had just been checking into our npm audit results and saw this vulnerability pop up and the PR get merged.  Do you happen to know when it'll be released?

jdalton:
Hi @BorisNaguet!

I'm not aware of vulndb.cyberriskanalytics and have not been notified in any of our security channels about this issue. If you're up for submitting a PR we'll review it but I'm a little miffed at the lack of communication with the folks creating these reports and the ones broadcasting them to users.

BorisNaguet:
@jdalton : I have no knowledge in vulndb.cyberriskanalytics, 

I actually found it strange that `npm audit` (which I thought was a standard npm tool) reported me an issue with a link to that website, where I can't have any details without paying for VulnDB!

[...]

PR #4589, opened 2019-12-12
baseDifference.js HIGH vulnerability fix detected by audit in v4.17.15

lokhaeri:
Possible High vulnerability fix.


jdalton:
Hi @lokhaeri!

Thanks for making a PR for #4582, however I can't access the security report to judge if this is the right fix. I don't like how low the limit is since I know our methods are using on large data sets.

lokhaeri:
Hi, look at the latest reports, they say about memory exhaust.

falsyvalues:
There is no real reasoning to limit such thing within library.

PR #4595, opened 2019-12-18
Added partitionsBy

imChiragSuthar:
Have added new function partitionsBy which will help to create multiple partitions based on predicate function.


falsyvalues:
Thanks @imChiragSuthar!

imChiragSuthar:
This function help us to create partition of elements based on predicate function. It creates multiple partitions based on condition or property or value.

PR #4605, opened 2020-01-03
dropWhile support drop multiple elements

daemonchen:
`dropWhile` support drop multiple elements at the same time, don't care from right or left.
may be this way is better than the older version ?


bnjmnt4n:
Hi @daemonchen! This behaviour is identical to that of `_.filter(array, _.negate(iteratee))`.

PR #4622, opened 2020-01-08
Re-add isEqual function

blikblum:
Seems IsEqual function was removed accidentally

### DIFF
--- a/isEqual.js
+++ b/isEqual.js
@@ -1,0 +1,35 @@
+import baseIsEqual from './.internal/baseIsEqual.js';
+
+/**
+ * Performs a deep comparison between two values to determine if they are
+ * equivalent.
+ *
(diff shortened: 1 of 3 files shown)


isEqual.js
```diff
@@ -0,0 +1,35 @@
...
+ * var other = { 'a': 1 };
+ *
+ * _.isEqual(object, other);
+ * // => true
+ *
+ * object === other;
+ * // => false
+ */
+function isEqual(value, other) {
+  return baseIsEqual(value, other);
```
ZimGil: Why not `var isEqual = baseIsEqual`?

blikblum:
I Just readded the old code.

But you have a point. Maybe better rename baseIsEqual to IsEqual

This can be done in a follow PR

PR #4623, opened 2020-01-08
Handle leading zeros in quote html entity

gwwar:
HTML entities may have [leading zeros](https://en.wikipedia.org/wiki/Character_encodings_in_HTML), so in this case `#39` `#039` both refer to the same character `'`

We noticed this while fixing this issue https://github.com/WordPress/gutenberg/pull/19187

I ran into trouble building and running tests locally, so I'm happy to follow up if there are more detailed instructions elsewhere for running this locally.

### Testing Instructions

- Verify that tests still pass
- No other regressions

### DIFF
--- a/test/unescape.js
+++ b/test/unescape.js
@@ -26,6 +26,12 @@ describe('unescape', function() {
     assert.strictEqual(unescape(escape(unescaped)), unescaped);
   });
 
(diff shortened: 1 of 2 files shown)


falsyvalues:
Thank You @gwwar!

PR #4627, opened 2020-01-10
Fix prototype pollution vulnerabilities, fix failing tests

joeattardi:


test/test.js
```diff
@@ -14923,7 +14963,7 @@
     });
 
     test('should throw a TypeError when receiving a non-array `array` argument', 1, function() {
-      raises(function() { _.spread(4, 2); }, TypeError);
```
jdalton: Does `raises` still work? In v3 we avoided `throws` because of some WebKit engines (older for sure by now).

joeattardi: When I ran the tests it was throwing errors that `raises` was not defined. Was not aware of the issues with `throws` though.

jdalton: That's okay, no worries. Thank you!

jbegleiter-sf:
@jdalton is this released?

remarkablemark:
This is not, I created my own repository for npm package with the fix: https://github.com/remarkablemark/lodash

jbegleiter-sf:
Thanks @remarkablemark , @jdalton any chance we could have this released as part of official? Would save me a headache.

Issue #2339, opened 2016-05-15
Suggestion: _.intersperse


AaronHarris:
A feature request wherein a single entity may be inserted between each element in an Array.

    _.intersperse([1, 2, 3], "&") //=> [1, "&", 2, "&", 3]

I think this is something that could be useful (it's there in [Haskell](http://zvon.org/other/haskell/Outputlist/intersperse_f.html)!) and using existing functions results in a lot of extra computation (eg. `map`, `flatten`, and `slice`). However it may be too simple of a function / doesn't do enough for lodash. Not sure how popular this would be, so curious to see what people think.

Also, if I'm willing to submit a PR for this, should I go ahead and do it or wait for votes?

jdalton:
Thanks @AaronHarris!

We'll keep an eye on [the popularity of the request](https://github.com/lodash/lodash/blob/master/.github/CONTRIBUTING.md#feature-requests).

jtheisen:
I for one would have expected this to be in lodash. You need this in list pretty printing, for example, when the built-in `Array.join` isn't enough.

kromit:
how many :+1:  are needed to be considered as worthy?

jtheisen:
After implementing something like that for a special use case, I figured that the proposed signature may not be quite enough. I suggest the following (additional) more complex usage:

```
_.intersperse([1, 2, 3], (x, y) => x + y) //=> [1, 3, 2, 5, 3]
```

There are two things that are sometimes needed here:

- The interspersed object may need to have distinct object identities (think HTML DOM elements) and thus be created individually and
- sometimes the interspersed object need to depend on the surrounding elements. I needed this in a case where I interspersed a `<span>,<span>` in a React JSX expression and React was telling me I should put unique keys on the interspersed elements.

Obviously the simple usage should also work.

jtheisen:
I would propose an additional, more sophisticated usage as well:

```
_.intersperse([1, 2, 3], (l, r) => l + r) // => [1, 3, 2, 5, 3]
```

I would suggest this because of some experience I just made with implementing an intersperse for a specific use case:

* The interspersed elements sometimes need to be newly created copies rather than the references to a single instance - think DOM elements.
* Sometimes the interspersed elements depend on the neighbors. In my case, I needed to have a `key` property defined on React elements for React to be happy. It was natural to have the key be derived from one of the neighboring elements.

There's another usage suggestion that may also be useful for the common case of formatting elements as "one, two *and* three":

```
[...]

PR #4629, opened 2020-01-15
add intersperse function

eladFrizi:
intersperse is a function with two parameters, an array and a separator that would be inserted between the elements. [ramda version](https://ramdajs.com/docs/#intersperse)
I think that it's a very useful function, I needed this in my work and I thought I would found it in Lodash. I didn't, but I saw that I'm not the only one who thinks like that. I found a very popular issue about this function. 

https://github.com/lodash/lodash/issues/2339

### DIFF
--- a/intersperse.js
+++ b/intersperse.js
@@ -4,10 +4,10 @@
  *
  * @param {Array} arr The array to iterate over.
  * @param {any} separator an element to insert between the collection elements.
- * @returns {Array} the array with the separator between each elements.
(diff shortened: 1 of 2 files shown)


falsyvalues:
Hi @eladFrizi!

As mentioned in #2339 we will keep an eye on [the popularity of the request](https://github.com/lodash/lodash/blob/master/.github/CONTRIBUTING.md#feature-requests).

PR #4655, opened 2020-02-13
Update random with exclude

tanrax:
Update random to generate a random number excluding an array of numbers.

As I have encountered the problem twice, I think it could be very interesting to include it.


jdalton:

Issue #4648, opened 2020-02-11
_.debounce().flush() fix never made it into produciton


TeaSeaLancs:
Hi lodash team,

Having been tracing an issue through my codebase, I discovered that running `flush` on a debounced function doesn't clear any timeouts.

There was a pull request raised a few years ago for this, https://github.com/lodash/lodash/pull/4078, which was merged, however those fixes don't seem to have made it into production, as can be evidenced at:

https://github.com/lodash/lodash/blob/4d5fdc492feabf6406dd37e5e843a1261414cbf8/debounce.js (esm version of 4.17.15)

https://github.com/lodash/lodash/blob/15e1557b2a97c8bbee22d873832d90ed3ba50ba7/debounce.js (latest update to debounce.js which is in master at time of writing)

I could re-raise this as a PR but was wondering if there was a reason why the fix never made it in?

falsyvalues:
Hi @TeaSeaLancs!

Thanks for catching this I will address that #4656.

PR #4657, opened 2020-02-14
Resurrect #4011 Fix _.omit mutating objects w/ Arrays (#4007)

falsyvalues:
#4648 inspired me to do some search and I also found that #4011 is missing in `4.17.12` and above.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -5614,7 +5614,7 @@
      * @returns {*} Returns the uncloned value or `undefined` to defer cloning to `_.cloneDeep`.
      */
     function customOmitClone(value) {
-      return isPlainObject(value) ? undefined : value;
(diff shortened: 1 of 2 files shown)


ccjmne:
I'm all for having the patched code pushed!  
But as the author of these commits, I'd prefer to keep having these authored by myself :)

You could just apply my commits, in your fork:

```shell
git remote add ccjmne https://github.com/ccjmne/lodash
git fetch ccjmne
git cherry-pick b26c7a4 5ae4e89
```

---

Is Lodash still being maintained? Do you need more maintainers?

I know it should be somewhat obsoleted by the newer versions of ECMAScript but... seeing the npm downloads still hovering at 40 MILLION per week, I think it's worth keeping in a working state a little longer!

PR #4665, opened 2020-02-18
Modified ES5 to ES6 functions.

HarishRMaryada:
Have modified the function to array function


falsyvalues:
Thanks for the PRs @HarishRMaryada!

In this case we prefer the more readable, longer syntax form.

PR #4666, opened 2020-02-19
Removed The Size Constraint

Sharron4me:
Removed The Size Constraint That Limits User From Entering Non- Negative Values.


bnjmnt4n:
Hi @Sharron4me! I think in this case passing negative values is an invalid use case, and the current behaviour should be maintained.

PR #4680, opened 2020-02-28
Update isNil.js documentation

sulhadin:
Since undefined == null returns true, this function is also returns true for undefined variable. So this example should be added to the documentation.


falsyvalues:
Hi @sulhadin!

We already cover that just line below with `void 0`.

PR #4686, opened 2020-03-08
suggesting a cycle of arguments to a method

9kartik:
Consider a button which can sort a list in ascending or descending order.
It would call for a function which will just need an extra parameter which keeps toggling like true/false etc.
Extend this to many other 'states' of the button like 'ascending'/'original'/'descending'.
So this is just a suggestion of a wrapper function which will keep on cycling the parameter list provided by the array in the first initialisation call.

Yet to write the test cases for the suggested methods.

### DIFF
--- a/cycle.js
+++ b/cycle.js
@@ -1,0 +1,53 @@
+/**
+ * Invokes `func` with an additional parameter which
+ * toggles on each call
+ *
+ * @since 
+ * @category Function
+ * @param {Function} func The function to execute.
+ * @param {Array} array The array to cycle over.
+ * @returns {Function} Returns the cycled execution wrapper, which upon execution
+ * executes func with the current extra argument.
+ * @example
+ *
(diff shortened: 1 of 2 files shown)


cycle.js
```diff
@@ -0,0 +1,53 @@
...
+        let nextIndex = cycler()
+        return array[nextIndex]
+    }
+    function runner(){
+        currentIndex = cycler()
+        return func.call(this, getCurrentValue(currentIndex))
+    }
+    runner.setDirection = setDirection
+    runner.getCurrentValue = getCurrentValue
+    runner.getNextValue = getNextValue
```
9kartik: This offers the currentValue of the counter.

falsyvalues:
Hi @9kartik!

We will keep an eye on [the popularity of the request](https://github.com/lodash/lodash/blob/master/.github/CONTRIBUTING.md#feature-requests).

PR #4689, opened 2020-03-11
chore: Remove extraneous check

### LINKED ISSUE #4683: Remove unnecessary check in getTag.js
This check is not required since return toString.call() will be enough to handle even `undefined` and `null`.
https://github.com/lodash/lodash/blob/588bf3e20db0ae039a822a14a8fa238c5b298e65/.internal/getTag.js#L11-L13

aashutoshrathi:
Fixes #4683 
Removing this check since:
image


princechauhan1992:
@falsyvalues Is there an issue with this PR? Why is it not merged yet?

PR #4518, opened 2019-10-16
Sanitize sourceURL so it cannot affect evaled code

alexbrasetvik:
Wherever it comes from, `sourceURL` should not be allowed to contain code that gets eval()-ed.

`/\s/` should cover all the bases as far as I know, including the various unicode-y ways of encoding newlines.

```
> sourceURL = '\u2028\u2029\nconsole.log("o hai")'
'  \nconsole.log("o hai")'
> eval('//# sourceURL=' + sourceURL)
o hai
undefined
> eval('//# sourceURL=' + sourceURL.replace(/\s/g, ' '))
undefined
```

PR #4719, opened 2020-04-07
Add Unicode line terminators for sourceURL sanitization

corvid19:
U+2028 and U+2029 were missing in #4355.


alexbrasetvik:
See https://github.com/lodash/lodash/pull/4518

PR #4739, opened 2020-04-27
fix: fix the jsdoc of SetCache#has

Keith-CY:
The return type should be `boolean` instead of `number`


falsyvalues:
Thank you @Keith-CY!

PR #4740, opened 2020-04-28
use shorthand in _.get() return

itxtoledo:


bnjmnt4n:
Hi @itxtoledo! Using `||` would lead to unexpected errors as the result might be a falsy value (eg. `false` or `0`). This would cause the default value to be returned even though there is a result at the specified path.

PR #4756, opened 2020-05-06
add Checks if `value` is classified as a `BigInt` primitive.

xgqfrms:
add `isBigInt`


isBigInt.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * isBigInt(BigInt('1'))
+ * // => true
+ *
+ * isBigInt(123)
+ * // => false
+ */
+function isBigInt(value) {
+  const type = typeof value
+  return type === 'bigint'
```
aashutoshrathi, line 21: Should we use our inbuilt files like getTag.
And use type using it, like this:
```js
return getTag(value) === '[object BigInt]'
```

xgqfrms: 👍 OK, I will modify it.

isBigInt.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * isBigInt(BigInt('1'))
+ * // => true
+ *
+ * isBigInt(123)
+ * // => false
+ */
+
+function isBigInt(value) {
+  const tag = getTag(value)
```
aashutoshrathi, line 23: Not sure if this variable is needed.
Rest LGTM ✅

xgqfrms: Maybe, you are right!
I just want to reuse the `tag` for the future extension.

xgqfrms: Thanks for your code review.

falsyvalues:
Hi @xgqfrms!

Please clean up commits.

xgqfrms:
@falsyvalues 

Sorry, I'm not sure what should be clean.

It's that mean to my previous commits history？

aashutoshrathi:
I think you need to squash up your commits, specifically the ones concerning `label.yml` and `README.md` as there is no change in them in the end.

Issue #4747, opened 2020-05-01
Examples with indices should not use integers are data


geoynomous:
In contrast to 
```
_.lastIndexOf([1, 2, 1, 2], 2);
// => 3
 
// Search from the `fromIndex`.
_.lastIndexOf([1, 2, 1, 2], 2, 2);
// => 1
```
this has a better human readability to understand the index example
```
_.lastIndexOf(['a', 'b', 'a', 'b'], 'a');
// => 3
 
// Search from the `fromIndex`.
_.lastIndexOf(['a', 'b', 'a', 'b'], 'b', 2);
// => 1
```

Please check all such examples

dsiah:
I've opened a PR to update such examples. Waiting for feedback 😬

geoynomous:
Great ... I think you've covered it all ...thanks!

PR #4761, opened 2020-05-08
Updating documentation index examples #4747

dsiah:
Opening a PR to update the documentation regarding index methods regarding issue #4747.


geoynomous:
about the 'sorted' - is the array to be expected to be sorted? Then maybe the strings with 'red', 'blue', 'green' need another order ...

dsiah:
😂 gotcha, replaced with some sorted lists of non-integer values

PR #4779, opened 2020-05-14
Create isBigInt.js

xgqfrms:
add Checks if `value` is classified as a `BigInt` primitive.

### DIFF
--- a/isBigInt.js
+++ b/isBigInt.js
@@ -1,0 +1,26 @@
+import getTag from './.internal/getTag.js'
+
+/**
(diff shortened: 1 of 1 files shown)


xgqfrms:
# ES6 versions

```js
// ==
const isBigInt = (value) => getTag(value) == '[object BigInt]'

export default isBigInt
```


```js
// ===
const isBigInt = (value) => getTag(value) === '[object BigInt]'

export default isBigInt
```

```js
// value
const isBigInt = value => getTag(value) === '[object BigInt]'

export default isBigInt
```

PR #4788, opened 2020-05-25
Simplify isPlainObject

Terry-Su:
Recently I'm learning how to write a `isPlainObject` function, and find following codes really confusing:
```js
  let proto = value
  while (Object.getPrototypeOf(proto) !== null) {
    proto = Object.getPrototypeOf(proto)
  }
  return Object.getPrototypeOf(value) === proto
```

As mentioned in feature explanation:
> check if `value` is an object created by **`Object` constructor or one with a `[[Prototype]]` of `null`**

Why we have to "while loop" the  `Object.getPrototypeOf(proto) `, it equals with:
```js
const proto = Object.getPrototypeOf(value)
return proto === null || Object.getPrototypeOf(proto) === null
```
So I propose this update or just make a suggestion to make codes more simplified and understandable.

### DIFF
--- a/isPlainObject.js
+++ b/isPlainObject.js
@@ -31,14 +31,8 @@ function isPlainObject(value) {
   if (!isObjectLike(value) || getTag(value) != '[object Object]') {
     return false
   }
-  if (Object.getPrototypeOf(value) === null) {
-    return true
-  }
-  let proto = value
-  while (Object.getPrototypeOf(proto) !== null) {
-    proto = Object.getPrototypeOf(proto)
-  }
-  return Object.getPrototypeOf(value) === proto
+  const proto = Object.getPrototypeOf(proto)
+  return proto === null || Object.getPrototypeOf(proto) === null
 }
 
 export default isPlainObject
--- a/isPlainObject.js
+++ b/isPlainObject.js
@@ -31,7 +31,7 @@ function isPlainObject(value) {
   if (!isObjectLike(value) || getTag(value) != '[object Object]') {
     return false
   }
-  const proto = Object.getPrototypeOf(proto)
+  const proto = Object.getPrototypeOf(value)
   return proto === null || Object.getPrototypeOf(proto) === null
 }
 


isPlainObject.js
```diff
@@ -31,14 +31,8 @@ function isPlainObject(value) {
...
   }
-  if (Object.getPrototypeOf(value) === null) {
-    return true
-  }
-  let proto = value
-  while (Object.getPrototypeOf(proto) !== null) {
-    proto = Object.getPrototypeOf(proto)
-  }
-  return Object.getPrototypeOf(value) === proto
+  const proto = Object.getPrototypeOf(proto)
```
ConnorJamesLow, line 34: That will break _(Cannot access 'proto' before initialization)_. Did you mean `const proto = Object.getPrototypeOf(value)`?

jdalton:
I've seen this simplification in other code bases too. I'll think on it.

gnprice:
This simplification looks correct to me. (I was just looking at this code and thinking the same thing, and came here to see if there was already a PR for it.) It'd be good to have it merged, because it makes this function both more efficient and much simpler to understand.

Here's one way of thinking about why the existing code has the same behavior as the new code in this PR. The existing code says this:
```js
  if (Object.getPrototypeOf(value) === null) {
    return true
  }
  let proto = value		
  while (Object.getPrototypeOf(proto) !== null) {		
    proto = Object.getPrototypeOf(proto)		
  }		
  return Object.getPrototypeOf(value) === proto
```
The point is that the expression in that `return` at the end is asking precisely the question: did the loop body run exactly one time? And so the two `return` statements together are asking: would that loop run for either zero or one times?

Which is exactly what the new code says:
```js
  const proto = Object.getPrototypeOf(value)
  return proto === null || Object.getPrototypeOf(proto) === null
```

To see that the `return` at the end of the existing code is asking precisely whether the loop body ran exactly once, the main point is:
* On the one hand, after the loop body runs the first time, that condition `Object.getPrototypeOf(value) === proto` is true.
  * (It always runs at least once, because the first `return` means we don't get this far if it would run zero times.)
* If the condition `Object.getPrototypeOf(value) === proto` is ever true again, then that later value of `proto` is `===`-equal to the one we had after the first iteration.
  * That'd be a cycle in the prototype chain. I think that's forbidden by the spec.
  * But even if it can happen, it'd mean this is an infinite loop, and we'd never reach the final `return`.
* So if we get to the final `return` and the condition is true, we must have run the loop body exactly once.

---

That reasoning may look a little complicated. But all the complication is really from the existing code: the new code is much simpler and easier to understand. Merging this would mean that nobody else has to work through that reasoning again. :slightly_smiling_face:

Issue #4807, opened 2020-06-09
_.escapeRegExp does not escape '-'


sirasistant:
The character '-' determines ranges of characters in regexp. 
Not escaping it can generate some regexp to fail when including escaped characters, since it can make RegExp think that it's representing a range.
Example given:
```
const allowedChars = _.escapeRegExp('+-=');
const wrongRegex = new RegExp(`^[${allowedChars}]*$`);
console.log('0'.match(wrongRegex)); // Should not match, but it does, because 0 is between + and =
console.log('0'.match(new RegExp(/^[\+\-=]$/))); // expected escaping, no match
```

falsyvalues:
Closing as duplicate of #4760

PR #4808, opened 2020-06-09
Escape the character '-' in _.escapeRegExp.js

sirasistant:
This pull request makes _.escapeRegExp to also escape `-`, the RegExp range character. More information about this in the issue https://github.com/lodash/lodash/issues/4807

### DIFF
--- a/escapeRegExp.js
+++ b/escapeRegExp.js
@@ -2,7 +2,7 @@
  * Used to match `RegExp`
  * [syntax characters](http://ecma-international.org/ecma-262/7.0/#sec-patterns).
  */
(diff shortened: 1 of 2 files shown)


AmyShackles:
Wouldn't this make it so that you would be unable to use ranges in your character set when using escapeRegExp?

sirasistant:
Yes, since this function escapes regexp logic, it should also escape ranges IMHO, since they are regexp special characters.

sirasistant:
(you could always construct your character-set regexp like this)
```
new RegExp(`^[a-zA-Z\\d${escapeRegExp('^$*.[]{}()?"!@#%&/\\,><\':;|_~`-')}]*$`)
```

Issue #4822, opened 2020-06-17
merge() doesn't throw any Error when source object is frozen


sheff146:
### Description
I'm calling **merge()** on a frozen object.
```js
const frozen = Object.freeze({ a: "1" });
merge(frozen, { a: "2", b: "123" });
```

### Expected behavior
I would expect **merge()** to throw an **Error** in this case.

### Actual behaviour
**merge()** silently completes leaving the destination object without change.

Codesadbox illustrating the issue: https://codesandbox.io/s/lodash-merge-frozen-hos4h?file=/src/index.js.

gagyibenedek:
If no one is working on this I'll try my luck wit it.

bnjmnt4n:
This behaviour is intentional, as lodash attempts to avoid throwing errors as much as possible. For the same reason, passing a `null` or `undefined` value to `_.merge` also does not throw an error.

lukeapage:
lodash code is not in strict mode.

If you use lodash-es it works.

PR #4824, opened 2020-06-19
[issue-4822] - merge() doesn't throw any Error when object is frozen

gagyibenedek:
Fix for #4824. 
Added extra check to make sure an error is thrown.
Added a test case too and run it manually (since `npm test` is broken).

### DIFF
--- a/merge.js
+++ b/merge.js
@@ -10,7 +10,7 @@ import createAssigner from './.internal/createAssigner.js'
  * assignment. Source objects are applied from left to right. Subsequent
  * sources overwrite property assignments of previous sources.
  *
- * **Note:** This method mutates `object`.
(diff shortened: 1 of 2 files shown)


Doatti:
You tagged this PR instead of the issue. The correct issue number is #4822.

Doatti:
Maybe the `isFrozen` check should be added to the internal `createAssigner` function instead `merge`. That would fix the same issue with functions like `_.assign`, `_.assignIn` etc.

mhombach:
@gagyibenedek Are you still working on this? :)

bnjmnt4n:
See https://github.com/lodash/lodash/issues/4822#issuecomment-748752508.

Issue #4825, opened 2020-06-20
lodash-es modules should be published to NPM with "type": "module"


xiaoxiangmoe:
We should add `"type": "module"` in `pacakge.json` in lodash-es.

Otherwise, downstream users cannot easily use lodash-es in mjs.

bnjmnt4n:
Closing this for #4800.

PR #4826, opened 2020-06-20
feat: add type module for esm

xiaoxiangmoe:
close https://github.com/lodash/lodash/issues/4825

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "bugs": "https://github.com/lodash/lodash-cli/issues",
   "repository": "lodash/lodash",
   "license": "MIT",
+  "type": "module",
   "jsnext:main": "lodash.js",
   "main": "lodash.js",
   "module": "lodash.js",


package.json
```diff
@@ -7,6 +7,7 @@
   "bugs": "https://github.com/lodash/lodash-cli/issues",
   "repository": "lodash/lodash",
   "license": "MIT",
+  "type": "module",
```
JLHwung, line 10: Note that this change will break on npm <=3.9.5 (bundled in node <= 6.2.2).

C.f. https://github.com/babel/babel/issues/11000#issuecomment-573807377

damianobarbati:
@xiaoxiangmoe isn't the `export` field with the available packages needed as well? 
Otherwise you will always have to import the full lodash and destructure later instead of named-import specific functions.

xiaoxiangmoe:
> You will always have to import the full lodash and destructure later instead of named-import specific functions.

```js
import {identity} from 'lodash';
```

In most cases, importing specific`lodash/identity` is just for tree shaking, and esm is static and has good tree shaking support.

package.json
nickmccurdy: That's been unsupported for years hasn't it?

JLHwung: lodash requires `"node": ">=4.0.0"`, shipping this change implies that we should bump `"node"` because `npm <= 3.9.5` will throw on `"type"` properties in `package.json`.

We could specify `"npm": "> 3.9.5"`, too.

nickmccurdy: npm 3.9.5 is very old. It should be fine to raise the dependency on it in a breaking change, then use the next major for the long-term full ESM conversion

rifler: As I understand, this change must be applied only in `lodash-es` package. Otherwise it can break usual `lodash` package, because main nodejs program will think, that `lodash` is esm package and will throw error `require is not defined` or something like this

rifler: Or `lodash` can be converted to dual commonjss/esm package:
https://nodejs.org/dist/latest-v14.x/docs/api/esm.html#esm_dual_commonjs_es_module_packages

damianobarbati: @rifler or simpler, `lodash` could use semantic versioning as everybody else 🤷🏻‍♂️
If this release brings breaking changes then bump the major (possibly adjusting `engines` accordingly).

bnjmnt4n:
I've made this change in lodash-es v4.17.20. Thanks for all the feedback!

PR #4835, opened 2020-07-01
perf: Improved performance of flow, trim, trimStart, trimEnd

whiteand:
1. Added handling of empty list of funcs with reusage of identity function (that was a results of each separated call to flow() without parameters

2. Added usage of dictionary in charsEndIndex and charsStartIndex in order to check if char is in chrSymbols with O(1) complexity

### DIFF
--- a/.internal/charsEndIndex.js
+++ b/.internal/charsEndIndex.js
@@ -1,5 +1,3 @@
-import baseIndexOf from './baseIndexOf.js'
-
 /**
  * Used by `trim` and `trimEnd` to get the index of the last string symbol
  * that is not found in the character symbols.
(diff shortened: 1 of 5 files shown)


.internal/charsEndIndex.js
```diff
@@ -10,9 +8,14 @@ import baseIndexOf from './baseIndexOf.js'
  * @returns {number} Returns the index of the last unmatched string symbol.
  */
 function charsEndIndex(strSymbols, chrSymbols) {
-  let index = strSymbols.length
+  const chrSymbolsDict = {}
+  let index = chrSymbols.length
+  while (index--) {
+    chrSymbolsDict[chrSymbols[index]] = true
```
Toxicable, line 14: flyby review: if you're doing this for performance then use a `Map` instead of an object, it's much faster

whiteand: @Toxicable Look at performance comparison: https://jsperf.com/set-vs-obj-comparison/1

Toxicable: sure set works also :P

whiteand: > sure set works also :P

I used an object instead of Set(or Map) because Set is not implemented in old browsers. And for me it's not clear that in this case Set(or Map) will be a better way.

whiteand: > sure set works also :P

So what should I do?

bnjmnt4n:
This seems to be verging on the edge of micro-optimizations here. We'd rather prefer simpler code over optimizing for uncommon cases. The `trim*` functions are typically called with a small string of `chrSymbols`, so performing a dictionary lookup seems pretty unnecessary. Additionally, `flow` is typically called with more than 1 function, so there is no need to optimize for the unlikely cases.

PR #4871, opened 2020-07-22
Added duplicates method

VigneshMurugan:
Duplicates method is to find out duplicate values in an array

### DIFF
--- a/duplicates.js
+++ b/duplicates.js
@@ -1,0 +1,25 @@
+import uniq from './uniq';
+import difference from './difference';
+
+
+/**
+ * Creates a duplicate version of an array, using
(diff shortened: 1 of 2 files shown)


duplicates.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {Array} array The array to inspect.
+ * @returns {Array} Returns the new duplicates array.
+ * @example
+ *
+ * duplicates([2, 1, 2])
+ * // => [2]
+ */
+function duplicates(array) {
+    return (array != null && array.length)
+      ? difference(array, uniq(array))
```
lonewarrior556, line 21: lodash difference is a set operation.  This will always be empty
```js
function duplicates(array) {
    return (array != null && array.length)
      ? _.difference(array, _.uniq(array)) 
      : []
  }

duplicates([2, 1, 2])
// => []
```

Issue #4865, opened 2020-07-18
Expand _.pullAt  to take negative index 


lonewarrior556:
Make `_.pullAt` consistent with `_.nth`
```js
var array = ['a', 'b', 'c', 'd'];
 
_.nth(array, 1);
// => 'b'
 
_.nth(array, -2);
// => 'c';

/* Current Behavior of pullAt*/
_.pullAt(array,[1,-2]) 
// => ["b", undefined]

/* Should be */
_.pullAt(array,[1,-2]) 
// => ["b", "c"]
```

This would make finding the Bounds of an array simpler
```js
_.pullAt(_.sortBy([9,1,8,2,6,4,9,11,1,0,50,1,2,3,4]), 0, -1)
// => [0, 50]
```

bnjmnt4n:
`_.at` is an Object method, meaning it is designed to work on objects primarily. Adding this special handling of negative numbers will does not seem like a good idea for `_.at`, and for consistency, we might also not want to change `_.pullAt`.

lonewarrior556:
@bnjmnt4n updated request to only modify Array: `_.pullAt`

`_.pullAt` is already inconsistent with `_.at` as it mutates the passed array,

PR #4889, opened 2020-08-07
added negative index support for at. Fix for #4865

dopecodez:
Added negative index support for `at` as per #4865 . I could do the same for `pullAt` but cannot get the test cases to run for `pullAt`.

### DIFF
--- a/.internal/baseAt.js
+++ b/.internal/baseAt.js
@@ -1,4 +1,5 @@
 import get from '../get.js'
+import isNumber from '../isNumber'
 
 /**
  * The base implementation of `at` without support for individual paths.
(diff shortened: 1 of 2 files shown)


lonewarrior556:
There was a good comment made on the request that `_.at` is an object method; 
assuming you have this
```
var arr = ['zero', 'one'];
arr['-1'] = 'negative one';

what should 
```js
_.at(arr, -1);
```
return?

As such I changed #4865 to only modify `pullAt`

dopecodez:
I agree with the feature, but I dont think loadash is actively maintained these days.

I'll be happy to close this if you can accomodate the _.at change present in this branch to #4865 .

PR #4903, opened 2020-08-18
catching undefined and null meanBy

gondolin24:
meanBy only checks for nulls and not undefined.
This PR will fix this and handle both cases

### DIFF
--- a/meanBy.js
+++ b/meanBy.js
@@ -21,7 +21,7 @@ const NAN = 0 / 0
  * // => 5
  */
 function meanBy(array, iteratee) {
-  const length = array == null ? 0 : array.length
+  const length = (array) ? 0 : array.length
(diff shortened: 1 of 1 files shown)


spmartin823:
`(undefined == null) === true`. Your change will incorrectly check for false cases.

meanBy.js
```diff
@@ -21,7 +21,7 @@ const NAN = 0 / 0
  * // => 5
  */
 function meanBy(array, iteratee) {
-  const length = array == null ? 0 : array.length
+  const length = (array) ? 0 : array.length
```
aminya, line 24:
```suggestion
  const length = !array ? 0 : array.length
```

bnjmnt4n:
Hi @gondolin24! Using the loose equal `==` means that both null and undefined are already caught.

PR #4905, opened 2020-08-18
Optimization: avoid checking null proto twice

matthieusieben:
When the loop is run for the first time, we already know that the protoype is not null (because of the if before). We can thus make one fewer evaluation of `Object.getPrototypeOf(proto) !== null ` by performing a `do while` loop

### DIFF
--- a/isPlainObject.js
+++ b/isPlainObject.js
@@ -35,9 +35,9 @@ function isPlainObject(value) {
     return true
   }
   let proto = value
-  while (Object.getPrototypeOf(proto) !== null) {
+  do {
     proto = Object.getPrototypeOf(proto)
-  }
+  } while (Object.getPrototypeOf(proto) !== null)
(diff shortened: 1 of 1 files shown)


isPlainObject.js
```diff
@@ -35,9 +35,9 @@ function isPlainObject(value) {
     return true
   }
   let proto = value
-  while (Object.getPrototypeOf(proto) !== null) {
+  do {
     proto = Object.getPrototypeOf(proto)
-  }
+  } while (Object.getPrototypeOf(proto) !== null)
```
aminya, line 40: You are changing the while head check here. In the very first check, `Object.getPrototypeOf(proto)` is checked to not be `null`. But in your code, you don't check if the resulting `proto = Object.getPrototypeOf(proto)` is `null` or not.

PR #4911, opened 2020-08-27
update shuffle.js

kafelix496:
Hello,
I try to update shuffle.js file in case of shuffle string.
Please review this.
Thank you!

### DIFF
--- a/shuffle.js
+++ b/shuffle.js
@@ -1,31 +1,50 @@
 import copyArray from './.internal/copyArray.js'
+import isString from './isString.js'
+import isArrayLike from './isArrayLike.js'
 
 /**
  * Creates an array of shuffled values, using a version of the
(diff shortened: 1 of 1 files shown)


bnjmnt4n:
I'd like to avoid adding extra paths for maintenance. It's simple enough to convert a string to an array before calling shuffle and converting it back. In fact, the current implementation already shuffles a string and returns an array of characters (due to strings having a `length` property and being indexable just like arrays).

kafelix496:
> I'd like to avoid adding extra paths for maintenance. It's simple enough to convert a string to an array before calling shuffle and converting it back. In fact, the current implementation already shuffles a string and returns an array of characters (due to strings having a `length` property and being indexable just like arrays).

Okay, thank you very much for this comment!

Issue #4927, opened 2020-09-13
isArrayLike returns true for objects containing 'length' property


abhinav-sachdeva:
The method when passed a regular object with property named 'length' returns true. Does not accurately check if the argument is an array like iterable.

`_.isArrayLike({ length: 1 }); // true`

https://github.com/lodash/lodash/blob/86a852fe763935bb64c12589df5391fd7d3bb14d/isArrayLike.js#L27

happy257:
This also fixes issue #4921

Gaurav2048:
Documentation defines array like as **_Checks if value is array-like. A value is considered array-like if it's not a function and has a value.length that's an integer greater than or equal to 0 and less than or equal to Number.MAX_SAFE_INTEGER._** 

https://lodash.com/docs/4.17.15#isArrayLike

To check if array or not use Array.isArray(arr) function.

abhinav-sachdeva:
That's exactly the issue, isArrayLike should return true for arrays and strings too but it's returning true for a regular object with length property.

Gaurav2048:
But the doc defines **arrayLike** as something and everything but not a function.

freakymittal:
@abhinav-sachdeva `ArrayLike` and `Array` are two different things. `ArrayLike` has a proper definition according to standards, you can go through https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from and see that Array.from takes `arraylike` entity as input. The major difference between `array` and `arraylike` is that arraylike doesn't have to have the iterable properties e.g. when you get dom elements using document.getElementsbyClassName the resulting entity is not array but arraylike which you can convert to array using Array.from.

And just in case you do want to check if the entity is an array you can use something like `Array.isArray([1,2,3])`

freakymittal:
@abhinav-sachdeva I totally get your point, but they (lodash's contributors) can't change the definition of a thing which is already widely accepted.

abhinav-sachdeva:
Does that mean this is the expected behaviour then?

FelixRilling:
Yes, according to <https://tc39.es/ecma262/#sec-lengthofarraylike>
> The abstract operation LengthOfArrayLike takes argument obj. It returns the value of the "length" property of an array-like object. It performs the following steps when called:

>  1) Assert: Type(obj) is Object.
>  2) Return ? ToLength(? Get(obj, "length")).

> An array-like object is any object for which this operation returns an integer rather than an abrupt completion.

[...]

PR #4928, opened 2020-09-13
Fix: isArrayLIke returns incorrectly for {length: 1}

abhinav-sachdeva:
isArrayLike returns true for objects containing 'length' property
This PR fixes the issue https://github.com/lodash/lodash/issues/4927


abhinav-sachdeva:
CLA is signed and now in read-only mode.

abhinav-sachdeva:
Fix not needed

PR #4935, opened 2020-09-19
feat: add isStrictObject for strict object boolean checking

umairx97:
### add `isStrictObject` function 

```javascript 
 isStrictObject([200, 100, 250, 150])
 // => false
 
 isStrictObject({})
 // => true
```


bnjmnt4n:
Hi! This performs the same functionality as [`_.isPlainObject`](https://lodash.com/docs/#isPlainObject).

PR #4944, opened 2020-09-30
Adding get decimal part (of number) function

ajomuch92:
This function receives a number or string and gets the decimal part. Return null if it's not a number


bnjmnt4n:
Hi @ajomuch92! I believe this function is too specialized to fit into lodash. Feel free to make use of some of the more specialized string and number handling utility libraries.

PR #4949, opened 2020-10-05
add recursiveMap function

iiPanda:
I'm not sure if i've done everything properly (documented etc.), so please let me know if not.

### DIFF
--- a/recursiveMap.js
+++ b/recursiveMap.js
@@ -1,0 +1,35 @@
+/**
+ * Creates an array of values by running each element of `array` thru `iteratee`, if
+ * the element is an array it runs recursiveMap on it.
+ * The iteratee is invoked with one argument: (value).
+ *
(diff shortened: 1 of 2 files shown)


ZimGil:
Can't you just do the same with the existing `map` like so:

```javascript
function recursiveSquare(value) { return _.isArray(value) ? _.map(value, recursiveSquare) : value * value; }

const foo = [2, 4, [6, 8]];
recursiveSquare(foo)
// [4, 16, [36, 64]]
```

bnjmnt4n:
Hi @iiPanda! This seems like a pretty specific use case, which is easily composable with the existing lodash functions, so we'll defer on adding this function for now. We'll keep an eye on the [popularity of the request](https://github.com/lodash/lodash/blob/master/.github/CONTRIBUTING.md#feature-requests).

Issue #4971, opened 2020-10-31
_.sum does not return correct output for [null]


addnab:
`_.sum([1, null, 0])` and `_.sum([null, null])` returns expected values while `_.sum([null])` returns `null`.

```js
> _.sum([])
0
> _.sum()
0
> _.sum('a')
'a'
> _.sum(null)
0
> _.sum([1, null, 0])
1
> _.sum([null])
null <--- should be 0
> _.sum([null, null])
0
```

Semigradsky:
You really think that it is a correct `sum` usage?

radekmie:
Same goes for `false` (and other falsy values).

bnjmnt4n:
Marking this as a duplicate of #4971.

PR #4986, opened 2020-11-25
Optimization: _.sum does not return correct output for [null] #4971

lzj822:
_.sum([null]) and _sum([undefined]) returns expected 0


leqwasd:
Was  this fixed?
I can still reproduce this issue now.
https://playcode.io/2157306
```js
console.log(_.sum([])); // 0
console.log(_.sum([null])); // null
console.log(_.sum([null, null])); // 0
```

PR #4987, opened 2020-11-26
Optimization: _.sum does not return correct output for [null] #4971

lzj822:
_.sum([null]) or _sum([undefined]) returns expected 0


Wrathcoin:
Tooshaay

PR #4990, opened 2020-11-28
FIXED _.size,_.isEmpty behavoir on ArrayBuffer and SharedArrayBuffer

geareo:
FIXED the behavior of _.size and _.isEmpty methods on ArrayBuffer and SharedArrayBuffer instances


keyhan-haqiq:
why dont you accept ?

PR #5013, opened 2020-12-17
fix(debounce):when use event properties, will be null

aminwangaa:


bnjmnt4n:
We can't accept this fix, as `debounce` is a general purpose function which might not just be used for React events.

Issue #4918, opened 2020-09-04
Intersects return boolean


iwhoppock:
Would love this to be simplified to `_.intersects(a1, a2) -> Boolean`:
```
function intersects(a1, a2) {
  return !_.isEmpty(_.intersection(a1, a2));
}
```

bnjmnt4n:
Given that this request is simply just checking the length of the result of `_.intersection`, it is simple enough to remain outside of lodash.

szimek:
It would be useful if there was such function that returns early, without calculating the whole intersection of 2 arrays.

So something like:
```js
const intersects = a.some(x => b.includes(x));
```
rather than
```js
const intersection = a.filter(x => b.includes(x));
```

PR #5015, opened 2020-12-17
Issue #4918 fixed, intersects function simplified

SepehrAsh:
Issue #4918 fixed, intersects function simplified and test added.


bnjmnt4n:
See #4918.

Issue #4763, opened 2020-05-09
The proposal of add '_.until' function


alperen:
Sometimes, we would need to call a function more than once to get or generate our expected value. That functions have side-effects that such as changing by the time, API calls or full of randomized values. In this case, we should wrap our function with another function. This wrap functions should call our actual function until return the value on our expecting way. In examples that will you see in the below, we call that reference function as the barrier function.

## Sketch
![image](https://user-images.githubusercontent.com/3244438/81487496-1089b080-9266-11ea-99d6-c16cc2528115.png)

## Syntax:
```js
_.until(actualFunction, [...arguments], barrierFunction, {retry: Number}?);
```

## Glossary:

- **The actual function**  is the function that could we want to call it more than once.
- **Arguments**: The arguments array exists for the pass to data to our actual function.
- **The barrier function** is the reference function to stop the iteration and return the last generated value. That function should return boolean values. If it returns true the iteration will break and `._until` the function should return the last value of the actual function.
- **options**: There is just one option argument on my mind at the moment. The `retry` value defines the maximum iteration count of the `._until` function. If we can't find the expected value (based to the barrier function) before exceed the retry number, we should the last returned value immediately. This saves us to enter endless loops.

### Example 1
```js
const randomizedNumber = _.until(_.random, [1, 5], (number) => number % 2);
```
In this example, `randomizedNumber` variable is absolutely even number. The `._until` function will iterate the actual function until the barrier broke.

### Example 2
```js
const randomizedNumber = _.until(_.random, [1, 5], (number) => number % 2, {retry: 3});
```
In this example, `randomizedNumber` variable could be even or odd number. The `._until` function will iterate 3 times maximum then return the last value without test the data with barrier function.

ncjones:
You can use [retry-assert](https://www.npmjs.com/package/retry-assert). IMO this is the best solution out there for this problem in the Nodejs ecosystem. I've been using it for a few years now with all my Cucumber JS test suites. Disclaimer: I'm the author.

bnjmnt4n:
This seems like a very specific use case, and there seems to already be a well-defined library for that. Hence, I'm closing this issue for now as we'd prefer to keep this method outside of lodash but we'll keep an eye on the [popularity of the request](https://github.com/lodash/lodash/blob/master/.github/CONTRIBUTING.md#feature-requests).

PR #5018, opened 2020-12-17
Feature: implement until function (#4763)

ahmadsalimi:
Implementation of #4763


bnjmnt4n:
See https://github.com/lodash/lodash/issues/4763#issuecomment-748624113. We'd rather defer from adding this into lodash for now.

Issue #4831, opened 2020-06-23
_.group(array, amount)


Richienb:
Like chunk but takes in the number of chunks to return to evenly distribute the items over.

```js
const _ = require("lodash")

_.group([1, 2, 3, 4, 5, 6], 3)
//=> [[1, 2], [3, 4], [5, 6]]

_.group([1, 2, 3, 4, 5], 2)
//=> [[1, 2, 3], [4, 5]]
```

Doatti:
I can't imagine this being used all that much. Either way, it's quite simple to implement using `_.chunk`:
```javascript
function group(array, amount) {
    return _.chunk(array, Math.ceil(array.length/amount));
}
```

bnjmnt4n:
Given that this is simple enough to compose from _.chunk with just a minimal addition, I'm closing this issue for now. That being said, we'll look at the [popularity of the request](https://github.com/lodash/lodash/blob/master/.github/CONTRIBUTING.md#feature-requests).

matthewvalentine:
The problem is subtler than you'd expect, and so the solution @Doatti posted doesn't work and in fact `chunk` cannot be used to directly solve it. For example:
```javascript
f = (xs, a) => _.chunk(xs, Math.ceil(xs.length/a))

// Wrong number of groups
f([1,2,3,4,5,6], 4)
[ [ 1, 2 ], [ 3, 4 ], [ 5, 6 ] ]

// Groups are not evenly sized
f([1,2,3,4,5,6,7,8,9,10], 4)
[ [ 1, 2, 3 ], [ 4, 5, 6 ], [ 7, 8, 9 ], [ 10 ] ]
```
Sometimes the answer requires a few of the buckets to be one size and the rest to be another. Because `chunk` always returns buckets of the same size (except maybe the last), it can't directly accomplish that.

PR #5020, opened 2020-12-18
group function - evenly distributes items over groups

ArmanZarei:
[#4831](https://github.com/lodash/lodash/issues/4831)


bnjmnt4n:
See https://github.com/lodash/lodash/issues/4831#issuecomment-748997803.

Issue #4611, opened 2020-01-05
Add an _.assignOnly method [Feature request][Code proposal]


danielsum:
# Description of the expected behavior

This method will be like an Object.assign except that it accept only 1 source of copy and have instead a third parameter that will filter only the properties that needs to be assigned. If the third parameter is empty, by default it will only assign the properties of the first parameter object.

# Use cases

This could be really useful when you need to copy only particular properties and prevent to copy other properties and avoid like that "prototype like pollution".

```
function Foo() {
   this.a = 1;
   this.b = 1
}
 
function Bar() {
  this.a = 2;
  this.b = 2;
  this.c = 2;
}
 
Foo = _.assignOnly(Foo, Bar, [a]);
// => Foo = { 'a': 2, 'b': 1 }

Foo = _.assignOnly(Foo, Bar, [a,c]);
// => Foo = { 'a': 2, 'b': 1 ,"c":2}

Foo = _.assignOnly(Foo, Bar);
// => Foo = { 'a': 2, 'b': 2}
```

# Code proposal : 

```
var pick = require('./pick');

/**
 * This method is like `_.assign` except that the third parameter is an array of properties to copy or if is null it will assign only the properties of the first object parameter.
 *
 * @static
 * @since xxxx
 * @memberOf _
 * @category Object
 * @param {Object} object The source object.
 * @param {Object} objectToCopy The object to copy
 * @param {Array} arrayKeysToCopy An array of properties keys to copy
 * @returns {Object} Returns the new object.
 * @example
 *
 * var object = { a: 1, b: 1 };
 * var object2 = { a: 2, b: 2, c: 2 };
 *
 * _.assignOnly(object, object2);
 * // => {a: 2, b:2}
 *
 * _.assignOnly(object, object2; [a,c]);
 * // => {a: 2, b:1,c:2}
 */
var assignOnly = function(object, objectToCopy, arrayKeysToCopy) {

  if (!arrayKeysToCopy) {
    arrayKeysToCopy = Object.keys(object);
  }

  return Object.assign(object, _.pick(objectToCopy, arrayKeysToCopy));
}

module.exports = assignOnly;
```

bnjmnt4n:
Since this is a simple combination of _.assign and _.pick, and can be easily composed with a one liner, I'd like to defer from adding this to lodash.

danielsum:
I think it's much more than 2 simple combinations, actually 3 because most of the times you have to write : 

[...]

PR #5024, opened 2020-12-19
Assign only

baharkhd:
added asignOnly function to clone an object with specific keys.


bnjmnt4n:
See https://github.com/lodash/lodash/issues/4611#issuecomment-748923791.

PR #5037, opened 2020-12-22
Fix trim methods examples

felipecrs:


felipecrs:
Okay, I just noticed that the examples were actually right. But why? What's the reason, does anyone knows?

felipecrs:
Ok, got it. The trim methods remove all the characters, not the substring.

PR #5055, opened 2021-01-15
Add increment function

tanrax:


dec.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @since 4.17.15
+ * @category Math
+ * @param {number} number The number to decrease.
+ * @param {number} number Decrease.
+ * @returns {number} Returns the decrease.
+ * @example
+ *
+ * dec(4)
+ * // => 5
```
David3141, line 14: I randomly stumbled across this here 😄 I think you meant 3? 🙂 
```suggestion
 * // => 3
```

PR #5075, opened 2021-02-10
perf: more concise (and slightly more performant on larger arrays) version of the cmpact-method

ygorbunkov:
More succinct version of the `compact`-method with slightly better performance for larger arrays is suggested.

### DIFF
--- a/compact.js
+++ b/compact.js
@@ -11,20 +11,8 @@
  * compact([0, 1, false, 2, '', 3])
  * // => [1, 2, 3]
  */
-function compact(array) {
-  let resIndex = 0
-  const result = []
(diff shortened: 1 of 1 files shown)


huyennbl:
I think your implementation is different from the current `_.compact`. So I don't think it would be able to merge as people are using this function widely IMO.

```js
_.compact(null)
// => []

_.yourCompact(null) 
// => null
```

ygorbunkov:
@huyennbl, nice catch. It passes unit tests, though. I, will, probably, commit slight fix to address that issue along with some unit tests update. Thanks for your comment.

PR #5102, opened 2021-02-26
New feature: capitalizeWord

huyennbl:
Returns a new string with all the words that match 2nd capitalized. 
Partially matches are ignore (e.g.:  capitalizeWord('github', 'git') will still return 'github')

```js
capitalizeWord('frank talks about it frankly', 'frank');
// => 'Frank talks about it frankly'
capitalizeWord('Lorem ipsum...', 'ipsum')
// => 'Lorem Ipsum...'
```

Use case: Capitalize a person's name in a text

### DIFF
--- a/capitalizeWord.js
+++ b/capitalizeWord.js
@@ -1,0 +1,27 @@
+import capitalize from './capitalize'
+
+function asWord(word) {
+  return `\\b${word}\\b`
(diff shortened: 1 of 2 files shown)


alesmit (approved):
LGTM

PR #5118, opened 2021-03-12
perf: use `===` instead of `==`

jacob-lcs:


jdalton:
Thank you @jabclab-lcs! It's like a super micro perf but over the years I've come to favor explicit over implicit. So I'm okay with this or other `==` to `===` cases.

Issue #5131, opened 2021-04-06
isEmpty(new Date) returns true instead of false


dellyoung:
`_.isEmpty(new Date)`

Actual returns:

`true`

Expected:

`false`

dellyoung:
PR: [https://github.com/lodash/lodash/pull/5132]('https://github.com/lodash/lodash/pull/5132')

dellyoung:
[PR]('https://github.com/lodash/lodash/pull/5132')

Tracer1337:
Why do you expect ``false`` here? I would consider an empty ``Date`` object as an object without a timestamp, but that does not exist (as far as I know).

graup:
According to the documentation, "objects are considered empty if they have no own enumerable string keyed properties." Thus IMO `true` is the expected value for `isEmpty(new Date)` as a date is an object with no own properties.

Tracer1337:
Considering the ``Object.getOwnPropertyNames`` methods, all ``Date`` objects are empty 
```bash
> Object.getOwnPropertyNames(new Date)
[]
> Object.getOwnPropertyNames(new Date())
[]
> Object.getOwnPropertyNames(new Date("01.01.2021"))
[]
```
That is also counterintuitive.

jdalton:
> According to the documentation, "objects are considered empty if they have no own enumerable string keyed properties." Thus IMO `true` is the expected value for `isEmpty(new Date)` as a date is an object with no own properties.

Yes. This method name has always carried with it some confusion. Instead of tweaking the method I think we should rename it. I'm open to PRs for that :)

dellyoung:
> > According to the documentation, "objects are considered empty if they have no own enumerable string keyed properties." Thus IMO `true` is the expected value for `isEmpty(new Date)` as a date is an object with no own properties.
> 
> Yes. This method name has always carried with it some confusion. Instead of tweaking the method I think we should rename it. I'm open to PRs for that :)

Thank you, I was misunderstood by the name, I think it's really necessary to change a clear name.

Issue #4878, opened 2020-07-28
sumBy returns boolean instead of number


marian-galik:
`_.sumBy([true], (i) => i === true)`

Actual returns:

`true`

Expected:

`1`

asvsfs:
#2058

maximillianfx:
Hi! The behavior is really wrong, I made a simple solution here, can I take this like a bug-fix?

LarryBattle:
Easiest solution is to always make sure your `iteratee` function always return a number.

Example: `_.sumBy([true], (i) => +(i === true)) // returns 1`

I created a PR for the fix here https://github.com/lodash/lodash/pull/4883

Anyone know how to run all the tests for lodash? I didn't have time to figure it out.
```
npm install && npm test
``` 
fails because it can't create `node lib/main/build-dist.js`.... :-(

dopecodez:
Same here, I was looking at #4865 , and could not figure out how to run all the tests.
Only test script that worked was `test:wip` and that only seems to run a certain number of tests. In my case, the `at.test.js` was running but not the `pullAt.js` test. It'd be helpful for maintenance if you guys could clear up the tests.

bnjmnt4n:
I believe this behaviour is working as intended for now. Although `_.sum` and `_.sumBy` are Math methods, they have been able to perform `_.sum(["a", "b"])`. I believe that summing should also occur with strictly defined and consistent types, and passing booleans or null and undefined to `_.sum` is always going to lead to unexpected issues.

tshma-lex:
I am not sure how is this issue is closed (as well as the closed "duplicates" for null), as I can sum strings (it will concatenate). It might be expected behaviour, but if so - the documentation should change to reflect the return value is not always a number (it can also still return a null or undefined, if all the values are such)

tshma-lex:
Where is the "fresh start" for this issue? I see many like issues (for instance, the sum() return 0 for null, and [null, null] yet return null for [null]), but all are closed, or marked as duplicate to other closed one. It seems no solution was expected. We use the overhead of lodash in our codebase, as it is stable - on opposition to many quirks of pure javascript. So, it's not an excuse that js has unexpected behavior to leave it like this.

PR #5132, opened 2021-04-06
fix return value of isEmpty function #5131

dellyoung:
[#4878](https://github.com/lodash/lodash/issues/5131)


jdalton:
See #4878. Closing in favor of renaming the method.

PR #5141, opened 2021-04-17
Add 'pickArray' function

Tracer1337:
Added a function which picks an object property from every object in an array.

Example:

```js
const arr = [
  { 'a': 1, 'b': 2 },
  { 'a': 1, 'b': 3 },
  { 'a': 2, 'b': 4 }
]
pickArray(arr, 'a')
// => [1, 1, 2]
```

Also works with nested arrays:

```js
const arr = [[1, 2], [1, 3], [2, 4]]
pickArray(arr, 0)
// => [1, 1, 2]
```

### DIFF
--- a/pickArray.js
+++ b/pickArray.js
@@ -1,0 +1,29 @@
+import map from './map.js'
+import flatten from './flatten.js'
+
+/**
+ * Creates an array containing the picked `object` properties.
+ *
+ * @since 5.0.0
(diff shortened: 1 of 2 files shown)


Vitalii4as:
How is it different from `_.map(arr, 'a')`? https://lodash.com/docs/4.17.15#map

PR #5174, opened 2021-05-27
Added Pascal Case Function

sangharsh1008:


dev-itsheng:
Hi.

I also noticed today that a similar function is missing from Lodash, so I searched for Issue & Pull Requests and found it here, and saw your PR.

But your implementation has some problems. According to the link you provided https://wiki.c2.com/?PascalCase, it should capitalize the first letter of each word, and then splicing it directly (no hyphen is required), so The code should be like this:

```javascript
import upperFirst from './upperFirst.js'
import words from './words.js'
import toString from './toString.js'

/**
 * Converts `string` to [camel case](https://en.wikipedia.org/wiki/CamelCase).
 *
 * @since 3.0.0
 * @category String
 * @param {string} [string=''] The string to convert.
 * @returns {string} Returns the camel cased string.
 * @see lowerCase, kebabCase, snakeCase, startCase, upperCase, upperFirst
 * @example
 *
 * camelCase('Foo Bar')
 * // => 'FooBar'
 *
 * camelCase('--foo-bar--')
 * // => 'FooBar'
 *
 * camelCase('__FOO_BAR__')
 * // => 'FooBar'
 */
const camelCase = (string) => (
  words(toString(string).replace(/['\u2019]/g, '')).reduce((result, word, index) => {
    word = word.toLowerCase()
    return result + upperFirst(word);
  }, '')
)

export default camelCase
```

It looks like a long time has passed, would you still be willing to modify the code? If there is no response for a while, I can try to raise a PR myself.

sangharsh1008:
Hii
thanks for informing me, I will modify the code and push the commit in the
same PR.


Thanks and regards
Sangharsh Sovitkar

PR #5181, opened 2021-06-06
Date obj check in isEmpty #5167

### LINKED ISSUE #5167: isEmpty check Date return true
I have a issue when try to check whether  a variable with Date type if empty . 
This is my code: https://codepen.io/vungoccuong/pen/PopzMBv

shreyansh3010:
Fixes #5167

In this PR I've added a date obj check in the isEmpty function which should return false if we pass the date.


shreyansh3010:
@bnjmnt4n can you please check this

PR #5182, opened 2021-06-06
add: titleize and integerDivide functions

electromeow:
Added titleize function to capitalize all words in a string.
Added integerDivide function to return the integer division. Inspired by Python's x // y operator.
Hope these functions are going to be useful for all lodash users.

### DIFF
--- a/titleize.js
+++ b/titleize.js
@@ -1,0 +1,21 @@
+import capitalize from "./capitalize";
+
+/**
+ * Capitalizes each word in the `string`.
+ * @category String
+ * @param {string} [string=''] The string to titleize.
+ * @returns {string} The titleized string.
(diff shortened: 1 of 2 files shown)


titleize.js
```diff
@@ -0,0 +1,21 @@
...
+ * @example
+ *
+ * titleize('HeLLo wOrLD')
+ * // => 'Hello World'
+ */
+
+const titleize = (string = "") => {
+  var splittedString = string.split(" ");
+  for (let i = 0; i < splittedString.length; i++)
+    splittedString[i] = splittedString[i].capitalize();
```
mulfyx, line 17: is it just me, or is there something wrong here

PR #5192, opened 2021-06-17
Add Feat:: _.insert string method

venkateshsundaram:
Added new string method called insert
Hope these functions are going to be useful for all lodash users.


venkateshsundaram:
insert_string_lodash

venkateshsundaram:
@bnjmnt4n could u check this PR

Issue #5186, opened 2021-06-12
Suggestion: comparing objects' common properties


ImRodry:
I'd like to suggest a function similar to isEqual that only compares the common properties between two objects, meaning that it wouldn't throw false for one of them having more or less properties than the other and would simply worry about checking if the properties that they have have the same values. I found a situation in my code where this would be useful and I believe other people could benefit from this too.

porziopablo:
Hi @ImRodry, together with @mariquena we study software engineering, and we're in a course about open source software at our university. We have an assignment in which we have to make a PR for a Linux Foundation project of our choice. We decided to contribute to `lodash` as it is a library we use a lot in our part-time jobs as front-end developers.

We hope this feature satisfies your needs!

ImRodry:
Thank you so much! Looking forward to see this pull request reviewed and accepted and I wish you the best of luck in your assignment!

airarrazaval:
@ImRodry This is a very good idea.  I think it can be further improved by turning them into two functions:

### intersect(object, [sources])

Returns an object whose properties intersect with the sources' properties, maintaining the object's original value if an intersection happens. Returns null if no intersection is found.

**Example**

```js
const obj1 = { a: 1, b: { c: 3, d: 4 } };
const obj2 = { b: { c: 0, e: 5 } };
const obj3 = { f: 1 };

intersect(obj1, obj2);

// output
{
  b: {
    c: 3
  }
}

intersect(obj1, obj3);

// output
null
``` 

### intersectWith(object, [sources], customizer)

The same as `intersect` except that it accepts customizer which is invoked to apply additional intersection rules.

**Example**

```js
const obj1 = { a: 1, b: { c: 3, d: 4 } };
const obj2 = { b: { c: 0, e: 5 } };
const obj3 = { b: { d: 4 }, c: 3 };

function customizer(objValue, srcValue) {
  return objValue === srcValue;
}

intersectWith(obj1, obj2, customizer);

// output
null

intersectWith(obj1, obj3);

// output
{
  b: {
    d: 4
  }
}
```

This will give this function more flexibility and would allow to satisfy more complex scenarios.

ImRodry:
@jdalton why was this closed? Where is the commit that completed this?

anurag-roy:
@ImRodry They closed all open issues and PRs with the `issue bankruptcy` tag.

[...]

PR #5194, opened 2021-06-20
Add Feature: isEqualCommonProps

porziopablo:
This PR adds a new feature which compares objects only by the properties they have in common, as requested in this issue: https://github.com/lodash/lodash/issues/5186.

The solution consists in comparing both objects' list of properties, retain only the ones in common, and build new objects with only these properties. These will then be compared using the normal `isEqual`. 

Together with @mariquena we study software engineering, and we're in course about open source software at our university. We have an assignment in which we have to make a PR for a Linux Foundation project of our choice. We decided to contribute to `lodash` as it is a library we use a lot in our jobs as front-end developers.

### DIFF
--- a/isEqualCommonProps.js
+++ b/isEqualCommonProps.js
@@ -1,0 +1,69 @@
+import baseIsEqual from './.internal/baseIsEqual.js'
+import isEmpty from './isEmpty';
+import intersectionWith from './intersectionWith';
+import isObjectLike from './isObjectLike.js';
+
+/**
+ * This method is like `isEqual` except that it only compares the common properties
+ * between both objects. The method first checks equality with `isEqual`, and if it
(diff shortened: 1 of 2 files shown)


itsdeekay:
Hi @porziopablo, I see your approach. Could you give more insights why you are creating other objects in your code from provided objects. You could have simply take keys from one and check the values from both objects. Any specific reason for your approach?

porziopablo:
Hi @itsdeekay. I create new objects because I think it would be better to compare the common props using the original `isEqual` instead of using `===`. That way the comparison would be more comprehensive, as it's using the original function, which we already know it works fine and includes lots of use cases.

airarrazaval:
@porziopablo see https://github.com/lodash/lodash/issues/5186#issuecomment-890713629

Issue #5225, opened 2021-08-05
lodash.cloneDeep cannot clone the attribute value of the array


spencer17x:
demo: https://codesandbox.io/s/jolly-worker-qtko0?file=/src/index.ts
```ts
import _ from "lodash";

interface Source {
  [k: string]: any;
}

const source: Source = ["React", "Vue"];

source["name"] = "from source";

const result = _.cloneDeep(source);

console.log("result.name", result.name); // undefined
console.log("source.name", source.name); // 'from source'
```

PR #5228, opened 2021-08-07
feat: Add copyArrayWithOwnProperty to clone array property

spencer17x:
to fix issue：https://github.com/lodash/lodash/issues/5225

### DIFF
--- a/.internal/copyArrayWithOwnProperty.js
+++ b/.internal/copyArrayWithOwnProperty.js
@@ -3,19 +3,20 @@
  *
  * @private
  * @param {Array} source The array to copy values from.
- * @param {Array} [array=[]] The array to copy values to.
+ * @param {Array} [target=[]] The array to copy values to.
  * @returns {Array} Returns `array`.
  */
-function copyArrayWithOwnProperty(source, array) {
-  const length = source?.length ?? 0
-  array || (array = new Array(length))
+function copyArrayWithOwnProperty(source, target) {
+  if (!Array.isArray(source)) throw new Error('source is not an array!')
+  const length = source.length
+  target || (target = new Array(length))
 
   for (let key in source) {
-    if (array.hasOwnProperty(key)) {
-      array[key] = source[key]
+    if (target.hasOwnProperty(key)) {
+      target[key] = source[key]
     }
   }
-  return array
+  return target
 }
 
 export default copyArrayWithOwnProperty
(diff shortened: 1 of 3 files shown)


.internal/copyArrayWithOwnProperty.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copies the values of `source` to `array`.
+ *
+ * @private
+ * @param {Array} source The array to copy values from.
+ * @param {Array} [array=[]] The array to copy values to.
+ * @returns {Array} Returns `array`.
+ */
+function copyArrayWithOwnProperty(source, array) {
+  const length = source.length
```
HaseebIjaz, line 10: const length = source?.length ?? 0

HaseebIjaz (changes requested):
use a fallback value for length in case undefined/null is passed in source param

.internal/copyArrayWithOwnProperty.js
Spencer17x: I fixed it, but I think it’s better to judge by Array.isArray

Spencer17x: What do you think?

HaseebIjaz: Yes, it seems to be more appropriate for the said scenario. And in case the source is not array we should throw an exception. Thoughts ?

HaseebIjaz: `!Array.isArray(source) && throw new Error('source is not an array!') ;`

HaseebIjaz: Now , in the suggested workflow , it will throw an exception if the supplied argument is not an array.

.internal/copyArrayWithOwnProperty.js
```diff
@@ -7,7 +7,7 @@
  * @returns {Array} Returns `array`.
  */
 function copyArrayWithOwnProperty(source, array) {
-  const length = source.length
+  const length = source?.length ?? 0
   array || (array = new Array(length))
```
HaseebIjaz: We should check wether both the params are arrays or not first as discussed in the comments as well.

HaseebIjaz (changes requested):

.internal/copyArrayWithOwnProperty.js
Spencer17x: Yes, i fixed it

HaseebIjaz (approved):
Good Work !

spencer17x:
> Good Work !

Can you merge？

spencer17x:
@HaseebIjaz

PR #5229, opened 2021-08-11
Add _.count

canscaw28:
Lodash supports `_.countBy`. It would be great if Lodash also supported a `_.count` option in place of writing `_.countBy` with the `_.identity` iteratee. It's a common pattern for lodash to support two variants of the same function (e.g. `_.min` and `_.minBy`, or `_.sum` and `_.sumBy`). It would be great if we did the same for the `_.countBy` method.

### DIFF
--- a/count.js
+++ b/count.js
@@ -16,7 +16,7 @@ import countBy from './countBy.js'
  * countBy(users, value => value.active);
  * // => { 'blue': 2, 'red': 3, 'green': 1 }
  */
-function count(collection, iteratee) {
+function count(collection) {
   return countBy(collection, value => value)
 }
 
--- a/count.js
+++ b/count.js
@@ -1,0 +1,24 @@
+import countBy from './countBy.js'
+
+/**
+ * Creates an object composed of keys generated from the values of `Collection`.
+ * The corresponding value of each key is the number of times the key was
+ * returned.
+ *
+ * @since 4.17.0
+ * @category Collection
+ * @param {Array} collection The collection to iterate over.
+ * @returns {Object} Returns the composed aggregate object.
+ * @example
+ *
+ * const colors = ['blue', 'red', 'green', 'red', 'red', 'blue']
+ *
+ * countBy(users, value => value.active);
+ * // => { 'blue': 2, 'red': 3, 'green': 1 }
+ */
+function count(collection, iteratee) {
+  return countBy(collection, value => value)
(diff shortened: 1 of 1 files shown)


canscaw28:
Would love to get a review! cc: @jdalton @jashkenas
Wanted to start with something small like this, and ideally start contributing more regularly :)

count.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @since 4.17.0
+ * @category Collection
+ * @param {Array} collection The collection to iterate over.
+ * @returns {Object} Returns the composed aggregate object.
+ * @example
+ *
+ * const colors = ['blue', 'red', 'green', 'red', 'red', 'blue']
+ *
+ * countBy(users, value => value.active);
```
ballonura, line 16: change `users` to `colors`?

canscaw28: good call!

ZimGil:
I don't get it,
The default iteratee of `_.countBy` is `_.identity` which is exactly what this function implements... there's no reason for this function you could just call `_.countBy(foo)` and it's the same

canscaw28:
> I don't get it, The default iteratee of `_.countBy` is `_.identity` which is exactly what this function implements... there's no reason for this function you could just call `_.countBy(foo)` and it's the same

Hey @ZimGil, you are exactly correct that `_.countBy(X)` yields the same result as `_.count(X)`. Since lodash supports functions like `_.min` and `_.minBy`, my thinking here is that this is more about consistency and elegance. That is, _.count(X) is more readable than _.countBy(X) and this is consistent with existing functions (e.g. min/minBy, max/maxBy, mean/meanBy etc. etc.).

count.js
```diff
@@ -0,0 +1,24 @@
...
+ * @returns {Object} Returns the composed aggregate object.
+ * @example
+ *
+ * const colors = ['blue', 'red', 'green', 'red', 'red', 'blue']
+ *
+ * countBy(colors, value => value.active);
+ * // => { 'blue': 2, 'red': 3, 'green': 1 }
+ */
+function count(collection) {
+  return countBy(collection, value => value)
```
JJ, line 20: Since `countBy` with the default iteratee just uses identity

```js
  return countBy(collection)
```

would have the same effect. But, as @ZimGil says, adding a new function for that purpose would probably not be worth the while.

PR #5236, opened 2021-08-16
added findLongest function

qiudaoermu:
This method is to find the longest element in a Array

### DIFF
--- a/findLongest.js
+++ b/findLongest.js
@@ -11,11 +11,17 @@
  * // => "abcde"
  */
 function findLongest(array) {
-  if (!array.length) return "";
-  const len = array.map(item => item.length);
-  const max = Math.max(...len);
-  const index = array.findIndex(item => item.length === max);
-  return array[index];
+  if (!Array.isArray(array)) {
+    if (array === '' || array === undefined || array === null) {
+      return undefined
+    }
+    throw new TypeError('Expected an Array')
+  }
+  if (!array.length) { return undefined }
+  const len = array.map(item => item.length)
+  const max = Math.max(...len)
+  const index = array.findIndex(item => item.length === max)
+  return array[index]
 }
 
-export default findLongest;
+export default findLongest
--- a/findLongest.js
+++ b/findLongest.js
@@ -1,0 +1,21 @@
+/**
+ * This method is to find the longest element in a Array
+ *
+ * @since 5.0.0
+ * @category Collection
+ * @param {Array} collection The collection to inspect.
+ * @returns {string} Returns  longest element in a Array
+ * @example
+ *
+ * findLongest(["abc", 'abcd', "abcde"])
+ * // => "abcde"
+ */
+function findLongest(array) {
+  if (!array.length) return "";
+  const len = array.map(item => item.length);
+  const max = Math.max(...len);
+  const index = array.findIndex(item => item.length === max);
+  return array[index];
+}
+
+export default findLongest;
--- a/findLongest.js
+++ b/findLongest.js
@@ -2,8 +2,8 @@
  * This method is to find the longest element in a Array
  *
  * @since 5.0.0
- * @category Collection
- * @param {Array} collection The collection to inspect.
+ * @category Array
+ * @param {Array} collection The array to inspect.
  * @returns {string} Returns  longest element in a Array
  * @example
  *
--- a/test/findLonest.test.js
+++ b/test/findLonest.test.js
@@ -8,19 +8,21 @@ describe('last', function () {
     assert.strictEqual(last(array), 'abcde')
   })
 
-  it('should return the longest element', function () {
+  it('should return `undefined` if given an empty Array', function () {
     var array = [];
     assert.strictEqual(last(array), undefined)
   })
 
-  it('should return the longest element', function () {
+  it('should return `undefined` when the input is null', function () {
     var array = null;
     assert.strictEqual(last(array), undefined)
   })
 
-  it('should return the longest element', function () {
+  it('should return `undefined` when the input is undefined', function () {
     var array = undefined
     assert.strictEqual(last(array), undefined)
   })
+
+
 
 })
--- a/test/findLonest.test.js
+++ b/test/findLonest.test.js
@@ -1,7 +1,7 @@
 import assert from 'assert'
 import last from '../findLongest'
 
-describe('last', function () {
+describe('findLongest', function () {
 
   it('should return the longest element', function () {
     var array = ['abc', 'abcd', 'abcde']
--- a/test/findLonest.test.js
+++ b/test/findLonest.test.js
@@ -1,14 +1,26 @@
-import assert from 'assert';
-import last from '../findLongest';
+import assert from 'assert'
+import last from '../findLongest'
 
 describe('last', function () {
 
   it('should return the longest element', function () {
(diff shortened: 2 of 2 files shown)


findLongest.js
```diff
@@ -0,0 +1,21 @@
...
+ * @category Array
+ * @param {Array} collection The array to inspect.
+ * @returns {string} Returns  longest element in a Array
+ * @example
+ *
+ * findLongest(["abc", 'abcd', "abcde"])
+ * // => "abcde"
+ */
+function findLongest(array) {
+  if (!array.length) return "";
```
HaseebIjaz, line 14: Put a check a head of it to check whether the 'array' param is an array or not and if its not an array then throw an exception.

test/findLonest.test.js
```diff
@@ -0,0 +1,14 @@
...
+describe('last', function () {
+
+  it('should return the longest element', function () {
+    var array = ['abc', 'abcd', 'abcde'];
+    assert.strictEqual(last(array), 'abcde');
+  });
+  it('should return the longest element', function () {
+    var array = [];
+    assert.strictEqual(last(array), '');
+  });
```
HaseebIjaz, line 13: A test for handling null and undefined should also be added for null saftey check.

HaseebIjaz (changes requested):

test/findLonest.test.js
qiudaoermu: I fixed it

findLongest.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @since 5.0.0
+ * @category Array
+ * @param {Array} collection The array to inspect.
+ * @returns {string} Returns  longest element in a Array
+ * @example
+ *
+ * findLongest(["abc", 'abcd', "abcde"])
+ * // => "abcde"
+ */
```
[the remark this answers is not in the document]
HaseebIjaz: no, i am asking about the VS Code extension you use for these comments.

HaseebIjaz (approved):

findLongest.js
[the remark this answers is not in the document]
qiudaoermu: no extension, 
open preferences => user Snippets ,
image

open javascript.json, copy it
```
{
	"HEADER": {

		"prefix": "header",
		"body": [
			"/**",
			"*",
			"* @since 5.0.0",
			"* @categor",
			"* @param",
			"* @returns",
			"* @example",
			"*",
			"*",
			"*",
			"*",
			"*/"
		],
	}
}
```
then reBuild a js file,
input header
image
this is all;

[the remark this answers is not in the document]
HaseebIjaz: Thanks

findLongest.js
```diff
@@ -0,0 +1,24 @@
...
+function findLongest(array) {
+  if (!Array.isArray(array)) {
+    throw new TypeError('Expected an Array')
+  }
+  if (!array.length) { return undefined }
```
imadx, line 17: Should we return `null` instead?

findLongest.js
```diff
@@ -0,0 +1,24 @@
...
+  const len = array.map(item => item.length)
+  const max = Math.max(...len)
+  const index = array.findIndex(item => item.length === max)
+  return array[index]
```
imadx, line 21: Should we return `null` instead?

findLongest.js
```diff
@@ -0,0 +1,24 @@
...
+ * findLongest(["abc", 'abcd', "abcde"])
+ * // => "abcde"
+ */
+function findLongest(array) {
+  if (!Array.isArray(array)) {
+    throw new TypeError('Expected an Array')
+  }
+  if (!array.length) { return undefined }
+  const len = array.map(item => item.length)
+  const max = Math.max(...len)
```
imadx, line 19: Shall we add tests for arrays with non-string elements?
I think this returns `NaN` if the array contains items that are not strings

findLongest.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * findLongest(["abc", 'abcd', "abcde"])
+ * // => "abcde"
+ */
+function findLongest(array) {
+  if (!Array.isArray(array)) {
+    throw new TypeError('Expected an Array')
+  }
+  if (!array.length) { return undefined }
+  const len = array.map(item => item.length)
```
imadx, line 18: We might have to omit values that are non-strings.  
Or do you think we should throw an error if the item contains items other than just `strings`?

qiudaoermu: yes,it is better

findLongest.js
qiudaoermu: ok,it is more reasonable!

findLongest.js
qiudaoermu: i can't understand here, array is not empty

findLongest.js
qiudaoermu: i think throw an error is better..

findLongest.js
qiudaoermu: i will throw an error if arrays contain non-strings,ok?

findLongest.js
imadx: 🤔 This should've been put here by mistake. Resolving this one.

findLongest.js
imadx: Should be good, yeah.

And if we are restricting to just strings, should we rename the function to something like `findLongestString`?

findLongest.js
qiudaoermu: return array[index] || null
 Is that right？

findLongest.js
qiudaoermu: year, the name is better

findLongestString.js
```diff
@@ -0,0 +1,34 @@
...
+    throw new TypeError('input contains items that are not strings')
+  }
+  if (!array.length) {
+    return null
+  }
+  const arrayEveryItemLen = array.map((item) => item.length)
+  const maxLen = Math.max(...arrayEveryItemLen)
+  const longestStrArr = array.filter((item) => item.length === maxLen)
+  // if longest string has more than one return null
+  if (longestStrArr.length > 1) {
```
Yatin-kathuria, line 28: Instead of returning null , can we return the first largest string in the arr. I think this will be give somthing as a output.

Yatin-kathuria: and we can do this in single iteration by reduce as well.If you want i can share the snippet as well

qiudaoermu: well, i want to see your snippets😆,it's good.

qiudaoermu: well, i want to see your snippets😆,it's good.

Yatin-kathuria:
```javascript
  const longestStrArr = array.reduce((longest,cur) => cur.length > longest.length ? cur : longest,"");
```

qiudaoermu: it's great, but i think it will not pass the test, the type of input must be Array, the type of element must be String

Yatin-kathuria: actually this is just the replacement of lines 24 to 26 . and till line 24 we are sure that we have the array in input because in the starting of function we are checking for array.

qiudaoermu: thanks，i will use it

qiudaoermu: thanks，i will use it

qiudaoermu: hello,i have committed !!

PR #5237, opened 2021-08-18
Turkish language add support capitalize

mehmeTuna:
'Ihlamur'.toLowerCase() -> 'ihlamur' //this is not the expected output
'Ihlamur'.toLocaleLowerCase() -> 'ıhlamur' //this is the expected output

### DIFF
--- a/capitalize.js
+++ b/capitalize.js
@@ -14,7 +14,7 @@ import toString from './toString.js'
  * capitalize('FRED')
  * // => 'Fred'
  */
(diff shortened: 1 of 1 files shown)


ibiBgOR (changes requested):
I have some feedback, that you should improve your request for:
* Missing testcases
* Missing customizability (toLocaleLowerCase accepts an optional locale parameter which should be respected here to)
* Won't work if the first letter is turkish, as the internal function to handle the first letter still uses the wrong function

PR #5241, opened 2021-08-24
add separate method

behnammodi:
```js
const result = separate(['a', 'b', 'c', 'd', 'e'], '!')
 // => ['a', '!', 'b', '!', 'c', '!', 'd', '!', 'e']
```

Also, this is a very useful method for creating a list in UI with `seprator` ex:

```jsx
<ul>
  {seprator([1, 2, 3], <Spacer />).map(element => <li>{ element }</li>)}
</ul>
```

result is:
```jsx
<ul>
  <li>1</li>
  <li><Spacer /></li>
  <li>2</li>
  <li><Spacer /></li>
  <li>3</li>
</ul>
```

Also, I'm not sure about the name `seprator`, might `joiner` or `filler` is better, I don't know, please suggest

### DIFF
--- a/separate.js
+++ b/separate.js
@@ -1,0 +1,31 @@
+import { dropRight } from 'lodash'
+import each from './each'
+
+/**
+ * This method like join, but that returns an array
+ *
+ * @since 1.0.0
+ * @category Array
+ * @param {Array} array The array of grouped elements to process.
+ * @param {any} separator The element separator.
+ * @returns {Array} Returns the an array.
+ * @example
+ *
+ * const result = separate(['a', 'b', 'c'], '!')
+ * // => ['a', '!', 'b', '!', 'c']
+ */
+function separate(array, separator) {
+  if (!(array != null && array.length)) {
+    return []
+  }
+
+  const result = []
+  each(array, (element) => {
+    result.push(element)
+    result.push(separator)
+  })
+
+  return dropRight(result)
+}
+
+export default separate
--- a/test/separate.test.js
+++ b/test/separate.test.js
@@ -1,0 +1,21 @@
+import assert from 'assert';
+import separate from '../separate';
+
+describe('separate', function() {
+  var array = ['a', 'b', 'c'];
+
+  it('should return empty array if array not passed', function() {
+    var actual = separate();
+    assert.deepStrictEqual(actual, []);
+  });
+
+  it('should return array without change if seperator not passed', function() {
+    var actual = separate(array);
+    assert.deepStrictEqual(actual, ['a', undefined, 'b', undefined, 'c']);
+  });
+
+  it('should return array with seperator', function() {
+    var actual = separate(array, '!');
+    assert.deepStrictEqual(actual, ['a', '!', 'b', '!', 'c']);
+  });
+});


separate.js
```diff
@@ -0,0 +1,35 @@
...
+    return array
+  }
+
+  const result = []
+  each(array, (element, index) => {
+    result.push(element)
+    result.push(separator)
+  })
+
+  return dropRight(result)
```
imadx, line 32: Do you think adding (iterating) until the `lastIndex - 1` index and pushing it at the end is better?   
We could save one an extra `dropRight` execution.

separate.js
```diff
@@ -0,0 +1,35 @@
...
+ * @param {Array} array The array of grouped elements to process.
+ * @param {any} separator The element separator.
+ * @returns {Array} Returns the an array.
+ * @example
+ *
+ * const result = separate(['a', 'b', 'c'], '!')
+ * // => ['a', '!', 'b', '!', 'c']
+ */
+function separate(array, separator) {
+  if (!(array != null && array.length)) {
```
imadx, line 18: Should we use `isArray` here and check for length?

separate.js
```diff
@@ -0,0 +1,35 @@
...
+    return []
+  }
+
+  if (separator == null) {
```
imadx, line 22: if `separator` is `undefined`, this won't execute?

separate.js
```diff
@@ -0,0 +1,35 @@
...
+ * @category Array
+ * @param {Array} array The array of grouped elements to process.
+ * @param {any} separator The element separator.
+ * @returns {Array} Returns the an array.
+ * @example
+ *
+ * const result = separate(['a', 'b', 'c'], '!')
+ * // => ['a', '!', 'b', '!', 'c']
+ */
+function separate(array, separator) {
```
imadx, line 17: Even though `separate` is relevant to your use case, I also felt like this is a `join` that pre-empts elements for arrays.  
Not sure what a better name would be though if we are to avoid confusion `[].join`

separate.js
behnammodi: Could you write me a code sample

separate.js
behnammodi: this base on another API

separate.js
behnammodi: undefined is == null

separate.js
behnammodi: yes, I need better name for that

separate.js
imadx: Was thinking of something like the following. What do you think is the better approach?

```es6
let result = []

for (let i = 0; i < array.length - 1; i++) {
  result.push(array[i])
  result.push(separator)
}

result.push(array[array.length - 1])

return result
```

imadx: But now that I think about it, the following would be slightly performant since we are not dynamically updating the array size.

```es6
const finalLength = array.length * 2 - 1;
const result = new Array(finalLength);

for (let i = 0; i < array.length - 1; i++) {
  const resultIndex = i * 2;
  result[resultIndex] = array[i];
  result[resultIndex + 1] = separator;
}

result[result.length - 1] = array[array.length - 1];
```

aspirisen:
This would be very helpful, is there any progress on that, or maybe there is already similar function in lodash?

behnammodi:
> This would be very helpful, is there any progress on that, or maybe there is already similar function in lodash?

I couldn't find a similar function

behnammodi:
@jdalton I would be great to know why this issue closed

jdalton:
@behnammodi Issue bankruptcy. Working on v5 with fresh eyes after a few years.

PR #5243, opened 2021-08-26
Fill array

### LINKED ISSUE #5230: function for multiply arrays
Hi, there is no function that can multiply arrays. Eg. if I have arr = [1, 2, 3] and want to multiply this array by any count for example 3. The result should be [1,2,3,1,2,3,1,2,3]. It would be nice if there is a function for this like multiply(arr, 3) -> return [1,2,3,1,2,3,1,2,3]. Thank you.

It can be done with this:
const xx = [1, 2, 3]
const count = 3
const yy = [...xx, ..._.flatten(_.times(count, () => xx ))]

AlexHalbesleben:
Resolves #5230

I'm not sure if this is something that should be implemented or not. Also, I'm wasn't entirely sure what to name the function, so I went with `fillArray`. If someone can think of a better name, let's change it.


imadx:
Since we have `_.repeat` for strings, do you think `arrayRepeat` would make sense as a function name?

fillArray.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Creates a new array containing the elements of an existing array repeated
+ * @since 4.0.0
+ * @category Array
+ * @param {Array} array The array to repeat
+ * @param {Number} times An integer specifying how many times elements of the existing array should be repeated
+ * @returns {Array} an array containing the elements of the existing array repeated
+ * @example fillArray([1, 2, 3], 3) // returns [1, 2, 3, 1, 2, 3, 1, 2, 3]
```
imadx, line 8: We seem to have the examples after a new line, in our other examples. Should we do the same to be consistent with the rest?

fillArray.js
```diff
@@ -0,0 +1,18 @@
...
+ */
+function fillArray(array, times) {
```
imadx, line 10: We might need to validate the type for the inputs. Should we add tests for invalid inputs?

fillArray.js
```diff
@@ -0,0 +1,18 @@
...
+  let filledArray = []
+  for (let index = 0; index < times; index++) {
+    filledArray = [...filledArray, ...array]
+  }
+  return filledArray
+}
+
+export default fillArray
```
imadx, line 18: Missing a newline at the end?

fillArray.js
imadx: Also `if` we are renaming this to `arrayRepeat`, we might need to update the local variable names also

AlexHalbesleben:
I've renamed the function and corresponding files to `arrayRepeat`. I also added type-checking (and the corresponding tests). 

I noticed that `repeat` might actually be able to work for arrays as well as strings with some minor changes. Do you think it's a better idea to merge `arrayRepeat` and `repeat`?

imadx:
> I noticed that `repeat` might actually be able to work for arrays as well as strings with some minor changes. Do you think it's a better idea to merge `arrayRepeat` and `repeat`?

Do you think it'll be a performance hit on the existing usages?

arrayRepeat.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * Creates a new array containing the elements of an existing array repeated
+ * @since 4.0.0
```
imadx, line 3: This would also need to be updated I think, not sure what the correct version should be though.

DoctypeRosenthal:
What about `repeatArray()`?

PR #5244, opened 2021-08-27
Added flattenObject function

eitanwass:
The function will flatten an object with keys separated by delimiter.

### DIFF
--- a/flattenObject.js
+++ b/flattenObject.js
@@ -1,6 +1,9 @@
 import isPlainObject from './isPlainObject'
 import forEach from './forEach'
 import filter from './filter'
+
+const DEFAULT_DELIMITER = '.'
+
 /**
  * This method will flatten the given object onto a single layer,
  * with keys composed of the keys of each layer separated by a separator.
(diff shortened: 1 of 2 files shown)


flattenObject.js
```diff
@@ -0,0 +1,40 @@
+import isPlainObject from './isPlainObject'
+import forEach from './forEach'
+import filter from './filter'
+/**
```
imadx, line 4: Do we need a new line before this line?

flattenObject.js
```diff
@@ -0,0 +1,40 @@
...
+ * @category Object
+ * @param {Object} object The object to flatten.
+ * @param {string} [delimiter='.'] The string to separate the objects' layers.
+ * @returns {Object} Returns  flattened object.
+ * @example
+ *
+ * flattenObject({'a': 1, 'b': {'c': 3, 'd': 4}}, '.')
+ * // => {'a': 1, 'b.c': 3, 'b.d': 4}
+ */
+function flattenObject(object, delimiter) {
```
imadx, line 18: If we don't provide a delimiter here, it defaults to `,` from the `join`. Do we need to set the default value to `.`?  
Otherwise, we might have to update the jsDocs

PR #5249, opened 2021-09-12
Update memoize to override cache constructor per invocation if required

imadx:
Adding this as an improvement to have the ability to update the Cache Constructor for a single invocation, without setting it globally. Any thoughts?

### DIFF
--- a/memoize.js
+++ b/memoize.js
@@ -15,6 +15,7 @@
  * @category Function
  * @param {Function} func The function to have its output memoized.
  * @param {Function} [resolver] The function to resolve the cache key.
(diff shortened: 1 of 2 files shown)


ethanresnick:
I don't see any reason why the cache should be limited to a `Map` or `WeakMap` — the point of the customizability is to allow non-standard cache implementations (e.g., LRU with a max number of items, a cache with a TTL, etc.), as long as the cache instance _follows the Map interface_. Other than that, though, I love this! It feels really, really bug-prone to have to modify the lodash global to get different caching behavior today.

PR #5308, opened 2021-11-14
add isLowerCase(string)

Yatin-kathuria:
### Hi lodash developers 👋 🌹
I have created a function which is used to check wheather a string is in lowercase or not.

### Examples
```
isLowerCase("foobar") // return true

isLowerCase("FOOBAR") // return false

isLowerCase("FooBar") // return false
```

Hope this contribution is helpfull.

### DIFF
--- a/isLowerCase.js
+++ b/isLowerCase.js
@@ -1,0 +1,22 @@
+/**
+ * Checks if `string` is in `lowercase`.
+ *
+ * @category String
+ * @param {*} string The string to check.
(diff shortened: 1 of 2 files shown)


isLowerCase.js
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * isLowerCase("FOOBAR")
+ * // => false
+ *
+ * isLowerCase("FooBar")
+ * // => false
+ */
+function isLowerCase(string) {
+  return string === string.toLowerCase()
+}
```
cikay, line 20: What is benefit of this?

PR #5313, opened 2021-11-18
isEmpty improvements for number types

vkiranmaniya:
For numbers, `isEmpty` should follow the language rules. `isEmpty` should return false for truthy numbers like 1,5, etc and true for falsy numbers like 0.
```
 isEmpty(0) // true
 isEmpty(4) // false
```


vaynevayne:
I agree that basic types should use the features of the language instead of foolishly returning true. Lodash is dead. Is there any new library recommendation?

PR #5319, opened 2021-11-24
update: using reduce instead of hanging object. 

gondolin24:
Using reduce function over hanging object. Referencing the keys and iterating over them.
If you think about it, Object.keys(object) and iterating over them takes more operations rather than getting it for free in reduce. 
It will reduce calling the original reference by just key. 

https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/


nataliemok-barley:
If you can put a benchmark I think this would be a great contribution :D

Issue #5300, opened 2021-11-08
Feature request: `filterMap()`


ndrewtl:
Hi! Thanks so much for maintaining this project. It would be great to have a function to simultaneously map and filter a collection. That is, to map a collection of values to another collection of values, but to exclude a result if the mapping function returns falsy or `null`.

For instance, we could take a set of years, and return the current age of people born before the current year, dropping the years in the future:

```js
const years = [1942, 1978, 2000, 2021, 2022, 2030, 2045];
const currentYear = 2021;
const ages  = filterMap(years, year => year <= currentYear ? currentYear - year : null)
// => [79, 43, 21, 0]
```

A similar feature already exists in [Rust](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map), and I've found it to be quite useful.

Scttpr:
I think you can use the [reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) method for it in Javascript. For instance :

```js
const years = [1942, 1978, 2000, 2021, 2022, 2030, 2045];
const currentYear = 2021;
const ages = years.reduce((acc, year) => {
  if (year <= currentYear) {
    acc.push(currentYear - year);
  }
  
  return acc;
}, []);
```

ndrewtl:
I agree-- this is how I am currently doing it in my code. However, I would like a separate method. Most methods, even `map()`, could be implemented with `reduce()`, but having a specific method would express the semantics more clearly and simply and clarify my code

g-traub:
You could use [flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) :
```js
 const years = [1942, 1978, 2000, 2021, 2022, 2030, 2045];
const currentYear = 2021;
const ages  = years.flatMap(year => year <= currentYear ? [currentYear - year] : [])
// => [79, 43, 21, 0]
```

ndrewtl:
This still wouldn't work in the case that the value I want to return is an array. For instance:

```js
const strings = [
  'Hello world',
  'Lorem ipsum',
  '',
  'Friends, Romans, countrymen',
  'javascript'
]
// Intended value: each string is mapped to a list of words; the empty string is removed
const wordLists = filterMap(strings, str => str.length > 0 ? str.split(' ') : null)
[...]

PR #5327, opened 2021-12-02
Implement `filterMap()`

ndrewtl:
This implements the feature requested in #5300. Please let me know if you have any questions or comments

### DIFF
--- a/filterMap.js
+++ b/filterMap.js
@@ -26,7 +26,7 @@ function filterMap(array, f) {
     const mapped = f(element)
     if (mapped !== null) {
       // If our mapped value is nonnull, append it to result
-      result.append(mapped)
+      result.push(mapped)
     } // If not, our mapped value is considered 'filtered out'
   })
   return result
--- a/filterMap.js
+++ b/filterMap.js
@@ -21,7 +21,7 @@
  */
 function filterMap(array, f) {
   const result = []
-  array.forEach(element => {
+  array.forEach((element) => {
     // Apply the mapper function to the element, returning a mapped value
     const mapped = f(element)
     if (mapped !== null) {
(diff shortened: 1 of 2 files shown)


camsteffen:
I think it should filter using [`isNil`](https://lodash.com/docs/4.17.15#isNil).

ndrewtl:
@camsteffen the intention here is to have `null` be the one intended 'filter' value and only check for that, because other falsy values, such as `''`, `0`, etc. can be returned intentionally. In my opinion, a `null` signifies an intended non-value, which is a signal to the function to filter, while an `undefined` signifies that a value is missing, which is a valid return type. Furthermore for the Typescript typing, I would prefer to type the filtermap function as `(val: T) => U | null`, in which there is one clear return type that signals it should be filtered out, rather than the ugly `(val: T) => U | null | undefined`, which does not specify to the user if their filtering function should return `null` or `undefined`. Please let me know what you think and I'd like to get this merged.

camsteffen:
`isNil` does not return true for all falsey values, just `null` and `undefined`. I agree with your description of `null` and `undefined`, but I still think `filterMap` should filter both out just because it would be the [least surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) behavior IMO.

ndrewtl:
That's fair. Changed the PR to use `isNil`. Can you review @camsteffen ?

camsteffen:
I'm not a maintainer.

ndrewtl:
Hmm okay. Do you know who I can page then?

ndrewtl:
Hi @jdalton could you review and merge this? This is a function I've wanted to use from lodash for quite awhile, and it would be fantastic if I could get it in.

PR #5333, opened 2021-12-09
Fix: Fix split function to return [] whenever string is undefined

mohamadaminkarami:


split.js
```diff
@@ -25,7 +25,7 @@ const MAX_ARRAY_LENGTH = 4294967295
  */
 function split(string, separator, limit) {
   limit = limit === undefined ? MAX_ARRAY_LENGTH : limit >>> 0
-  if (!limit) {
+  if (!limit || string === undefined) {
```
danimajo, line 28: I would add:
if (!limit || string === undefined || string === null) {

danimajo (changes requested):

split.js
mohamadaminkarami: Hi, I've added that, thanks for your suggestion.

danimajo (approved):
LGTM

split.js
```diff
@@ -25,7 +25,7 @@ const MAX_ARRAY_LENGTH = 4294967295
  */
 function split(string, separator, limit) {
   limit = limit === undefined ? MAX_ARRAY_LENGTH : limit >>> 0
-  if (!limit) {
+  if (!limit || string === undefined || string === null) {
```
cikay, line 28: Use [Abstract Equality Comparison](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness) to catch null and undefined by checking only one like the following

```js
string == null
```

mohamadaminkarami: thanks for your comment. I've fixed it.

PR #5343, opened 2021-12-25
Update has.js

leirons:


has.js
```diff
@@ -21,8 +19,11 @@ const hasOwnProperty = Object.prototype.hasOwnProperty
  * has(other, 'a')
  * // => false
  */
+
+/** hasOwnProperty Used to check objects for own properties. */
+
 function has(object, key) {
-  return object != null && hasOwnProperty.call(object, key)
+  return object != null && object.hasOwnProperty.call(object, key)
```
TheMatrixan, line 26: Your implementation is incorrect for **objects without prototypes**. For example:
```js
const objWOPrototype = Object.create(null);
objWOPrototype.__proto__; // undefined
/// Old implementation: object != null && Object.prototype.hasOwnProperty.call(object, key)
has(objWOPrototype , 'test') // returns false

/// Your implementation: object != null && object.hasOwnProperty.call(object, key)
has(objWOPrototype, 'test') // <boom> Throws an error
// Uncaught TypeError: Cannot read properties of undefined (reading 'call')
//    at has (<anonymous>:2:50)
//    at <anonymous>:1:1
```

It is because `hasOwnProperty` is not in this object (its prototype is empty).

PR #5373, opened 2022-01-27
Added trimIndent

ArjixWasTaken:
Based on kotlin's trimIndent.

I have ran this code independently outside of lodash, but not within it.


ArjixWasTaken:
I guess this will never be merged, so I'm closing this.

PR #5380, opened 2022-02-02
Fix Prototype pollution

Sampaguitas:


jdalton:
Thanks @Sampaguitas but we are no longer patching individual packages.

PR #5392, opened 2022-02-18
Update compact.js for equal sign

simon-ouyang-demo:
use '===' to replace '=='

### DIFF
--- a/compact.js
+++ b/compact.js
@@ -15,7 +15,7 @@ function compact(array) {
   let resIndex = 0
   const result = []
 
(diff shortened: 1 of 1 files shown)


compact.js
```diff
@@ -15,7 +15,7 @@ function compact(array) {
   let resIndex = 0
   const result = []
 
-  if (array == null) {
+  if (array === null) {
```
lazytype, line 18: `==` is desirable and most likely intentional because it will also match `undefined`

PR #5465, opened 2022-06-21
⚡️reduction of lodash size at import

### LINKED ISSUE #5376: Please reduce NPM package size with the "files" option
Hello,
I've been trying to reduce the size of my project's dependencies, and I noticed that lodash takes up a huge amount of space in node_modules: 4.8 megabytes, to be exact. This is entirely unneeded, however, as you already build your entire script into an all-encompassing minified form of lodash.min.js. If you used the `files` parameter in your package.json, the _entirety_ of your package could take up 74 kb (72 kb for lodash.min.js and 2 kb for package.json. I would greatly appreciate it if you did this. You've already done 99% of the work, so I think the only thing you have to do is add `"files": ["lodash.min.js"]` and change `"main": "lodash.js"` to `"main": "lodash.min.js"`. As Lodash is used by literally over a hundred thousand packages, this would have an enormously beneficial impact on the entire community of Node.JS users. (Note that this benefit would be multiplied in the case where different dependencies use different release versions of lodash.)

If you're worried that packages somehow might need all the extra scripts, perhaps you can release a `lodash-complete` package which includes all the source files... but I don't think that would be necessary. After all, if someone wants all the lodash source files, they can just get them from this GitHub repo.

Thank you for your consideration.

image

kleimak:
Fix #5376 

Replacing main entry inside `package.json` by the minified file instead of  `lodash.js` and including the minified file through "files" package.json entry

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,8 @@
   "version": "5.0.0",
   "license": "MIT",
   "private": true,
-  "main": "lodash.js",
+  "files": ["lodash.min.js"],
+  "main": "lodash.min.js",
   "engines": {
     "node": ">=4.0.0"
   },


Kalle888dt (approved):

jdalton:
Let's leave minifying to the build step

JL102:
@jdalton Why? This has nothing at all to do with minifying. This has to do with preventing **completely unnecessary** downloads and saving people's space on disk.
Let's do some math:

Lodash is downloaded 48,000,000 times per week. There are about 3.4 megabytes of unnecessary source files in the Lodash source package. That leaves us with 163,200,000 megabytes of unnecessary downloads **per week.** That's 163.2 TERABYTES of bandwidth every week, or 8.4 PETABYTES of bandwidth every year. Why would you willingly let that pass when you can **so easily** prevent it by simply editing package.json?

PR #5480, opened 2022-07-16
Remapping Numeric Values

mehdikhody:
# Introduction
The remap function takes in five parameters: `value`, `fromLow`, `fromHigh`, `toLow`, and `toHigh`. 
The purpose of this function is to remap a number from one range to another. 

The remap function is used in cases where we need to change a numerical value from one range to another. For example, in graphic and audio programs, this function is used to scale and adjust the volume of audio files and images. Additionally, in programs that work with various sensors such as temperature, humidity, and pressure sensors, this function is used to convert the values read from the sensors into understandable and usable values for the program.

### Syntax
```js
remap(value, fromLow, fromHigh, toLow, toHigh)
```

Here is how each parameter is defined:
- `value`: This is the number that we want to remap.
- `fromLow`: This is the lower bound of the value's current range.
- `fromHigh`: This is the upper bound of the value's current range.
- `toLow`: This is the lower bound of the value's target range.
- `toHigh`: This is the upper bound of the value's target range.

The function will return the remapped value.

### Example 

Here is an example of how to use the remap function:

```js
const remappedValue = _.remap(50, 0, 100, 0, 10);
console.log(remappedValue); // Output: 5

// OR
const useRemap = _.createRemap(0, 100, 0, 10);
console.log(useRemap(50)); // Output: 5
```

In this example, we are remapping the value 50 from its current range of 0-100 to a target range of 0-10. The expected output is 5 because 50 is exactly halfway between the current range's bounds, and so it should be mapped to exactly halfway between the target range's bounds.

### DIFF
--- a/createRemap.js
+++ b/createRemap.js
@@ -1,0 +1,31 @@
+import remap from './remap'
+
+/**
+ * Re-maps a number from one range to another.
+ *
+ * @since 5.0.0
+ * @category Math
+ * @param {number} fromLow the lower bound of the value’s current range.
+ * @param {number} fromHigh the upper bound of the value’s current range.
+ * @param {number} toLow the lower bound of the value’s target range.
+ * @param {number} toHigh the upper bound of the value’s target range.
+ * @returns {(value: number) => number} the remap function.
+ * @see remap
+ * @example
+ *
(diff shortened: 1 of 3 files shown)


mehdikhody:
updated.

PR #5496, opened 2022-08-03
Create reverse.js

Capyhw:
this is a function to reverse part of array

### DIFF
--- a/reverse.js
+++ b/reverse.js
@@ -1,0 +1,16 @@
+/**
+ * @param {any[]} arr
+ * @param {number} start?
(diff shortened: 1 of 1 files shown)


Gauravms2143:
This reverse.js file (reversing an array elements) doesn't exist in lodash repo, But in test folder there is reverse.js is imported.
**Thanks Capyhw for adding this script file.**

Gauravms2143 (changes requested):
It would be good if you change function name from **myReverse() to reverse()**.

PR #5501, opened 2022-08-10
Nth order statistic

samavati:
# Description

I felt the need for a function that can calculate the kth smallest item in an array, so in that case, you can also calculate the median of the array.
in order to do that, I used the quick select algorithm. (also known as Hoare's selection algorithm)
with this algorithm, you will get the average complexity of theta(n), with a worst case of O(n^2).

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my feature works
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/nthOrderStatistic.js
+++ b/nthOrderStatistic.js
@@ -1,48 +1,49 @@
 /**
  * It swaps items at position _i_ and _j_ in the **array**.
  * @param {Array} array Input array to swap items in it
- * @param {Number} i 
- * @param {Number} j 
+ * @param {Number} i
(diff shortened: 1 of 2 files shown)


samavati:
@jdalton , @jashkenas , @falsyvalues 
Could you please review this PR?

PR #5506, opened 2022-08-15
Added JSON Object sorting function

PetiDev:
Adds a function that takes an Object and a Function as an argument and sorts the Object based on the provided function.


Gauravms2143:
Please check above code, **we can reduce iteration** and it'll improve performance.

Issue #4274, opened 2019-04-19
_.xor works incorrectly with 3+ arrays


the-spyke:
Testing on the website
```js
_.xor([3, 4, 1, 6], [0, 1, 2, 3], [3, 2, 7, 6]);
```
gives `[4, 0, 7]`, but should be `[3, 4, 0, 7]`.

jdalton:
Hi @the-spyke!

I believe `[4, 0, 7]` is correct.

the-spyke:
@jdalton If you look at the picture of [symmetric difference](https://en.wikipedia.org/wiki/Symmetric_difference) of 3 sets. The intersection in the center is included. This is the missing number `3`, which is present in all 3 sets.

jdalton:
The diagram in your link is first performing an XOR on A and B and then the result of that on C. In our case we're doing an XOR over all sets at once.

poteat:
@jdalton 

I can confirm the current behavior is per-spec wrong, see the [n-ary symmetric difference](https://en.wikipedia.org/wiki/Symmetric_difference#n-ary_symmetric_difference).

The correct procedure is to count how many times an element has appeared, total.  If it was odd, then that element is included.

The Lodash xor function instead includes elements which are present in only one array.  I agree that this is more useful and intuitive, but the docs shouldn't link to symmetric difference as Lodash xor is objectively not the symmetric difference for any case n >= 3.

jdalton:
Thanks for digging in @poteat!

I'm fine with our behavior.

PR #5514, opened 2022-08-29
Update xor.js Documentation

thelaughingwolf:
Document the difference between `xor` _n_-ary implementation and linked _n_-ary specification.

### DIFF
--- a/xor.js
+++ b/xor.js
@@ -4,8 +4,12 @@ import isArrayLikeObject from './isArrayLikeObject.js'
 /**
  * Creates an array of unique values that is the
  * [symmetric difference](https://en.wikipedia.org/wiki/Symmetric_difference)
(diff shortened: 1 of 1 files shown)


thelaughingwolf:
I've read #4274 and understand the argument there. It makes good sense. However, the documentation still explicitly points to a page that describes behavior NOT matched by `xor`. The documentation should mention that the lodash implementation is not in accordance with the _n_-ary symmetric difference described on the Wikipedia page and is more like the `union` of each input array's `difference` outputs.

thelaughingwolf:
@phapdinh Hello, I'm mentioning you because I see you've committed to Lodash recently, in #5427 .

Please let me know if there's a better way to go about finding somebody to take a look at & approve this PR. 😊

PR #5533, opened 2022-10-09
isEmpty changes  to fullfill developer expectation

eragon11:
1 is a value. Sometimes it was a number also. Even though it was boolean when we use if(isEmpty(1)) -> it returning true. It should return false. So I created a PR please merge it.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -11518,6 +11518,9 @@
       if (isPrototype(value)) {
         return !baseKeys(value).length;
       }
+      if (value == 1) {
+        return false;
+      }
       for (var key in value) {
         if (hasOwnProperty.call(value, key)) {
           return false;


rkuykendall:
@eragon11 Why not `isNumber`? Aren't all numbers non-empty?

MikeWinter:
> @eragon11 Why not `isNumber`? Aren't all numbers non-empty?

I'd argue quite the opposite. A number is a primitive value, not an object, so it can never "contain" anything making it always empty.

The `isEmpty` function is plainly aimed at "collection"-like objects. Anything that isn't one should be considered empty by definition.

It's reasonable that instances of `Object` only take on this "collection"-like characteristic if they have their own properties. Properties gained from the prototype chain shouldn't (and do not) qualify.

```javascript
const a = {};
"toString" in a;               // true
Object.hasOwn(a, "toString");  // false
_.isEmpty(a);                  // true

const b = {
  prop: "value",
};
"prop" in b;               // true
Object.hasOwn(b, "prop");  // true
_.isEmpty(b);              // false
```

Issue #5541, opened 2022-10-26
toString(0n) throw error


eastfire:
toString(0n) throw error: Cannot mix BigInt and other types, use explicit conversions

`return (result == '0' && (1 / value) == -INFINITY) ? '-0' : result; `  

1/0n cause it.

chichiger:
i can take this issue!

PR #5544, opened 2022-10-28
fix big int issue

chichiger:
fixing this issue; https://github.com/lodash/lodash/issues/5541

### DIFF
--- a/toString.js
+++ b/toString.js
@@ -37,6 +37,10 @@ function toString(value) {
   if (isSymbol(value)) {
     return value.toString()
   }
+  if (typeof value === 'bigint') {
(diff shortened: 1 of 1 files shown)


rkuykendall (approved):

toString.js
```diff
@@ -37,6 +37,10 @@ function toString(value) {
   if (isSymbol(value)) {
     return value.toString()
   }
+  if (typeof value === 'bigint') {
+    const result = `${Number(value)}n`;
```
0xflotus, line 41: Why not use the `BigInt` constructor?

```js
const result = BigInt(value);
```

Issue #5573, opened 2022-12-21
Missing `concat.js`


tzhangpccw:
in /test, there is `concat.js`
Line 3: import concat from '../concat.js';

but in the upper directory, there is no `concat.js`.

Please know, thanks. 

Error is like the 
Error [ERR_MODULE_NOT_FOUND]: Cannot find module 'lodash/concat.js' imported from "/lodash/test/concat.js". 
              
---------
  
<img src="https://user-images.githubusercontent.com/107375290/208830232-3e512070-74bd-453f-ac57-d6c21ffa79ef.png" width="400" height="350">

chhongzh:
或许这些文件是给他们测试用的?
Maybe these files are for their testing? 

> 在 /test 中，有`concat.js` 第 3 行： import concat from '../concat.js';
> 
> 但是在上层目录中，没有`concat.js`.
> 
> 请知道，谢谢。
> 
> 错误类似于 错误 [ERR_MODULE_NOT_FOUND]：找不到从“/lodash/test/concat.js”导入的模块“lodash/concat.js”。
> 
> <img alt="" width="400" height="350" src="https://user-images.githubusercontent.com/107375290/208830232-3e512070-74bd-453f-ac57-d6c21ffa79ef.png">

PR #5582, opened 2023-01-10
fix: deleted test/concat.js

Cadienvan:
This test can't pass as the library doesn't provide this function. Suggested in #5573


Gauravms2143 (approved):
LGTM👍

Cadienvan:
Guys is the library abandoned?

Seven-Y-Q-Guo:
@Cadienvan _.concat is still available from [here](https://lodash.com/docs/4.17.15#concat), you may need to put concat.js back rather than deleting it.

Seven-Y-Q-Guo:
We can't find it based on master branch, but if you checkout 4.17 branch, concat method will be there.

https://github.com/lodash/lodash/blob/f299b52f39486275a9e6483b60a410e06520c538/lodash.js#L6975

You see both lodash official doc and 4.17 branch have `concat.js`.
I think they forgot `concat.js` by mistake, That's why I suggest you can implement `concat.js` on master branch.

Seven-Y-Q-Guo:
If you check the test script in package.json
```
"test": "mocha -r esm test/*.test.js"
```
They just test those files which filename contains the keyword `test`, that means those files(not only concat.js) which not have the keyword `test` are not ready to be tested.

Cadienvan:
Oh got ya, I'd love to go further with it but it's clear this project is totally abandoned, all branches haven't been touched for at least 2 years..

Seven-Y-Q-Guo:
haha, I see

PR #5608, opened 2023-03-02
feat: change function get return data judgment

ohlyf:
I think if function get() hava defaultValue,when result===null or result===undefined all should return defaultValue


Seven-Y-Q-Guo:
Hi @ohlyf , have you checked the case you mentioned?
```js
_.get(null, '', 123) // output: 123
_.get(undefined, '', 123) // output: 123
```
Both are working fine, please go ahead to close your PR if it already meets your case.

PR #5609, opened 2023-03-06
Adding a check for the type that gets passed to map

### LINKED ISSUE #5604: _.map doesn't iterate through object at all
I'm not sure what is causing this, but map works with most objects but not the one below:

image

I originally thought that _.map was skipping keys with falsey values, but that doesn't seem correct:

image

Any help is much appreciated!

omarzawahry:
Adding a simple check if the type is an object to get the length from the keys, then if it's an array it gets the length normally otherwise it's 0

Closes #5604

### DIFF
--- a/map.js
+++ b/map.js
@@ -18,7 +18,8 @@
  */
 function map(array, iteratee) {
   let index = -1
-  const length = array == null ? 0 : array.length
+  const length = typeof array === 'object' ? array.keys().length :
+                 Array.isArray(array) ? array.length : 0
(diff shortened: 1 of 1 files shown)


rianonehub (approved):

Seven-Y-Q-Guo:
@rianonehub I think you'd better add unit test to cover this case.

SitanHuang:
Shouldn't the Array.isArray come before the typeof check? Since `typeof [] === "object"`

PR #5648, opened 2023-05-16
chore: use === instead of ==

anmol242:
This PR will strictly check type check instances by using === instead of ==


hexiaokang (approved):

anmol242:
@jdalton, @hexiaokang or anyone from team can we merge this PR?

PR #5668, opened 2023-06-20
Adding case functions

NorielSylvire:
- Added camelCaseKeys (changes the keys of an object to camelCase)
- Added pascalCase (converts string to PascalCase)
- Static Badge
- Static Badge
- Static Badge

### DIFF
--- a/camelCaseKeys.js
+++ b/camelCaseKeys.js
@@ -1,0 +1,21 @@
+/**
+ * Converts the keys of an object to camel case recursively.
+ * @param {object} obj - The input object to be processed.
+ * @returns {object} - The object with camel case keys.
+ */
+const  camelCaseKeys = (obj) => {
+  if (typeof obj !== 'object' || obj === null) {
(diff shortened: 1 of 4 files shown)


camelCaseKeys.js
```diff
@@ -0,0 +1,21 @@
...
+  if (typeof obj !== 'object' || obj === null) {
+    return obj
+  }
+
+  if (Array.isArray(obj)) {
+    return obj.map(camelCaseKeys)
+  }
+
+  return Object.keys(obj).reduce((acc, key) => {
+    const camelCaseKey = key.replace(/_([a-z])/g, (_, letter) => letter.toUpperCase())
```
Maquim4, line 16: The regular expression used for replacing underscores with camelCase is effective, but it's case-sensitive. If your data might contain uppercase letters in keys (e.g., "snake_Case"), you'll need to modify the regular expression to be case-insensitive by using the /i flag.

PR #5675, opened 2023-07-07
Fix baseClone bug when cloning sparse array

Taragryen:
Cloning a sparse array with cloneDeep will convert empty to undefined.
image

We can use the `in` operator to check for empty elements
image

If `key in value` is false, we can return it early to avoid converting the empty element to undefined

### DIFF
--- a/.internal/baseClone.js
+++ b/.internal/baseClone.js
@@ -232,6 +232,16 @@ function baseClone(value, bitmask, customizer, key, object, stack) {
       key = subValue
       subValue = value[key]
     }
+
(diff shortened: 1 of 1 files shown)


bendasgfyug:
> Cloning a sparse array with cloneDeep will convert empty to undefined.
> varanlesimage
> 502
> We can use the `in` operator to check for empty elements
>Impose santipn Bendasfgugimage
> True
> If `key in value` is false, we can return it early to avoid converting the empty element to undefined values

PR #5685, opened 2023-07-28
fix(sumBy): Handle string-to-number conversion and prevent concatenation in baseSum function

VadimBorodin:
### Pull request description
<i>This pull request resolves an issue in the baseSum function, where strings were being concatenated instead of converted to numbers, leading to incorrect results. The fix includes proper handling of string-to-number conversion, ensuring accurate summation in the presence of strings.</i>

### Type
- [x] Bug-fix
- [x] Testing

### DIFF
--- a/.internal/baseSum.js
+++ b/.internal/baseSum.js
@@ -1,3 +1,5 @@
+import baseToNumber from '.internal/baseToNumber.js'
+
 /**
  * The base implementation of `sum` and `sumBy`.
  *
@@ -10,8 +12,11 @@ function baseSum(array, iteratee) {
   let result
 
   for (const value of array) {
-    const current = iteratee(value)
+    let current = iteratee(value)
     if (current !== undefined) {
+      if (typeof current === 'string') {
+        current = baseToNumber(current)
+      }
       result = result === undefined ? current : (result + current)
     }
   }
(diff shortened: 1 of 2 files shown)


maloun96 (approved):

.internal/baseSum.js
```diff
@@ -1,3 +1,5 @@
+import baseToNumber from '.internal/baseToNumber.js'
```
cristijora, line 1: Let's rather prefix `_internal` as it might cause issues with `.` in linux or other OS systems

.internal/baseSum.js
```diff
@@ -10,8 +12,11 @@ function baseSum(array, iteratee) {
   let result
 
   for (const value of array) {
-    const current = iteratee(value)
+    let current = iteratee(value)
     if (current !== undefined) {
+      if (typeof current === 'string') {
+        current = baseToNumber(current)
```
cristijora, line 18: Just wondering, why don't we call this `stringToNumber` or `parseNumber` Feels more intuitive than `baseToNumber`

VadimBorodin: I think that the `baseToNumber` is more generic, we can convert other types: binary, hexadecimal...

Issue #4701, opened 2020-03-24
`.get` works different from standard property accessors


jas0ncn:
```javascript
const res1 = { data: { text: 'hi' }, 'data.text': '<img />' }

_.get(res1, 'data.text') // "<img />"

res1.data.text // "hi"
```

jmm:
It would be good to document how `.get` handles this situation where there's a literal key matching the path argument, but if you want to be sure you get the path you had in mind you can pass it as an array:

```js
_.get(res1, 'data.text'.split("."))
```

jas0ncn:
@jmm Agree! It should be documented.

PR #5686, opened 2023-07-28
chore: update get() docs

zcericola:
Add additional documentation for the `_.get()` method to illustrate behavior when the object's key is an exact match of the path argument. This is in response to open issue #4701.

### DIFF
--- a/get.js
+++ b/get.js
@@ -13,7 +13,7 @@ import baseGet from './.internal/baseGet.js'
  * @see has, hasIn, set, unset
  * @example
  *
- * const object = { 'a': [{ 'b': { 'c': 3 } }] }
+ * const object = { 'a': [{ 'b': { 'c': 3 }, 'b.c': 4 }] }
  *
  * get(object, 'a[0].b.c')
  * // => 3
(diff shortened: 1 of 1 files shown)


nealeu:
Please could you merge this. 

It's a real "WTF!" when you trip over this, and with some frameworks it does happen that you have `obj['a.b']` and `obj.a.b` for different purposes (I've no idea why!).

I thought I had found a bug in lodash, then tried to reproduce and lodash worked as documented, but that was because my reproduction was not complete.

nealeu:
Wow. Why should people bother to PR?

jdalton:
Lodash v5 will be getting rid of string based paths since we now have [optional chaining syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) in the JS :tada:

nealeu:
> Lodash v5 will be getting rid of string based paths since we now have [optional chaining syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) in the JS 🎉

That doesn't help as we've had it for ages for people sensible enough to use Typescript.  

I wouldn't need lodash in the first place for this if I weren't working with frameworks that use string based paths for data driven forms and tables.

jdalton:
@nealeu 
> I wouldn't need lodash in the first place for this if I weren't working with frameworks that use string based paths for data driven forms and tables.

Interesting! Could you provide a little snippet of an example of your usage. Is it through variables or dynamically generated paths of some kind?

Issue #5690, opened 2023-08-18
The isEmpty method is not very rigorous in judging empty objects.


fsiaonma:
The isEmpty method is not very rigorous in judging empty objects.

```js
const obj = {
  [Symbol('a')]: 1
}
```
when I call `_.isEmpty(obj)`,  it return `true`;

Have you considered using `Reflect.ownKeys()` to solve this problem?

dev-davidyoon:
I would like to be assigned this issue. I am working on pushing up a PR to address this, thanks.

fsiaonma:
@dev-davidyoon thanks

angelsupp:
how about this [PR](https://github.com/lodash/lodash/pull/5702)? Does it solve the issue?

deal-talenya:
Hi, we need this fix because mongoose ObjectId value is identified as empty by isEmpty function
but it looks like earlier it worked as expected

n7s4:
哲哥太强了

PR #5705, opened 2023-08-31
update isEmpty.js

humingMocking:
Solved the problem where this method cannot solve scenarios where there is a symbol data type within an object
#5690


deal-talenya:
+1

Issue #5730, opened 2023-09-20
bun test failed on main


tisonkun:
I don't know if my environment misses something or version mismatches:

```
$ bun --version
1.0.2
$ uname -a
Darwin xxx.local 22.5.0 Darwin Kernel Version 22.5.0: Mon Apr 24 20:53:19 PDT 2023; root:xnu-8796.121.2~5/RELEASE_ARM64_T6020 arm64
```

Output:

https://gist.github.com/tisonkun/ac2317781001f326457b0d46b7ae889d

tisonkun:
@jdalton Thanks for your information! So before we start making any bugfixes or features, we should smooth the development setup?

I saw the other issues ask for some behavior bugs and I'm trying to dig into there, but a proper develop env is a prerequisite :D

PR #5733, opened 2023-09-21
test: partially fix broken tests

tisonkun:
From 280+ to 190+.

There are many syntax errors in the same pattern of fragmented array perhaps due to a brittle global find and replace.

This refers to #5730.


jdalton:
Thank you @tisonkun !

Gb2real:
Now I got it down learn something new every day

PR #5736, opened 2023-09-24
Modified ternary operator.

joyson-madtha0:


src/toFinite.ts
```diff
@@ -27,7 +27,7 @@ const MAX_INTEGER = 1.7976931348623157e308;
  */
 function toFinite(value) {
     if (!value) {
-        return value === 0 ? value : 0;
+        return 0;
```
jdalton, line 30: Hi @joyson-madtha0! This is actually checking for `-0` and allowing it to pass through.

joyson-madtha0: Correct, I am just curious, but `0 === -0` or `0 == -0` always results in true. So, it doesn't make a difference in computation.

PR #5739, opened 2023-09-26
feat: isStream

victimsss:
https://github.com/nodejs/node/blob/f16f41c5b3cf2345abc7bd4d5e951a43479ce4a7/deps/undici/src/lib/core/util.js#L17


jdalton:
Hi @victimsss! I prefer brand checking if possible. What is the result of:
```js
Object.prototype.toString.call(stream); // ?
```

ex1st:
Hi @victimsss Web Stream doesn't have "pipe" method, please check API https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Concepts#pipe_chains

PR #5741, opened 2023-09-26
test: fix cannot find root

tisonkun:


tisonkun:
@jdalton @jonchurch 

I noticed that all the commits trying to migrate to bun are overridden with a 2021 baseline commit tree. Is there any background on this?

Personally I like this move so that we can focus on library evolution instead of testing all the bun bugs :P

PR #5743, opened 2023-09-29
Added flattenObject function

eitanwass:
Function to flatten an object with keys separated by delimiter.
This is a re-open of a previous PR that closed when al other PRs closed.


jdalton:
Hi @eitanwass! The joining of the flattened keys seems counter intuitive to me. I think this would be a better fit as a one-off utility function.

PR #5754, opened 2023-10-14
Fixes #5751

### LINKED ISSUE #5751: _.size() function returns wrong value when one of the key is 'length'
Lodash

I have noticed that the `_.size` function in lodash is behaving unexpectedly. When I call `_.size({ 'a': 1, 'b': 2, 'length': 5 })`, I expect it to return `3` as it should count the properties 'a', 'b', and 'length'. However, it's returning `5`, which is unexpected.

Expected behavior:
`_.size({ 'a': 1, 'b': 2, 'length': 5 })` should return `3`.

Actual behavior:
`_.size({ 'a': 1, 'b': 2, 'length': 5 })` returns `5`.

Steps to reproduce:
1. Use the lodash library.
2. Call `_.size({ 'a': 1, 'b': 2, 'length': 5 })`.

I believe this issue might be related to how the `length` property is being handled, as it's not a string-keyed property but is still counted. Please investigate and correct this behavior.

Reterics:
According to the issue [#5751](https://github.com/lodash/lodash/issues/5751), the size is not calculating the length properly when we give it a native object with 'length' key, like this: `_.size({ 'a': 1, 'b': 2, 'length': 5 })`

I excluded native JS objects in size function by `"collection.constructor.name !== 'Object'"`. Other verification was not needed since the **isArrayLike** method checked before that we have an array like object.

### DIFF
--- a/src/size.ts
+++ b/src/size.ts
@@ -30,7 +30,7 @@ function size(collection) {
     if (collection == null) {
         return 0;
     }
-    if (isArrayLike(collection)) {
+    if (isArrayLike(collection) && collection.constructor.name !== 'Object') {
         return isString(collection) ? stringSize(collection) : collection.length;
     }
     const tag = getTag(collection);


src/size.ts
```diff
@@ -30,7 +30,7 @@ function size(collection) {
     if (collection == null) {
         return 0;
     }
-    if (isArrayLike(collection)) {
+    if (isArrayLike(collection) && collection.constructor.name !== 'Object') {
```
ex1st, line 33:
```javascript
const collection = Object.create(null);
collection.length = 1;
_.size(collection); // Uncaught TypeError: Cannot read properties of undefined (reading 'name')
```

jdalton:
`_.size` is a collection method. You'd want to use object only methods.

Issue #5767, opened 2023-10-29
test codes are seem broken


Howl-Kim:
In `test/debounce.spec.js`

<img width="776" alt="Screenshot 2023-10-29 at 5 41 19 PM" src="https://github.com/lodash/lodash/assets/89368498/f0bb79a0-4d1e-477e-8621-ee4f7b0cddc3">

targhs:
I can't figure out a way to run the test script. Hopefully this will fix.

moacirosa:
I'm don't know too much about _lodash_ development but I've noticed that it looks to be passing for some hard changes (ex: moving to [_bun_](https://bun.sh/)). For the test you mentioned it looks that a fix like above should resolve:

```diff
diff --git a/test/debounce.spec.js b/test/debounce.spec.js
index 68fdd858d..fdb3a64f7 100644
--- a/test/debounce.spec.js
+++ b/test/debounce.spec.js
@@ -11,14 +11,14 @@ describe('debounce', () => {
         }, 32);
 
         const results = [debounced('a'), debounced('b'), debounced('c')];
-        expect(results, [undefined, undefined).toEqual(undefined]);
+        expect(results).toEqual([undefined, undefined, undefined]);
         expect(callCount).toBe(0);
 
         setTimeout(() => {
             expect(callCount).toBe(1);
 
             const results = [debounced('d'), debounced('e'), debounced('f')];
-            expect(results, ['c', 'c').toEqual('c']);
+            expect(results).toEqual(['c', 'c', 'c']);
             expect(callCount).toBe(1);
         }, 128);
```

Other tests are also failing for a similar reason in this and other files. Maybe there is more info [in this discussion](https://github.com/lodash/lodash/discussions/5732).

jdalton:
Yes the tests are in flux. Any PRs to help get them up and running would be appreciated :D

ljharb:
CI is passing now; I think this is resolved.

PR #5769, opened 2023-10-31
fix: syntax errors for debounce tests

targhs:
#5767


jdalton:
Reverted to a stable branch.

PR #5771, opened 2023-11-03
Update flatMap.ts

HarishRMaryada:
Updated to arrow function


src/flatMapDeep.ts
```diff
@@ -23,8 +23,4 @@ const INFINITY = 1 / 0;
  * flatMapDeep([1, 2], duplicate)
  * // => [1, 1, 2, 2]
  */
-function flatMapDeep(collection, iteratee) {
-    return baseFlatten(map(collection, iteratee), INFINITY);
-}
-
-export default flatMapDeep;
+export const flatMapDeep = (collection, iteratee) => baseFlatten(map(collection, iteratee), INFINITY);
```
JakeerC, line 26: `export const` doesn't do default export

src/before.ts
```diff
@@ -13,12 +13,12 @@
  * jQuery(element).on('click', before(5, addContactToList))
  * // => Allows adding up to 4 contacts to the list.
  */
-function before(n, func) {
+export const before = (n, func) => {
```
JakeerC, line 16: `export const` doesn't do default export

src/flatMapDepth.ts
```diff
@@ -21,9 +21,8 @@ import map from './map.js';
  * flatMapDepth([1, 2], duplicate, 2)
  * // => [[1, 1], [2, 2]]
  */
-function flatMapDepth(collection, iteratee, depth) {
+export const flatMapDepth = (collection, iteratee, depth) => {
```
JakeerC, line 24: `export const` doesn't do default export

JakeerC (changes requested):
`export const` doesn't do default export

PR #5785, opened 2023-12-11
chore: combine the if statements into one

KEHyeon:
In the above line, the function is returned when both the variables are undefined, which makes it evident that when one of the two variables is undefined, one of them must have a value. So I personally think reducing the two if statements as one would be clearer.

### DIFF
--- a/src/.internal/createMathOperation.ts
+++ b/src/.internal/createMathOperation.ts
@@ -14,11 +14,8 @@ function createMathOperation(operator, defaultValue) {
     if (value === undefined && other === undefined) {
       return defaultValue
     }
-    if (value !== undefined && other === undefined) {
-      return value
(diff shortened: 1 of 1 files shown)


jdalton (approved):

KEHyeon:
thank youuuuuuuu

KEHyeon:
I'm a 20-year-old developer studying in Korea! I'm happy that I uploaded a PR to open source for the first time and got it merged.
I also want to become a good developer like you. I'm not going to college and I'm studying on my own. Will I be successful and get a job in Silicon Valley?
Can you give me some advice?
-----Original Message-----

Gb2real:
Got all sections done security clearance and everything is done now ready for green light marathon

PR #5799, opened 2024-01-04
Fix compareAscending for numeric values

deje1011:
I think this bug was introduced by accident when removing the hardcoded locale here: https://github.com/lodash/lodash/commit/47a6d538f5759fc5788f1bbb147caa7fde6b0a92


jdalton:
Reverted to a stable branch.

PR #5819, opened 2024-02-14
fixed #5818

### LINKED ISSUE #5818: Method `sumBy` returns string
There are cases when method `sumBy` could return `string` instead of `number`. 

**Current behavior:**
```ts
_.sumBy([{ a: 1 }, { a: 'f' }, { a: null }], 'a'); // '1fnull'
```
Returned values is of type string.

Expected behavior:
```ts
_.sumBy([{ a: 1 }, { a: 'f' }, { a: null }], 'a'); // NaN
```
Returned values is of type number.

lpabinash:
To handle non-numeric values and return NaN for the given input, updated the sumBy function to check if the result is not a number before returning it.
Modified the baseSum function to handle non-numeric values by checking the type of the current value and only adding it to the result if it's a number.


JakeerC (approved):

jdalton:
Reverted to a stable branch.

PR #5841, opened 2024-04-08
feat(head): add optional n elements parameter

danielbayley:
For example:
~~~ js
const array = [1, 2, 3, 4]

_.head(array, { n: 2 }) // [1, 2]
// or
_.first(array, { n: 2 }) // [1, 2]

_.head(array) // 1

[array].map(_.head) // [1]
~~~


Trott:
Currently, `head()` always returns the first element.

With this change, `head([0,1], 500)` returns an array, but `head([1], 500)` returns `1`. Seems like a footgun.

danielbayley:
> I imagine this breaks usage of `.head()` as a callback with `Array.prototype.map()` and similar functions. `map()` sends the index of the array as the second argument to the callback. With this change, that will now affect how many elements are returned.

@Trott  Fair point. My first contribution to this project, so not entirely sure what the design goal specifics are… But my assumption was that pure _functional_ methods—with strict arity—were for `lodash/fp`?

Anyway, just pushed a fix to avoid said potential foot injuries!

An alternative approach might be to have something like `headN`/`firstN` additional methods…

Trott:
> My first contribution to this project, so not entirely sure what the design goal specifics are… But my assumption was that pure _functional_ methods—with strict arity—were for `lodash/fp`?

Yeah, I'm not a maintainer or even much involved in `lodash` whatsoever. So feel free to disregard for the reasons you state and await a review from a project member.

PR #5842, opened 2024-04-08
feat(last): add optional n elements parameter

danielbayley:
For example:
~~~ js
const array = [1,2,3,4]

_.last(array, { n: 2 }) // [3, 4]

_.last(array) // 4

[array].map(_.last) // [4]
~~~

### DIFF
--- a/src/last.ts
+++ b/src/last.ts
@@ -1,18 +1,22 @@
 /**
- * Gets the last element of `array`.
+ * Gets the last [n] element(s) of `array`.
  *
  * @since 0.1.0
  * @category Array
  * @param {Array} array The array to query.
- * @returns {*} Returns the last element of `array`.
+ * @param {Number} n The number of elements.
+ * @returns {*} Returns the last [n] element(s) of `array`.
  * @example
  *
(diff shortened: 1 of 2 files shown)


Trott:
Adding this feature seems like it will break use cases where `last()` is used as a callback to `Array.prototype.map()` and similar functions. 

It also seems like there's a footgun where `.last([1], 500)` returns `1` but `.last([1,2,3], 500)` returns an array.

jdalton:
Our `last` method will be dropped for `Array.prototype.at` as in `array.at(-1)`:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at

danielbayley:
> Our `last` method will be dropped for `Array.prototype.at` as in `array.at(-1)`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at

@jdalton That’s nice, but it doesn’t give you a range as is the point of this PR…

jdalton:
Are you thinking of https://lodash.com/docs/4.17.15#takeRight ?

PR #5846, opened 2024-04-11
feat: asyncFlow and asyncFlowRight methods

andrii-milishchuk:
New methods which duplicates logic of `flow` and `flowRight` but with async functions support


Andrik264:
I've got a problem with my project where I need `flow` and `flowRight` methods but callbacks I pass to the `flow` are async, so I decided to add these 2 new methods to the lib

Trott:
Needs tests.

Andrik264:
So.. what is next?

PR #5852, opened 2024-04-22
Added default value field to toNumber function

kapilkumar-github:
Modified toNumber method to have two parameters
1) value
2) defaultValue - when given value can't be converted into number it will return the default value


Trott:
Needs tests and docs.

Doesn't seem like a great API function signature. If I look at code and see `_.toNumber(something)`, that's clear. If I see `_.toNumber(something, somethingElse)`, I'm not sure what is going on. I'm not sure about the performance implications, but it is probably better to use an options object in this case: `_.toNumber(something, {defaultValue: 0})`

kapilkumar9395:
@Trott thanks for the input. I have made the necessary changes

Trott:
> @Trott thanks for the input. I have made the necessary changes

I'm not a lodash maintainer or anything, so bear that in mind, but I would imagine that the tests should be added to `test/number-coercion-methods.spec.js` rather than creating a separate test file just for additional `toNumber()` tests. That also raises the question of whether symmetric changes should be made to the API of `toFinite()`, `toInteger()`, and  `toSafeInteger()`. But those never return `NaN` so I'm not sure it makes sense.

That in turn raises the issue of whether this API adjustment makes sense. It seems like it would typically be used to return `0` instead of `NaN`, but that's already what `toFinite()` does, and `toFinite()` has the added bonus of returning usable values for `Infinity` and `-Infinity`.

PR #5855, opened 2024-04-29
feat: use native `structuredClone()` in `_.cloneDeep()`

### LINKED ISSUE #5833: Attempt structuredClone in cloneDeep before doing manual clone
I would assume that the [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) API being native would allow JS engine makers to optimize the performance. Though it isn't as powerful as Lodash's clone deep, that power may not be needed in all use cases.

My suggestion is basically this:

```js
function cloneDeep (value) {
  try {
    if (structuredClone) {
      return structuredClone(value);
    }
  } catch {
    return doWhatLodashHasAlwaysDoneForCloneDeep(value);
  }
}
```

Just an idea, not sure if this would be a net positive or negative (more people dealing with the time it takes to do the failed attempt before falling back, versus more people getting the fast native solution and skipping the more complex JS solution).

Alternatively every user could just make their own helper function like this:

```js
import { cloneDeep } from 'lodash.cloneDeep';

export const deepClone = function (value) {
  try {
    if (structuredClone) {
      return structuredClone(value);
    }
  } catch {
    return cloneDeep(value);
  }
};
```

But the whole point of Lodash it to not need to make/copy a bunch of common helpers for stuff like this.

lewxdev:
resolves #5833

I tried to add tests but they don't seem to be running at the moment, this can be adjusted as necessary

Any feedback on the proposed approach and any adjustments or additions is greatly appreciated. Additionally, insights on enhancing the implementation are welcome as well.


kapilkumar9395:
Instead of having skipNativeCheck as a parameter, it would be better to have an option object something like -
`cloneDeep(value, { skipNativeCheck : false })`

It's more scalable, and descriptive

Trott:
The `yarn.loock` update seems unrelated?

lewxdev:
> Instead of having skipNativeCheck as a parameter, it would be better to have an option object something like - `cloneDeep(value, { skipNativeCheck : false })`
> 
> It's more scalable, and descriptive

That works, I just figured using an object for a single option would be unnecessary (and I didn't see too many other uses of options for the other utilities).

But the scalability argument makes sense and I'm open to making this change

lewxdev:
> The `yarn.loock` update seems unrelated?

Yeah, the `yarn.lock` got updated when I did `bun i`. Maybe I can restore the old version and try again with `--frozen-lockfile`?

test/clone-methods.spec.js
```diff
@@ -117,6 +117,14 @@ xdescribe('clone methods', function () {
         assert.notStrictEqual(actual, cyclical[`v${LARGE_ARRAY_SIZE - 1}`]);
     });
 
+    it('`_.cloneDeep` should accept the `skipNativeCheck` flag', () => {
+        const object = { primitive: 'a', fn: () => 'b' };
+        const actual = cloneDeep(object, true);
```
TheJaredWilcurt, line 122:
```suggestion
        const actual = cloneDeep(object, { skipNativeCheck: true });
```

src/cloneDeep.ts
```diff
@@ -20,9 +28,23 @@ const CLONE_SYMBOLS_FLAG = 4;
  * const deep = cloneDeep(objects)
  * console.log(deep[0] === objects[0])
  * // => false
+ *
+ * // The `skipNativeCheck` flag
+ * const unsupportedNativeObject = { fn: () => 'a' };
+ *
+ * const deep = cloneDeep(unsupportedNativeObject, true);
```
TheJaredWilcurt, line 35:
```suggestion
 * const deep = cloneDeep(unsupportedNativeObject, { skipNativeCheck: true });
```

src/cloneDeep.ts
```diff
@@ -20,9 +28,23 @@ const CLONE_SYMBOLS_FLAG = 4;
...
+ * // => false
  */
-function cloneDeep(value) {
-    return baseClone(value, CLONE_DEEP_FLAG | CLONE_SYMBOLS_FLAG);
+function cloneDeep(value, options = {}) {
+    try {
+        if (!options.skipNativeCheck && structuredClone) {
+            return structuredClone(value);
+        }
+        throw new Error('Unsupported structured clone');
```
TheJaredWilcurt, line 44: Slightly more accurate, though nothing uses this message so it doesn't really matter.
```suggestion
        throw new Error('structuredClone unsupported or skipped');
```

TheJaredWilcurt (approved):
Not sure about the yarn lock, but the rest looks good to me.

TheJaredWilcurt:
@jdalton I think this is good. Thoughts on the yarn lock?

lewxdev:
I went ahead and just removed the yarn-lock changes. It may need updating, but that seems out of scope of this PR and can be done separately. @jdalton I think this is ready to be merged when you have time to look at it.

ex1st:
Just note:

```javascript
// lodash v4
_.cloneDeep({ class: class Class {} })
// {class: ƒ}
_.cloneDeep({ function: function () {} })
// {function: ƒ}


structuredClone({ class: class Class {} })
// DOMException [DataCloneError]: class Class {} could not be cloned
structuredClone({ function: function () {} })
// DOMException [DataCloneError]: function () {} could not be cloned
```

PR #5858, opened 2024-05-02
fix(memoize): error when passing multiple arguments to memoized function without having a resolver available

WolvenSpirit:
Do not allow `memoize` to be used by a function with multiple arguments if it has no resolver available.
This is a change that is supposed to throw error in the face of devs during CI and not allow memoize to make it to production code if used incorrectly.

This added behavior is likely a breaking change.


WolvenSpirit:
Only opening this PR because I got bitten by the current behavior of this utility function.

jdalton:
Closing as stale: this conflicts with the current base and its diff has expanded into the full JS→TS migration in reverse. If the change is still relevant, please open a fresh PR against `main`.

PR #5866, opened 2024-05-24
Clarify what arguments goes where.

SimonLandeholm:


src/defaults.ts
```diff
@@ -16,8 +16,8 @@ const hasOwnProperty = objectProto.hasOwnProperty;
  *
  * @since 0.1.0
  * @category Object
- * @param {Object} object The destination object.
- * @param {...Object} [sources] The source objects.
+ * @param {Object} object The destination object. (the object without the default properties)
+ * @param {...Object} [sources] The source objects. (the object/objects denoting the default properties)
```
Trott, line 20:
```suggestion
 * @param {Object} object The destination object. (The object without the default properties.)
 * @param {...Object} [sources] The source objects. (The object/objects denoting the default properties.)
```

Cassieminkus1 (approved):

jdalton:
Thanks, but the diff here is a large, undiscussed rewrite (a full TypeScript migration) rather than the documentation clarification in the title, so I can't take it as-is. If you'd like to propose a focused doc change, please open a minimal PR. Closing this one.

PR #5869, opened 2024-05-26
feat: flatten an object functionality

david-pilny:
# `flattenObject` Function

The `flattenObject` function is a utility that recursively flattens a nested JavaScript object. It converts an object with nested properties into a single-level object with dot-separated keys representing the path to each value in the original object. I found it especially useful for partial updates when using Mongoose.

## Function Signature

```typescript
function flattenObject(object: any, prefix: string = ''): { [key: string]: any }
```
### Example input:
```typescript
const input = {
  a: 1,
  b: {
    c: 2,
    d: {
      e: 3
    }
  },
  f: 4
};
```

### Example output:
```typescript
const output = {
  a: 1,
  'b.c': 2,
  'b.d.e': 3,
  f: 4
};
```


david-pilny:
> Just wanted to ask what would be the functionality for array of objects? Because if I understand this correctly now it just keeps the array as is.

Thank you for your reply. I tried it and discovered some unexpected behavior. Anyway, I can modify the code so the function leaves the array of objects as it is or flattens all the objects in the array. The second option makes more sense to me, what do you think?

avivkal:
I think the second option makes more sense

david-pilny:
Just committed the suggested changes.

Issue #5870, opened 2024-05-27
Unexpected result of _.reduce when Object contains key 'length' with number value


rap1ds:
Steps to reproduce. Run the following:

```js
_.reduce({ 'a': 1, 'b': 2, 'length': 5 }, function(result, value, key) {
  result[key] = value;
  return result;
}, {});
```

Expected output:

```js
{a: 1, b: 2, length: 5}
```

Actual output

```js
{0: undefined, 1: undefined, 2: undefined, 3: undefined, 4: undefined}
```

Runkit: https://runkit.com/rap1ds/66543d1a7e77f100088019bc

Root cause:

I believe that the root cause is that [`baseEach` uses `isArrayLike`](https://github.com/lodash/lodash/blob/a67a085cc0612f5b83d78024e507427dab25ca2c/src/.internal/baseEach.ts#L16) which returns `true` for Object with key `length` and a number value.

rap1ds:
Somewhat related issue/PR:

* https://github.com/lodash/lodash/pull/5751
* https://github.com/lodash/lodash/pull/5754

baguse:
Hello @rap1ds, I have a PR for this issue. Maybe you can check it. Thanks

PR #5874, opened 2024-05-30
fix: reduce when we use object with 'length' attribute

baguse:
#5870 

This PR is for fixing the related issue


src/.internal/baseEach.ts
```diff
@@ -13,7 +13,7 @@ function baseEach(collection, iteratee) {
   if (collection == null) {
     return collection
   }
-  if (!isArrayLike(collection)) {
+  if (!isArrayLike(collection) || collection.constructor?.name === 'Object') {
```
rap1ds, line 16: To me, this looks like a breaking change.

The current implementation works well for JS classes like this:

```
class Rectangle {
  constructor(height, width) {
    this.height = height;
    this.width = width;
  }
}

_.reduce(new Rectangle(100, 100), function(result, value, key) {
  result[key] = value;
  return result;
}, {});
```

However, this change is now checking the constructor name, and `collection.constructor?.name === 'Object'` would return `false` to `Rectangle`, so it would probably break there.

Cassieminkus1 (approved):

jdalton:
Closing as stale: this conflicts with the current base and its diff has expanded into the full JS→TS migration in reverse. If the change is still relevant, please open a fresh PR against `main`.

PR #5887, opened 2024-06-18
feat: toPercent()

ianzone:
this function converts a fraction number to a string in percentage format.


UlisesGascon:
I will suggest to include tests, [here](https://github.com/lodash/lodash/blob/main/test/inRange.spec.js) you can find an example from the `inRange()` function ([docs](https://lodash.com/docs/4.17.15#inRange)).

ianzone:
I just added the test.

ianzone:
Hi, is it ok?

test/toPercent.spec.ts
```diff
@@ -0,0 +1,15 @@
+import toPercent from '../src/toPercent';
+
+describe('toPercent', () => {
```
UlisesGascon, line 3: I will suggest to include tests to cover the usage of `minPrecision` and `maxPrecision`.

src/toPercent.ts
```diff
@@ -0,0 +1,27 @@
...
+ * // => 20%
+ *
+ * toPercent(0.3333333)
+ * // => 33.33%
+ *
+ * toPercent(0.2, 2)
+ * // => 20.00%
+ *
+ */
+export default function toPercent(fraction: number, minPrecision: number = 0, maxPrecision: number = 2) {
```
UlisesGascon, line 21: Note  that `minPrecision` and `maxPrecision` are not used, so this example generates an unxpected output:

```
toPercent(12, 4, 7)
// '1,200%'
```

IMO, seems like the output is very deterministic as the locals are not included, ref https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat#using_locales.

Issue #5823, opened 2024-02-29
 Merge mutates source object if nested property is a class instance


DeveloperDanjo:
An original issue #5269 was originally filed about this, but there was no feedback from the issue creator, and it was closed.

My team is experiencing the same situation. However, ours would not be correctly fixed by omitting the original value before merging as was done by a second commenter on that issue.

-- **Original issue text starts (as it is still applicable)** --

If property of the source objects is created from a class, it will be mutated.

```js
import { merge } from "lodash"

class someClass {
  aa = 0
}

const src1 = { pg: { a: new someClass() } }
const src2 = { pg: { a: { aa: 2 } } }

console.log(JSON.stringify(src1)) // {"pg":{"a":{"aa":0}}}
console.log(JSON.stringify(src2)) // {"pg":{"a":{"aa":2}}}

const dest = merge({}, src1, src2)

console.log(JSON.stringify(src1)) // {"pg":{"a":{"aa":2}}} // However, aa should still be 0
console.log(JSON.stringify(src2)) // {"pg":{"a":{"aa":2}}}
```

-- **Original issue text ends** --

Our situation is the same; we are working with an options object for some widget which contains a renderer where the default renderer is an instance of a basic "Renderer" object.

When instantiating a new widget, the author is allowed to provide an object that conforms to the same interface as the "Renderer", but does not have to extend it.

When the new widget is created, the custom `render` function on the Renderer is attached to the default instance, causing future widgets to use that `render` function instead:

```js
class Renderer {
  render() { return 'original'; }
}

const DEFAULT_OPTIONS = {
  /** @type {{ render(): string; }} */
  renderer: new Renderer()
};

class Widget {
  #options = undefined;
  constructor(options) {
    this.#options = merge({}, DEFAULT_OPTIONS, options);
  }
}

// ....

const widget = new Widget({
  renderer: { render() { return 'custom'; } }
});
```

Results in:

`widget.#options.renderer`:
```
{
  render() { return 'custom'; },
  <prototype>: {
    constructor: class Renderer,
    render() { return 'default'; }
  }
}
```

Which is OK, but a little unexpected. I would have expected this to be the same plain object I passed in the options

`DEFAULT_OPTIONS.renderer`:
```
{
  render() { return 'custom'; },
  <prototype>: {
    constructor: class Renderer,
    render() { return 'default'; }
  }
}
```

[...]

Issue #5269, opened 2021-10-04
Merge mutates source object if nested property is a class instance


belspirit:
If property of the source objects is created from a class, it will be mutated. But the documentation says that source objects are not mutable.

```
import { merge } from "lodash"

class someClass {
  aa = 0
}

const src1 = { pg: { a: new someClass() } }
const src2 = { pg: { a: { aa: 2 } } }

console.log(JSON.stringify(src1))
console.log(JSON.stringify(src2))

const dest = merge({}, src1, src2)

console.log(JSON.stringify(src1))
console.log(JSON.stringify(src2))
```

Logs:
```
{"pg":{"a":{"aa":0}}} // src1
{"pg":{"a":{"aa":2}}} // src2
{"pg":{"a":{"aa":2}}} // src1 -> expected aa: 0
{"pg":{"a":{"aa":2}}} // src2
```

halan:
I'm investigating that with @iamnoah.

halan:
I'm investigating that with @iagodahlem

kokushkin:
I've encountered the same behavior with "omit"

`
import _ from "lodash";

class A {
  constructor() {
  this.b = "asdasd";
  this.c = "asdfasdas";
  }
}

const obj = {
  a: new A()
};

const obj1 = _.omit(obj, ["a.c"]);
console.log(obj);

`

![1tl04wt2.png](https://user-images.githubusercontent.com/9853656/169505638-fd852a1a-f693-4a13-854b-43604b274d7b.png)

https://codesandbox.io/s/eager-mendel-kw5jlk?file=/src/index.js:23-215

kokushkin:
My use case was to hide session data before logging it

`
const cleanedConfig = _(config).omit(["httpsAgent._sessionCache"])

`

It was super unclear what's going on cause all what I've seen were these messages.. sometimes..

`
TypeError: Cannot read properties of undefined (reading 'list')
  at Agent._evictSession (node:https:**:**)
  at TLSSocket.<anonymous> (node:https:**:**)
  at Object.onceWrapper (node:events:***:**)
  at TLSSocket.emit (node:events:***:**)
  at node:net:***:**
  at TCP.done (node:_tls_wrap:***:*)

`

So  as _sessionCache was undefined on the original object therefore _sessionCache.list crashed.

(I've intentionally hidden (**) the lines numbers)

PR #5890, opened 2024-06-20
suggested note to _.merge

dan-zinger-bjss:
The purpose of this PR is to add a note to the documentation of _,merge to highlight to users how to avoid sources mutating when merging with multiple.
This is in response to open issue https://github.com/lodash/lodash/issues/5823

I have responded to the issue with a comment. 

I believe it is intended behaviour. This is the second issue raised of its type (first being https://github.com/lodash/lodash/issues/5269).

Since it does seem a likely scenario that users would want to merge multiple sources and not mutate the sources, and also this seems an easy trap to fall into,  I believe it would be beneficial to the community to have this minor addition to the documentation.


jdalton:
Thanks for this. The note itself is reasonable, but the diff replaces the whole repo rather than the single `_.merge` JSDoc change described. If you can resubmit a minimal PR touching only `lodash.js`, I'm happy to take it. Closing this one for now.

PR #5945, opened 2024-11-29
[feature: feature/baseSum-zero] Consider 1 undefined, null, false

zxczoxc125:
# as-is
`_.sum([undefined]); // undefined`
`_.sum(undefined); // 0`
`_.sum([1, undefined]); // 1`

`_.sum([null]); // null`
`_.sum([1, null]); // 1`
`_.sum(null); // 0`

`_.sum([false]); // false`
`_.sum([1, false]); // 1`
`_.sum(false); // 0`

# to-be
`_.sum([undefined]); // 0`

`_.sum([null]); // 0`

`_.sum([false]); // 0`

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -960,8 +960,8 @@
    * @returns {number} Returns the sum.
    */
   function baseSum(array, iteratee) {
-    var result,
-        index = -1,
+    var result = 0;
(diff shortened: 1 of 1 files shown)


jdalton:

PR #5971, opened 2025-03-05
Docs: Fix inconsistent terminology between reject and unescape functions

### LINKED ISSUE #5940: Docs: Inconsistent wording and referencing in help text

2: Example:

vedant713:
Fixes https://github.com/lodash/lodash/issues/5940

This PR addresses the inconsistent terminology in the documentation:

reject was described as "The opposite of _.filter"
unescape was described as "The inverse of _.escape"
I've standardized the terminology to use "opposite" consistently across both functions to improve documentation clarity.

The issue also mentioned cross-referencing, but both functions already correctly reference each other with @see tags.

### DIFF
--- a/dist/lodash.js
+++ b/dist/lodash.js
@@ -15200,7 +15200,7 @@
     }
 
     /**
-     * The inverse of `_.escape`; this method converts the HTML entities
+     * The opposite of `_.escape`; this method converts the HTML entities
      * `&amp;`, `&lt;`, `&gt;`, `&quot;`, and `&#39;` in `string` to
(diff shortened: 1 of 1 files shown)


Cassieminkus1 (approved):

jonchurch:
inverse is correct here, as it implies that it is reversible

The linked issue's main point as I read it is about discoverability. They make a valid point that it would be useful to find the related methods from each other.

The docs do not parse out the `@see` annotations from the JSDoc, so unless the description actually mentions another method, a link will not be created.

Issue #5953, opened 2024-12-29
Feature Request: Extend Lodash Methods to Support Mixed Data Formats (Flat and Nested)


Rangarajanl-OpenSource:
#### **Overview**:
This feature request proposes extending **Lodash’s core capabilities** to handle inputs that include both **nested and flat data** structures. Currently, Lodash methods like `uniqBy` are limited in their ability to work with **nested objects** when processing mixed data formats. By allowing **seamless processing** of both flat and nested data, we aim to make **Lodash** more flexible and efficient for modern data manipulation tasks. This is especially important for use cases such as **data preprocessing for machine learning**, **user profile management**, and other scenarios involving **complex data structures**.

The goal is to standardize the handling of **nested properties** via a common iterator that can easily traverse both flat and nested paths (e.g., `person.name` or `name`). This enhancement will ensure **consistent behavior** across all relevant Lodash methods, enabling users to process **mixed data structures** without needing custom logic.

#### **Current Issue**:
The current **issue #[5951](https://github.com/lodash/lodash/issues/5951)** requests this feature specifically for the `uniqBy` method. However, this functionality would benefit **a wider range of Lodash methods**, ensuring that Lodash can handle complex data scenarios more effectively. These methods include:

1. **`forEach`**: Iterating over mixed data structures with nested and flat properties.
2. **`map`**: Applying transformations while handling both nested and flat data seamlessly.
3. **`filter`**: Filtering collections with both types of properties in a unified manner.
4. **`reduce`**: Aggregating results from mixed data without requiring custom processing logic.

##### **1. `forEach`**: Iterating over mixed data structures with nested and flat properties.

**Before Implementation**:
```javascript
const data = [
  { name: 'Foo' },
  { person: { name: 'Foo' } },
  { name: 'Bar' },
  { person: { name: 'Bar' } }
];

_.forEach(data, (item) => {
  console.log(item.name);  // Will break for nested objects as 'name' is not directly available
});
```

**After Implementation**:
```javascript
const data = [
  { name: 'Foo' },
  { person: { name: 'Foo' } },
  { name: 'Bar' },
  { person: { name: 'Bar' } }
];

_.forEach(data, (item) => {
  const name = getValue(item, 'person.name') || item.name;
  console.log(name);  // Handles both flat and nested properties
});
```

##### **2. `map`**: Applying transformations while handling both nested and flat data seamlessly.

**Before Implementation**:
```javascript
const data = [
  { name: 'Foo', sales: 100 },
[...]

Issue #5951, opened 2024-12-18
`uniqBy` returns incorrect values


vinaysharma14:
```js
import _ from 'lodash';

const result = _.uniqBy(
    [
        { name: 'Foo' },
        { name: 'Foo' },
        { name: 'Bar' },
        { name: 'Bar' },
        { person: { name: 'Foo' } },
        { person: { name: 'Foo' } },
        { person: { name: 'Bar' } },
        { person: { name: 'Bar' } },
    ],
    'person.name',
);

// result
[
  { name: 'Foo' }, // it shouldn't be present
  { person: { name: 'Foo' } },
  { person: { name: 'Bar' } }
]
```

hamza-elmoudden:
If you want to use _.uniqBy to ensure that duplicates are removed based on the person.name property only, you can add a condition to check if the object has a person property before manipulating the data.


import _ from "lodash"

const result = _.uniqBy(
    [
        { name: 'Foo' },
        { name: 'Foo' },
        { name: 'Bar' },
        { name: 'Bar' },
        { person: { name: 'Foo' } },
        { person: { name: 'Foo' } },
        { person: { name: 'Bar' } },
        { person: { name: 'Bar' } },
    ].filter(item => item.person?.name),
    (item) => item.person?.name
);

console.log(result);

Rangarajanl-OpenSource:
@vinaysharma14 I personally see this as a potential core feature that could be extremely useful. If you agree, please consider reviewing and upvoting the issue: [#5953](https://github.com/lodash/lodash/issues/5953).
I’ve made sure to update the ticket diligently with all relevant details. Feel free to add any further insights you think are necessary. If you're open to it, we could also collaborate on implementing this change together!

FMaxey3:
@eugenemaxey [](url)curl -o media3-migration.sh \
   #"https://raw.githubusercontent.com/google/ExoPlayer/r2.19.1/media/3-migration.sh"

PR #5972, opened 2025-03-05
Add uniqByExisting function to handle missing property paths

vedant713:
Adds a new uniqByExisting function that behaves like uniqBy but skips elements where the specified property path doesn't exist. This addresses issue #5951 and partially implements the feature request in #5953 for mixed data formats.


jdalton:

Issue #5931, opened 2024-10-16
Split Based on Multiple Words


MohitVachhani:
For example:

```
const input = "apple-orange banana_grape";
const result = _.split(input, ['orange', 'banana']); 
// Expected result: ['apple-', ' grape']
```

However, this approach doesn’t seem to work. Is there currently any functionality in Lodash that allows splitting a string by multiple words? 

Use Case:
This feature would be valuable when parsing text or documents where specific words serve as delimiters, especially in data extraction and text processing scenarios.

dozsolti:
Hi, this is a good idea, but I don't know if this library will add support for this anytime soon.
So here is a way to do it, for now, with what is out there:

```js
function splitByArray(input, separators, limit){
	
  let result = [input];
  
  for(let sep of separators){
  	result = _.split(result.join(''), sep, limit)
  }
  
  return result;
}
```

frontman-git:
You can pass a regular expression as the separator, like

`
_.split(input, /orange|banana\_/)
// Expected [ 'apple-', ' ', 'grape' ]
`

jdalton:
Thanks for the suggestion. Lodash isn't taking on new API surface right now; keeping the existing methods stable and focused is the priority. I'm closing this as out of scope for the moment. It's a reasonable idea and worth revisiting if that direction changes. Appreciate you writing it up.

PR #5973, opened 2025-03-05
add splitByWords utility

vedant713:
This adds a new string utility that extends _.split to handle multiple word separators passed as an array. It properly splits strings at each separator and handles connecting characters between words.

Fix for issue #5931


jdalton:

Issue #5985, opened 2025-04-25
sumBy returns boolean instead of number when summing booleans


johnmasonyxe:
**description:**

when using `_.sumBy` on an array of objects with boolean values, the result is a boolean when the array contains only one item

**reproduction**

run this code on https://playcode.io/lodash:

```
import _ from 'lodash';

const result = _.sumBy([{ isComplete: true }], 'isComplete');

console.log(_.VERSION);         // 4.17.21
console.log(result);            // expected: 1, actual: true
console.log(typeof result);     // expected: 'number', actual: 'boolean
console.log(result === true);   // expected: false, actual: true
console.log(result === 1);      // expected: true, actual: false
```

**expected behaviour**
lodash should coerce boolean values to numbers (true → 1, false → 0) consistently, even for single-item arrays.

![Image](https://github.com/user-attachments/assets/305a6415-30b0-4fd4-b206-ae44fd6ce169)

adamsol:
I'd also be interested in seeing this fix merged. The current behavior is a potent trap, as code may seem to work for 0 and for 2 elements, but fails for 1 element.

This issue has already been reported multiple times:
* https://github.com/lodash/lodash/issues/5818
* https://github.com/lodash/lodash/issues/4878
* https://github.com/lodash/lodash/issues/3912
* https://github.com/lodash/lodash/issues/3234

FWIW, there's a similar utility function in Python, [`sum`](https://docs.python.org/3/library/functions.html#sum), which works as expected (and it doesn't support strings; here's why summing strings this way is not recommended: https://stackoverflow.com/a/3525380).

```py
>>> sum([True])
1
```

```py
>>> sum(['x'])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

jdalton:
Related to: #5818 — same underlying `_.sumBy`/`_.sum` numeric-coercion behavior.

PR #5986, opened 2025-04-25
fix: sumBy returns boolean instead of number when summing booleans #5985

johnmasonyxe:
**problem**
- `_.sumBy` called with a single boolean element returns `boolean` a instead of `number`

**solution**
- coerce boolean to number for the base case

**notes & links**
- added tests for base case
- [link to issue](https://github.com/lodash/lodash/issues/5985)


jdalton:
Cross-linking #5986 and #6072 — both touch `baseSum`, but they fix different reports (#5985 vs #5818) and disagree on semantics: #5986 coerces booleans to numbers, while #6072 returns `NaN` for any non-number (which would make booleans `NaN`). Leaving both open since which behavior is correct is a maintainer call, not a straight dedupe.

PR #5996, opened 2025-06-05
test: add test for size with string input

frontman-git:
Adds a test to verify that `_.size` correctly returns the length of string values.

This ensures that string inputs like `'hi'` return the expected length (2), aligning with Lodash documentation and behavior for string primitives.


Cassieminkus1 (approved):

ljharb (approved):

jonchurch:
pushed a commit to capture the unicode behavior of `_.size`

for string _.size relies on [`stringSize`](https://github.com/lodash/lodash/blob/b5e672995ae26929d111a6e94589f8d03fb8e578/lodash.js#L1337), which also relies on [`unicodeSize`](https://github.com/lodash/lodash/blob/b5e672995ae26929d111a6e94589f8d03fb8e578/lodash.js#L1387)

jonchurch (approved):

ljharb (changes requested):
i'm pretty sure this is incorrect - `_.size` counts enumerable own string-keyed properties, not code unit pairs or graphemes.

ljharb:
sorry for the back and forth, but all the docs i can find suggest it's wrong. if it's *right* then this PR should fix all the docs also.

PR #5997, opened 2025-06-06
Remove _.template JSDoc examples from lodash.js to prevent false posi…

MikeTheCyberGuy:
…tive source disclosure findings

Removed the JSDoc comment block for the _.template function from lodash.js. The documentation included usage examples such as <% print("hello " + user); %> and <%- value %>, which are detected by security scanners (e.g., Burp Suite) as potential ASP Classic or server-side template source code disclosure, resulting in persistent false positive findings during automated scans.

While this is not an actual vulnerability, the comment block triggers unnecessary alerts and adds overhead to remediation workflows. The removal is limited to documentation comments only and does not affect lodash’s runtime functionality.

This change reduces noise in security reports and simplifies compliance reviews without impacting the library’s intended behavior.


lodash.js
```diff
@@ -14731,110 +14731,6 @@
...
 
-    /**
-     * Creates a compiled template function that can interpolate data properties
-     * in "interpolate" delimiters, HTML-escape interpolated data properties in
-     * "escape" delimiters, and execute JavaScript in "evaluate" delimiters. Data
-     * properties may be accessed as free variables in the template. If a setting
-     * object is given, it takes precedence over `_.templateSettings` values.
-     *
-     * **Note:** In the development build `_.template` utilizes
-     * [sourceURLs](http://www.html5rocks.com/en/tutorials/developertools/sourcemaps/#toc-sourceurl)
```
Gb2real, line 14742: Complete with all the details and information about this event

PR #6009, opened 2025-07-21
docs: document behavior of zip functions with arrays of different lengths

### LINKED ISSUE #6006: Document behavior of functions when lengths of input arrays do not match
`zip` and its variants take multiple arrays and operate across all of them. The documentation should explain what happens when the arrays have differing lengths.

It appears that iteration continues through the longest array and you get `undefined` values when one array runs short:

```
> const _ = require('lodash');
> _.zip(['a', 'b'], [1,2,3]);
[ [ 'a', 1 ], [ 'b', 2 ], [ undefined, 3 ] ]
> _.zip(['a', 'b', 'c'], [1,2]);
[ [ 'a', 1 ], [ 'b', 2 ], [ 'c', undefined ] ]
```

`zipObject`'s behavior is understandably somewhat inconsistent with this. If the `props` argument is the short array, then it appears to stop early:

```
> _.zipObject(['a', 'b'], [1,2,3]);
{ a: 1, b: 2 }
```

However, if its `values` array is short, it again populates with `undefined`:

```
> _.zipObject(['a', 'b', 'c'], [1,2]);
{ a: 1, b: 2, c: undefined }
```

These details should all be clearly documented so that developers face no surprises. Documenting the behavior also reminds developers to consider and handle this potential pitfall in their own code.

I assume this behavior is validated in tests, but that should also be added if not.

If there is some way to modify this behavior (such as forcing it to stop when the shorter array runs out or providing a default value instead of `undefined`), that should also be documented (or linked if it is already documented elsewhere). But I'm not aware of any.

I don't know all the methods that take multiple array arguments and can be affected by differing lengths, but it at least affects `zip`, `zipWith`, `zipObject`, and `zipObjectDeep`.

gingeekrishna:
document behavior of zip functions with arrays of different lengths
- Add documentation for zip(), zipWith(), zipObject(), and zipObjectDeep()
- Clarify that zip/zipWith iterate until longest array is exhausted
- Explain that missing elements are filled with undefined
- Document that zipObject/zipObjectDeep ignore extra values and assign undefined for extra props
- Add examples showing behavior with arrays of different lengths

Fixes #6006

### DIFF
--- a/dist/lodash.js
+++ b/dist/lodash.js
@@ -8687,6 +8687,10 @@
      * first elements of the given arrays, the second of which contains the
      * second elements of the given arrays, and so on.
      *
+     * **Note:** When arrays have different lengths, iteration continues until
+     * the longest array is exhausted. Missing elements from shorter arrays
+     * are filled with `undefined`.
+     *
      * @static
      * @memberOf _
      * @since 0.1.0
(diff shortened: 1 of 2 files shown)


Cassieminkus1 (approved):

gingeekrishna:
@Cassieminkus1 is that any reason my PR is still pending to merge?

Issue #6019, opened 2025-10-11
Backlog: Let's restore the CI


UlisesGascon:
**Hey team!** 👋  

As discussed recently, here’s a coordinated effort to **restore and expand our CI** 🎉  

### ✅ Current PRs  
- [x] Set up a pipeline for Bun (https://github.com/lodash/lodash/pull/6023)  
- [x] Fix linting issues (shared commit on https://github.com/lodash/lodash/pull/6022 and https://github.com/lodash/lodash/pull/6023)  
- [x] Enable Node tests — expand coverage from Node 4 → 25 (https://github.com/lodash/lodash/pull/6022)  
- [x] Add Node 25 to the pipeline (pending release next week — https://github.com/nodejs/node/pull/59896)  
- [x] Add a dedicated pipeline for docs checks (https://github.com/lodash/lodash/pull/6020)  
- [x] Create a browser testing pipeline using Playwright (https://github.com/lodash/lodash/pull/6021)  

### 💬 Open Questions  
- [x] I’m assuming Lodash v4 is compatible starting from Node 4 — please confirm if that’s not the case.  
- [x] I’m not sure what our current browser compatibility matrix looks like. Playwright works well for modern browsers, but if we want to include older ones (e.g. IE), we might need an external provider like BrowserStack (I’ve been experimenting with it in my fork).  
- [x] Should we also run Node tests on Windows/macOS? My view is that Ubuntu coverage should be sufficient, but open to thoughts.  

### 🔖 Important

Let’s use this thread to **discuss general CI topics**, and keep **implementation details within each PR** for better tracking and organization.

jonchurch:
> ### 💬 Open Questions
> * Should we also run Node tests on Windows/macOS? My view is that Ubuntu coverage should be sufficient, but open to thoughts.


Lets test only on ubuntu, cross plat testing is harder to justify value of here than other projects. Especially since I don't think there was ever a time when it was non linux platforms were being tested?

WikiRik:
Apologies for commenting on a closed issue, but what does this mean to the [TravisCI](https://github.com/lodash/lodash/blob/main/.travis.yml) that is also set up? It's using sauce-labs, something that isn't used in the new CI.

Happy to open a separate issue (or opening a PR removing it all) if that's preferable

EDIT: Semi-related to this; lodash still has both a yarn.lock and a package-lock.json file. yarn is only used in the TravisCI so I would propse removing yarn.lock with those changes as well

PR #6022, opened 2025-10-11
Add CI pipeline for Node

UlisesGascon:
### Main Changes
This PR solve linting issues (c07419e) and introduce a new pipeline for running (`npm run validate`) against Node versions from 4 to 24

### Assumptions
- If this PR lands before Node@25 is available I will create a separate PR to include it in the Matrix, otherwise I will push that change once the version is ready.
- The default npm version included in Node@5 has some weird conflicts with the lockfile, so I opted to define an specific npm version in that scenario.
- I only included ubuntu considering that most of Lodash requirements are related to v8 and not Node.js API. So the CI take less to run and also the pipeline is cleaner.  

### Context
- Related https://github.com/lodash/lodash/issues/6019
- The linting commit (c07419e) is also included in #6023 otherwise that PR will fail.

### DIFF
--- a/.github/workflows/ci-node.yml
+++ b/.github/workflows/ci-node.yml
@@ -4,8 +4,6 @@ on:
   push:
     branches: [ main ]
   pull_request:
-    # Run on every PR, regardless of branch
-    branches: [ '*' ]
 
 jobs:
   test:


jonchurch (approved):

.github/workflows/ci-node.yml
```diff
@@ -0,0 +1,61 @@
...
+          '16',
+          '17',
+          '18',
+          '19',
+          '20',
+          '21',
+          '22',
+          '23',
+          '24',
+          '25'
```
jdalton: We can talk about reducing these. Or one to, skip a fews here but for now it is something.

PR #6023, opened 2025-10-11
Add CI pipeline for Bun

UlisesGascon:
### Main Changes

This PR solve linting issues (c07419e) and introduce a new pipeline for running (`bun run validate`) against Bun versions from 1.0.0 to 1.3.0

### Assumptions
- I only included Ubuntu considering that most of Lodash requirements are related to JSC and not Bun API. So the CI take less to run and also the pipeline is cleaner.

### Context
- Related https://github.com/lodash/lodash/issues/6019
- The linting commit (c07419e) is also included in #6022 otherwise that PR will fail.


.github/workflows/ci-bun.yml
```diff
@@ -0,0 +1,36 @@
...
+        uses: actions/checkout@v4
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v1
+        with:
+          bun-version: ${{ matrix.bun-version }}
+
+      # TODO: Remove this step once the lockfile has been updated to v3 in the repository
+      - name: Install npm dependencies (lockfile v3)
+        run: npm i --lockfile-version 3 --frozen-lockfile
```
jonchurch: huh, `--frozen-lockfile` isn't a documented npm cli flag

I know this is from the `bun install` error message but Im wondering if they made a mistake, this is a yarn flag for sure though

for npm we'd have to `npm ci` or even better we can _only_ update the lockfile without installing anything via `--package-lock-only`
```suggestion
        run: npm i --lockfile-version 3 --package-lock-only
```

.github/workflows/ci-bun.yml
```diff
@@ -0,0 +1,36 @@
+name: CI Bun
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    # Run on every PR, regardless of branch
+    branches: [ '*' ]
```
jonchurch:
```suggestion
  pull_request:
```

Redundant wildcard

jonchurch (changes requested):

PR #6026, opened 2025-10-11
security: Include a threat model

UlisesGascon:
### Main Changes

This PR include a new threat model (dadedacb12a74db6b572e95e69c5e442a3137759) and includes a new reference in the `SECURITY.md` (40bd77a63b253d82056ea9ee40a34ab4379c42ed)

Heavily inspired in Webpack ([ref](https://github.com/webpack/security-wg/blob/main/docs/threat-model.md)) and Express ([ref](https://github.com/expressjs/security-wg/blob/main/docs/ThreatModel.md)) threat models. This is an early version with the idea to help us during the security triage process.

**This PR is very open to feedback and suggestions.**

### Context

Related to https://github.com/lodash/lodash/issues/6027


jdalton:
Thank you @UlisesGascon!, we can follow-up with tweaks to taste.

Issue #6027, opened 2025-10-21
Backlog: Adopt Security Best Practices


UlisesGascon:
**Hey team!** 👋  

As discussed recently, here’s a coordinated effort to **adopt security best practices** 🔐 

### ✅ Current PRs  

- [ ] Add a Threat Model: https://github.com/lodash/lodash/pull/6026
- [ ] Include CNA Escalation in the `SECURITY.md`: https://github.com/lodash/lodash/pull/6025

### 💬 Open Questions  

- [ ] In the IRP is included a reference to the Security Triage Team. I will start to work on a proposal to define that team responsibilities and how to articulate it.

### 🔖 Important

Let’s use this thread to **discuss general CI topics**, and keep **implementation details within each PR** for better tracking and organization.

PR #6029, opened 2025-10-21
Add dependabot

UlisesGascon:
### Main Changes

Adds a simple Dependabot configuration that runs monthly for GitHub Actions and npm dependencies, while avoiding semver-major updates for npm.

### Assumptions

All our dependencies are dev only, so probably we can con with a longer time (quarterly?)

### Context

Related to https://github.com/lodash/lodash/issues/6027

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,7 @@ updates:
     schedule:
       interval: monthly
     cooldown:
-      default-days: 7
+      default-days: 14
 
   - package-ecosystem: npm
     directory: /


falsyvalues:
I would like to suggest to increase that to 14 days, we can discuss on whys.

.github/dependabot.yml
```diff
@@ -0,0 +1,21 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly
+    cooldown:
+      default-days: 7
```
UlisesGascon:
```suggestion
      default-days: 14
```

PR #6049, opened 2025-11-16
ci(node): test on versions supported by the library

falsyvalues:


ljharb:
Indeed, if engines.node says >= 4, then we need to keep testing on every node version.

PR #6068, opened 2025-12-17
Bump lodash npm to v4.17.22

jdalton:
We'll need to add the fixes to the minified forms


ljharb (approved):

UlisesGascon (approved):

jonchurch (changes requested):
we need to update the lodash.js file in this branch as well to apply the patch

I can push up a change for it, and run the minification via @bnjmnt4n's forked `lodash-cli`

jonchurch (changes requested):
I pushed a commit that takes care of the `lodash.js` and `lodash.min.js` files

there's a stray semicolon added which is because...

these files are the exact same as what is distributed from `main`. Those files on main have had the patch applied. So I copied them over to here, that is all (and tweaked the version to not clobber JDD's bump commit, bc main still has `4.17.21` listed in those files). that stray semicolon update came w/

**The commit I pushed can be safely rebased to before the Bump commit, I didnt push that bc I wanted yall to see the commit before rewriting history.** Feel free to `git rebase -i b1ec57f3^` and move b1ec57f to the bottom (to put it at tip of HEAD)

jdalton:
@jonchurch This mean it's gtg. Also @falsyvalues 👀 
curious is core.js is okay

falsyvalues (approved):
We didn't touch core methods so it seems to be ok, whats left is mentioned rebase I think.

For the reference, previous bump:  https://github.com/lodash/lodash/commit/c6e281b878b315c7a10d90f9c2af4cdb112d9625

jonchurch (approved):

PR #6072, opened 2025-12-20
Fix sumBy to return NaN for non-numeric values (fixes #5818)

### LINKED ISSUE #5818: Method `sumBy` returns string

VB5475:
## Description
Fixes issue #5818 where `sumBy` was returning a string concatenation instead of `NaN` when encountering non-numeric values.

## Changes
- Modified `baseSum` function to check if values are numbers before addition
- Returns `NaN` immediately when a non-numeric value is encountered
- Added test cases to verify the fix

## Testing
- ✅ Normal numeric sums still work correctly
- ✅ Non-numeric values (strings, null, objects) now return `NaN`
- ✅ Edge cases like `Infinity`, `-Infinity`, and `undefined` are handled correctly

## Example
Before: `_.sumBy([{ a: 1 }, { a: 'f' }, { a: null }], 'a')` → `'1fnull'`
After: `_.sumBy([{ a: 1 }, { a: 'f' }, { a: null }], 'a')` → `NaN`

Fixes #5818

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -967,6 +967,9 @@
     while (++index < length) {
       var current = iteratee(array[index]);
       if (current !== undefined) {
+        if (typeof current != 'number') {
+          return NaN;
+        }
         result = result === undefined ? current : (result + current);
       }
     }
--- a/test/test.js
+++ b/test/test.js
@@ -21529,6 +21529,14 @@
       assert.strictEqual(_.sumBy(arrays, 0), 6);
       assert.strictEqual(_.sumBy(objects, 'a'), 6);
     });
+
+    QUnit.test('should return NaN when non-numeric values are encountered', function(assert) {
+      assert.expect(3);
+
+      assert.ok(isNaN(_.sumBy([{ 'a': 1 }, { 'a': 'f' }, { 'a': null }], 'a')));
+      assert.ok(isNaN(_.sumBy([{ 'a': 1 }, { 'a': 'string' }], 'a')));
+      assert.ok(isNaN(_.sumBy([{ 'a': 1 }, { 'a': {} }], 'a')));
+    });
   }());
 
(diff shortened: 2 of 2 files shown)


lodash.js
```diff
@@ -967,6 +967,9 @@
     while (++index < length) {
       var current = iteratee(array[index]);
       if (current !== undefined) {
+        if (typeof current != 'number') {
+          return NaN;
```
kushmistry2002:
```suggestion
          return NAN;
```

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #6072). The diff contains 33 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

jdalton:
Cross-linking #5986 and #6072 — both touch `baseSum`, but they fix different reports (#5985 vs #5818) and disagree on semantics: #5986 coerces booleans to numbers, while #6072 returns `NaN` for any non-number (which would make booleans `NaN`). Leaving both open since which behavior is correct is a maintainer call, not a straight dedupe.

PR #6076, opened 2026-01-04
Fix getRawTag robustness for non-configurable tags and Proxies

scrrlt:
### **Summary**

This PR hardens `getRawTag` against runtime errors when processing objects with non-configurable `Symbol.toStringTag` properties or throwing accessors (e.g. certain `DataView` implementations or revoked Proxies).

### **Changes**
- Wrapped descriptor access in `try-catch` to handle hostile proxies/objects.
- Switched to `Object.defineProperty` for safer unmasking (shadowing) and restoration.
- Ensures original getters/setters are preserved exactly as they were after the check.

### **Tests**
- Added `test/getRawTag.spec.js` covering these edge cases.
- Verified with `npm test`.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -6,152 +6,154 @@
  * Based on Underscore.js 1.8.3 <http://underscorejs.org/LICENSE>
  * Copyright Jeremy Ashkenas, DocumentCloud and Investigative Reporters & Editors
  */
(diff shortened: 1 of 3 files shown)


scrrlt:
grah stupid formatter

PR #6081, opened 2026-01-11
fix(isPlainObject): return true for objects with nested null prototype

bradvogel:
Objects created via Object.create(Object.create(null)) should be considered plain objects since their prototype chain consists only of null-prototype objects and eventually ends in null, without any custom constructors.

This fix was made in Zod here https://github.com/colinhacks/zod/pull/4574/changes

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -12149,8 +12149,18 @@
         return true;
       }
       var Ctor = hasOwnProperty.call(proto, 'constructor') && proto.constructor;
-      return typeof Ctor == 'function' && Ctor instanceof Ctor &&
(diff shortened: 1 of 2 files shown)


bradvogel:
Closing since this could result in dangerous false positives.

ljharb:
That doesn't make it a plain object - a plain object is one with no prototype methods whatsoever, meaning, it must inherit directly from null. Every single object's prototype chain terminates in `null`.

bradvogel:
Yup - this was a misunderstanding I had. Thanks!

PR #6099, opened 2026-02-01
docs: add security notice for `_.template` in threat model and API docs

UlisesGascon:


falsyvalues:
@UlisesGascon we need to add this notice also into `_.template` docs.

falsyvalues (approved):

ljharb (approved):

PR #6100, opened 2026-02-02
build: regenerate dist

UlisesGascon:
I was getting some weird additions on #6099 and I decided to do a separate PR for them. Seems like we forgot to include these changes on #6062


falsyvalues (approved):
Looks like we missed that during release.

PR #6114, opened 2026-02-22
Modify test command to include environment variable

Jerad551:
Updated test command to include environment variable for MJStudio.db-viewer installation.


ljharb (changes requested):
the diff doesn't match the description, so i assume this is spam?

PR #6122, opened 2026-03-02
Fix clone/cloneDeep for shadowed keys with non-writable inherited props

### LINKED ISSUE #6112: clone() doesn't clone own 'hasOwnProperty' field if Object.prototype is frozen
Example:

```js
const _ = require('lodash');

const orig = {
  foo: "bar",
  hasOwnProperty: "a string"
};

const cloned = _.clone(orig);

```

nthbotast:
## Summary
Fixes cloning of own properties that shadow inherited keys when the inherited property is non-writable (for example when a prototype is frozen).

`baseAssignValue` now uses `defineProperty` not only for `__proto__`, but also when assigning a key inherited from `Object.prototype` and missing as an own property on the target object.

## Tests
- Added a regression test in clone methods suite for objects whose prototype has a frozen `hasOwnProperty` and whose own `hasOwnProperty` should be preserved after cloning.
- Ran: `node test/test.js` (PASS: 6804, FAIL: 0)

Fixes #6112

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -2594,7 +2594,8 @@
      * @param {*} value The value to assign.
      */
     function baseAssignValue(object, key, value) {
-      if (key == '__proto__' && defineProperty) {
+      if (defineProperty && (key == '__proto__' ||
+            (key in objectProto && !hasOwnProperty.call(object, key)))) {
         defineProperty(object, key, {
           'configurable': true,
           'enumerable': true,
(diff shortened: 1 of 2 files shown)


jdalton:
Duplicate of #6132, which I'm keeping as the consolidated fix for cloning own properties that shadow non-writable / frozen `Object.prototype` members (it extends `baseAssignValue` and includes a regression test). Closing this one to keep the fix in a single place. Thanks for the contribution.

PR #6126, opened 2026-03-04
docs(mean/meanBy): document NaN return for empty arrays

nthbotast:
## Summary
- update `_.mean` docs to explicitly mention `NaN` on empty arrays
- update `_.meanBy` docs with the same return-value note
- address the documentation gap discussed in #5901

## Testing
- not run locally (`jscs` is not available in this checkout)

PR #6143, opened 2026-03-14
docs: align opposite wording between filter/reject and escape/unescape

### LINKED ISSUE #5940: Docs: Inconsistent wording and referencing in help text

nthbotast:
## Summary
- describe `_.filter` as the opposite of `_.reject` for bidirectional discoverability
- align `_.unescape` wording with existing docs style by using "opposite" (matching `_.reject` phrasing)

Fixes #5940

## Validation
- `git diff --check`
- targeted doc assertion script confirming both updated phrases are present in `lodash.js`


ljharb (approved):

jdalton:
Closing this in favor of #6126, which makes the same `unescape` "inverse" → "opposite" wording fix and is the one I'm keeping for that change (it also documents the `NaN`-on-empty return for `mean`/`meanBy`). The filter/reject description tweak here is reasonable on its own; if you'd like that part, feel free to reopen or send it as a focused PR. Thanks for the contribution.

PR #6145, opened 2026-03-20
fix(debounce/throttle): treat undefined options as missing

### LINKED ISSUE #5857: When `debounce` or `throttle` options objects have undefined values, those options are unexpectedly set to non-defaults
This issue is adapted from #5495 which was closed with `Issue Bankruptcy` before receiving a reply. This issue is still a problem in the latest versions of `lodash`

## Background
`debounce` and `throttle` take optional `options` objects:

For debounce:

https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L10278-L10282

For throttle:

https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L10879-L10883

Notice that each of these options has sensible defaults.

## Bug
If you set these properties to `undefined` rather than leaving them out of the object, for example:

```js
{
  leading: undefined,
  maxWait: undefined,
  trailing: undefined
}
```

The options dangerously get set to unexpected values.

For debounce:

* ✅ `leading` falls back to `false`
* ❗ `maxWait` falls back to `0`
* ❗ `trailing` falls back to `false`

For throttle:

* ❗ `leading` falls back to `false`
* ❗ `trailing` falls back to `false`

### Cause
The cause of this behavior is that the code is checking for settings using the `in` operator. Instead of `'property' in options`, a much better check would be `options.property !== undefined`.

The buggy code is here:

Debounce:

https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L10320-L10325

Throttle:

https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L10904-L10907

## Expected Behavior
In JavaScript, we nearly always treat missing properties and `undefined` properties as the same thing. Even though this is not _technically_ true because an object can have a property with an ndefined` value, it's very surprising when code doesn't do this.

While explicitly calling these functions with explicitly `undefined` properties is probably very uncommon, it's not uncommon at all to implicitly do so. For example, if a function is wrapping `throttle` to vend its behavior, one can easily see how this could happen and cause very subtle and unexpected bugs.

### Example Use Case
This example is adapted from @k-funk's example in #4973, where a React hook is extending `debounce`:

```js
function useDebounce(fn, delay, { leading, trailing, maxWait }): {
  // unrelated setup code
  const fnRef = useRef(fn)
  useLayoutEffect(() => fnRef.current = fn, [fn])

  return useMemo(
    // ❗ Note the subtle bug here - leading, trailing, and maxWait could all be set incorrectly!
    () => debounce(() => fnRef.current(), delay, { leading, trailing, maxWait }),
    [leading, trailing, maxWait]
  )
)
```

## Related Issues
* [debounce's options.trailing being unexpectedly set #4973](https://github.com/lodash/lodash/issues/4973) is a duplicate of this from 4 years ago.

Potentially related:

* [does _.throttle() work as expected? #4869](https://github.com/lodash/lodash/issues/4869)
* [The func is invoked second time within 1s immediatly when debounce's option trailing is false and leading is true #4782](https://github.com/lodash/lodash/issues/4782)
* [`_.throttle`'s `wait` not guaranteed in special case #4650](https://github.com/lodash/lodash/issues/4650)

KJyang-0114:
## Problem

When `debounce` or `throttle` options are explicitly set to `undefined` (e.g., `{ leading: undefined, maxWait: undefined, trailing: undefined }`), the `'key' in options` check returns `true`, causing options to fall back to unexpected default values instead of the function's actual defaults.

## Root Cause

The `'key' in options` operator returns `true` for properties explicitly set to `undefined`, unlike `key !== undefined`.

**Debounce** (lodash.js:10433, 10435):
- `'maxWait' in options` returns `true` when `options.maxWait = undefined`, triggering `maxWait = 0` instead of keeping default
- `'trailing' in options` returns `true` when `options.trailing = undefined`, falling back to `false` instead of default `true`

**Throttle** (lodash.js:11016, 11017):
- `'leading' in options` returns `true` when `options.leading = undefined`, falling back to `false` instead of default `true`
- `'trailing' in options` returns `true` when `options.trailing = undefined`, falling back to `false` instead of default `true`

## Fix

Replace `'key' in options` checks with `options.key !== undefined`:

```js
// debounce
maxing = options.maxWait !== undefined;
trailing = options.trailing !== undefined ? !!options.trailing : trailing;

// throttle
leading = options.leading !== undefined ? !!options.leading : leading;
trailing = options.trailing !== undefined ? !!options.trailing : trailing;
```

## Testing

Manual verification confirms fix:
```js
const db = _.debounce(fn, 50, { maxWait: undefined, trailing: undefined });
// Now correctly uses defaults instead of falling back to 0/false
```

## Risk & Rollback

Risk: Low — only changes the condition for detecting whether an option was provided; defaults remain unchanged.
Rollback: Revert the 4 line changes in lodash.js.

Fixes #5857

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -10430,9 +10430,9 @@
       wait = toNumber(wait) || 0;
       if (isObject(options)) {
         leading = !!options.leading;
-        maxing = 'maxWait' in options;
+        maxing = options.maxWait !== undefined;
         maxWait = maxing ? nativeMax(toNumber(options.maxWait) || 0, wait) : maxWait;
-        trailing = 'trailing' in options ? !!options.trailing : trailing;
+        trailing = options.trailing !== undefined ? !!options.trailing : trailing;
       }
 
       function invokeFunc(time) {
@@ -11013,8 +11013,8 @@
         throw new TypeError(FUNC_ERROR_TEXT);
       }
       if (isObject(options)) {
-        leading = 'leading' in options ? !!options.leading : leading;
-        trailing = 'trailing' in options ? !!options.trailing : trailing;
+        leading = options.leading !== undefined ? !!options.leading : leading;
+        trailing = options.trailing !== undefined ? !!options.trailing : trailing;
       }
       return debounce(func, wait, {
         'leading': leading,


nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #6145). The diff contains 27 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #6145). The diff contains 27 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

kirahowe:
Has there been any discussion of taking this approach to fixing the issue? We've worked around this bug in a similar way in our own codebase but would be happy to contribute to an upstream fix if a direction is decided :) Thanks for proposing this solution, seems like a great option IMO!

PR #6149, opened 2026-03-22
fix(baseAssignValue): use defineProperty fallback when assignment is blocked by frozen prototype

### LINKED ISSUE #6112: clone() doesn't clone own 'hasOwnProperty' field if Object.prototype is frozen

novavale:
## Summary

Fixes #6112

When `Object.prototype` (or any prototype in the prototype chain) is frozen, assigning a property whose name matches a non-writable property on that frozen prototype fails silently in sloppy mode or throws a `TypeError` in strict mode. This caused `_.clone` and `_.cloneDeep` to silently drop own properties whose names collide with frozen prototype properties (e.g. `hasOwnProperty`).

### Root cause

`baseAssignValue` used a bare assignment for all keys except `__proto__`:

```js
object[key] = value;  // silently drops the value when proto is frozen
```

### Fix

Wrap the assignment in a `try/catch` and verify the own property was actually created via `hasOwnProperty.call()`. If the assignment threw (strict mode) or the own property is still absent (sloppy mode), fall back to `Object.defineProperty` — matching the existing handling already in place for `__proto__`.

```js
// Before
object[key] = value;

// After
try {
  object[key] = value;
  succeeded = hasOwnProperty.call(object, key);
} catch (e) {
  succeeded = false;
}
if (!succeeded && defineProperty) {
  defineProperty(object, key, { configurable: true, enumerable: true, value: value, writable: true });
}
```

### Reproduction (from the issue)

```js
const _ = require('lodash');

const orig = { foo: 'bar', hasOwnProperty: 'a string' };
const cloned = _.clone(orig);

// Before fix:
console.log(cloned.hasOwnProperty); // [Function: hasOwnProperty] ← WRONG
// After fix:
console.log(cloned.hasOwnProperty); // 'a string' ✅
```

### Test

Added a QUnit test using `vm.createContext` to safely isolate the irreversible `Object.freeze(Object.prototype)` call so it does not contaminate the shared test globals.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -2602,7 +2602,27 @@
           'writable': true
         });
       } else {
-        object[key] = value;
+        // In strict mode, assigning to a property whose name shadows a
+        // non-writable property on a frozen prototype (e.g. a frozen
+        // Object.prototype) throws a TypeError.  In sloppy mode the same
+        // assignment silently fails, leaving no own property on `object`.
+        // Fall back to Object.defineProperty so that own properties with
+        // prototype-shadowing names are correctly set in both modes.
+        var succeeded;
+        try {
+          object[key] = value;
+          succeeded = hasOwnProperty.call(object, key);
+        } catch (e) {
+          succeeded = false;
+        }
+        if (!succeeded && defineProperty) {
+          defineProperty(object, key, {
+            'configurable': true,
+            'enumerable': true,
+            'value': value,
+            'writable': true
+          });
+        }
       }
     }
 
--- a/test/test.js
+++ b/test/test.js
@@ -2924,6 +2924,33 @@
         assert.notStrictEqual(actual, object);
       });
 
+      QUnit.test('`_.' + methodName + '` should clone own properties that shadow frozen `Object.prototype` properties', function(assert) {
+        assert.expect(2);
+
+        // Object.freeze(Object.prototype) is irreversible so run in an isolated
+        // vm context to avoid contaminating the global prototype for other tests.
+        var _vm = (function() { try { return require('vm'); } catch(e) { return null; } }());
+        if (!_vm || !Object.freeze) {
+          assert.ok(true, 'test skipped: requires Node.js vm module and Object.freeze');
+          assert.ok(true, 'test skipped: requires Node.js vm module and Object.freeze');
+          return;
+        }
+
+        var context = _vm.createContext({ _: _, JSON: JSON, Object: Object });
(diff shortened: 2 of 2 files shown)


nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))

## Summary
This PR fixes a critical bug in `baseAssignValue` where property assignment could silently fail when `Object.prototype` (or any prototype in the chain) is frozen and the key matches a non-writable property on that frozen prototype. The fix ensures `_.clone` and `_.cloneDeep` correctly clone own properties that shadow frozen prototype properties.

## Architecture
The fix modifies the `baseAssignValue` helper function (lodash.js:2605-2623) to wrap property assignment in a try/catch and verify success via `hasOwnProperty.call()`. If assignment fails (throws in strict mode) or silently does nothing (sloppy mode), it falls back to `Object.defineProperty`. This mirrors the existing `__proto__` handling and fixes issue #6112.

The PR also adds comprehensive tests (test/test.js:2927-2953) that run in an isolated `vm` context to avoid contaminating the global prototype.

## Issues
No issues found. This is a well-crafted fix:

- **lodash.js:2605-2623 — excellent error handling** — The try/catch pattern correctly handles both strict mode (throws TypeError) and sloppy mode (silent failure) cases. Using `hasOwnProperty.call()` to verify the property was actually created is the right approach.

- **lodash.js:2624-2631 — correct fallback** — The `Object.defineProperty` fallback with `configurable: true, enumerable: true, writable: true` matches the semantics of property assignment and can set properties even when the prototype is frozen.

- **test/test.js:2927-2953 — good test isolation** — Using Node.js `vm` module to freeze `Object.prototype` in an isolated context prevents contaminating other tests. This is thoughtful and prevents test flakiness.

- **No regression risk** — The fallback only triggers when the normal assignment fails, so existing behavior is unchanged. The `defineProperty` guard ensures compatibility with environments without that method.

## Cross-file Impact
- Modified: `lodash.js` (baseAssignValue helper)
- Modified: `test/test.js` (added new test for clone/cloneDeep with frozen prototype)
- No public API changes
- No breaking changes
- Fixes #6112

## Assessment
✅ Approve

This is a solid fix for a real bug that could cause data loss in cloning operations. The solution is well-tested, handles both strict and sloppy modes, and maintains backward compatibility.

PR #6160, opened 2026-03-31
chore: regenerate dist/lodash.core from commit b9d70ee35

jonchurch:
## problem

The `dist/lodash.core.js, dist/lodash.core.min.js` files are slightly stale

## the fix

Rebuild the core files from commit b9d70ee35

This PR is the result of running the lodash CLI against b9d70ee35 to rebuild the core files.

It was behind by a single JSDoc comment.

### context
While prepping the new tag on main and the dist releases I realized that the `dist/lodash.core.*` files are not built during the `npm run build` step, so the sync check in #6137 can't enforce hygiene here.

This PR just syncs them based on commit b9d70ee35 like we have done w/ the actual dist files #6154 #6155 #6156 . The dist files we actually ship to the registry are generated and checked in to their own dist branches (`npm, amd, es, npm-packages`), so not these files im updating.

The dist on main is powering the older CDN distribution channels which pull straight from a github repo and commitish. Those are... well few and far between these days! that'd be like 

`<script src=raw.usercontent.github.com/lodash/lodash/4.17.23...` and 
`<script src=cdn.jsdelivr.net/gh/lodash/lodash...`

jsdelivr and almost everything else defaults to serving from the npm registry, so few (but a measureable few!) are actually using the dist files from the repo

the lodash-cli just isn't part of the devDeps here, that process has been done out of band when publishes happen. hence why it isn't run in the `npm run build` script


ljharb (approved):

jonchurch:
I've got a writeup coming about this. I think we just axe the `dist/` files entirely on the main branch. We have too many places we distribute lodash rn

UlisesGascon (approved):

falsyvalues (approved):

PR #6188, opened 2026-04-04
feat: try native `structuredClone` in `cloneDeep`

### LINKED ISSUE #5833: Attempt structuredClone in cloneDeep before doing manual clone

lewxdev:
resolves #5833

functionality is ported from #5855

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -11166,13 +11166,21 @@
     }
 
     /**
-     * This method is like `_.clone` except that it recursively clones `value`.
+     * This method is like `_.clone` except that it recursively clones `value`. Will attempt to use the native
+     * [`structuredClone`](https://developer.mozilla.org/docs/Web/API/structuredClone)
+     * function, if `value` [is supported](https://developer.mozilla.org/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types)
+     * (and the native implementation is available). Otherwise it will fallback to a
+     * custom implementation.
      *
      * @static
      * @memberOf _
      * @since 1.0.0
      * @category Lang
      * @param {*} value The value to recursively clone.
+     * @param {object} [options]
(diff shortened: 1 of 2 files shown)


tj-commits:
+1

tj-commits:
also why was #5855 closed

lewxdev:
> why was #5855 closed

@tj-commits at the time, it seems like the codebase was going through restructuring... if I recall correctly, a lot of PRs got closed due to this

jdalton:
We're enthusiastic about this. structuredClone's behavior is still settling, though, and v4 is solid as-is, so we'd rather not shift cloneDeep's semantics in the 4.x line. Saving it for the next major and closing for now. Thanks for the contribution — we'll revisit when the major opens up.

PR #6194, opened 2026-04-11
Add GitHub Actions workflow for npm package publishing

jazminegraham97:
This workflow runs tests and publishes a Node.js package to GitHub Packages upon release creation.


ljharb:
Thanks, but this sort of thing should only be initiated by maintainers.

PR #6201, opened 2026-04-19
fix: use baseAssignValue to handle __proto__ in invert and invertBy

### LINKED ISSUE #6195: _.invert and _invertBy removes key when _proto_ value
=== _.invert ===
Has __proto__ key? false

=== _.invertBy ===
Has __proto__ key? false

Antoliny0919:
Closes https://github.com/lodash/lodash/issues/6195

Thanky you for reporting this issue @1Jesper1

Thanks to @JetProc and @vlmbuyd for working on this together.

### Changes

Direct property assignment (e.g. `result[value] = key`) does not work correctly when the value is `__proto__`, as it modifies the object prototype instead of setting an own property. Used `baseAssignValue` instead to safely handle this case in `invert` and `invertBy`, and added tests for this behavior.


Antoliny0919:
Hi :) Is there anything I might have missed? Like the contributor guidelines or anything along those lines!
If so, please let me know, I'd really appreciate it 🙌

1Jesper1:
Maybe we can ask @ljharb ?

1Jesper1:
Or @jonchurch ?

PR #6203, opened 2026-04-21
docs: add threat model exclusions for common report patterns

UlisesGascon:
Adds three new out-of-scope sections to the threat model based on recurring report patterns:

- **Stateful/accessor-backed path arguments**: paths with getters/proxies require code execution, not data
- **Inherent JS prototype behavior**: read-only prototype access is standard JS, not a vuln
- **Vulnerability chaining and gadgets**: must be independently exploitable through Lodash alone

### DIFF
--- a/threat-model.md
+++ b/threat-model.md
@@ -81,6 +81,12 @@ JavaScript's prototype chain is a language-level feature. If a report demonstrat
 
 Lodash guards against writing to built-in prototypes (e.g., blocking `__proto__` and `constructor.prototype` keys in `_.set`, `_.merge`, etc.), but it does not and cannot prevent reading inherited properties. Operations like reading `obj.constructor`, traversing `__proto__`, or accessing `Object.prototype` through normal property resolution are expected runtime behavior, not Lodash vulnerabilities.
 
+### Vulnerability Chaining and Gadgets
+
+If a report demonstrates impact only by combining a Lodash behavior with a separate, independent vulnerability in another library or application component, the vulnerability lies in the downstream consumer, not in Lodash. Each reported vulnerability must be independently exploitable through Lodash alone.
+
+If a Lodash function produces a specific object shape that, when later passed to a different library's vulnerable function, leads to code execution or other impact, the root cause is in the downstream library, not Lodash.
+
 ### Vulnerabilities in the JavaScript Runtime or Platform
 
 If a Lodash method triggers a bug in the JavaScript engine (e.g., V8, SpiderMonkey, JavaScriptCore) that leads to memory corruption or incorrect behavior, the vulnerability lies in the engine, not Lodash.
--- a/threat-model.md
+++ b/threat-model.md
@@ -69,6 +69,12 @@ Applications using Lodash are responsible for input validation. Passing attacker
 
 If a developer intentionally merges user input into global objects or fails to isolate data structures, that is a misuse of Lodash’s documented API, not a Lodash defect.
 
+### Stateful or Accessor-Backed Path Arguments
+
+Lodash functions that accept path arguments (e.g., `_.get`, `_.set`, `_.pick`, `_.unset`, `_.omit`) expect paths to be plain data: strings, or arrays of strings and numbers. If a report relies on path arguments that use accessor properties (getters/setters), Proxy objects, or other stateful mechanisms that change value between reads, it is out of scope.
+
+Constructing such objects requires executing JavaScript code (e.g., `Object.defineProperty`), which cannot be expressed in JSON or any serialization format. If an attacker can execute `Object.defineProperty`, they already have arbitrary code execution in the runtime. This falls under the "code that invokes Lodash" trust boundary (see Elements Lodash Trusts, item 3).
+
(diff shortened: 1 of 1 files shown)


tobie (approved):
This is super clear. Thanks for the writeup!

jdalton (approved):

falsyvalues (approved):
LGTM

ljharb (approved):

threat-model.md
```diff
@@ -69,6 +69,24 @@ Applications using Lodash are responsible for input validation. Passing attacker
 
 If a developer intentionally merges user input into global objects or fails to isolate data structures, that is a misuse of Lodash’s documented API, not a Lodash defect.
 
+### Stateful or Accessor-Backed Path Arguments
+
+Lodash functions that accept path arguments (e.g., `_.get`, `_.set`, `_.pick`, `_.unset`, `_.omit`) expect paths to be plain data: strings, or arrays of strings and numbers. If a report relies on path arguments that use accessor properties (getters/setters), Proxy objects, or other stateful mechanisms that change value between reads, it is out of scope.
+
+Constructing such objects requires executing JavaScript code (e.g., `Object.defineProperty`), which cannot be expressed in JSON or any serialization format. If an attacker can execute `Object.defineProperty`, they already have arbitrary code execution in the runtime. This falls under the "code that invokes Lodash" trust boundary (see Elements Lodash Trusts, item 3).
+
```
jonchurch, line 79:
```suggestion

Path arguments to Lodash functions are expected (and documented) to be data, not code. Functions like `_.get`, `_.set`, `_.pick`, `_.unset`, and `_.omit` take paths as strings or arrays of strings and numbers.

Getters, setters, and Proxies are code, not data. They run JavaScript code when a property is read, and they can't be serialized as JSON or sent over the wire. Something has to call `Object.defineProperty` (or equivalent) from inside the process to set them up.

If a report relies on a path element returning different values on different reads (a check/use mismatch), the attack needs code running inside the process, not just attacker-supplied data. That puts it outside Lodash's trust boundary; see "code that invokes Lodash" under [Elements Lodash Trusts](#elements-lodash-trusts).

```
This framing made the point immediately clear on reading for me, "path arguments are meant to be data not code", and called out the class of bug which we are seeing reported under this: "time of check, time of use"

threat-model.md
```diff
@@ -69,6 +69,24 @@ Applications using Lodash are responsible for input validation. Passing attacker
...
+### Inherent JavaScript Prototype Behavior
+
+JavaScript's prototype chain is a language-level feature. If a report demonstrates read-only prototype access (e.g., `_.get(obj, 'constructor.prototype')` returning `Object.prototype`), that describes standard JavaScript semantics, not a security flaw.
+
+Lodash guards against writing to built-in prototypes (e.g., blocking `__proto__` and `constructor.prototype` keys in `_.set`, `_.merge`, etc.), but it does not and cannot prevent reading inherited properties. Operations like reading `obj.constructor`, traversing `__proto__`, or accessing `Object.prototype` through normal property resolution are expected runtime behavior, not Lodash vulnerabilities.
```
jonchurch:
```suggestion
Lodash blocks writes to built-in prototypes. Keys like `__proto__` and `constructor.prototype` are filtered in `_.set`, `_.merge`, and similar functions. It does not and cannot block reads.

Reading inherited properties is how JavaScript works. `obj.constructor`, traversing `__proto__`, or reaching `Object.prototype` through normal property resolution are language semantics, not Lodash behavior.

If a report shows only read-only prototype access (e.g. `_.get(obj, 'constructor.prototype')` returning `Object.prototype`), it's describing JavaScript, not a Lodash vulnerability.
```
leads with the read vs write distinction from the second paragraph

threat-model.md
```diff
@@ -69,6 +69,24 @@ Applications using Lodash are responsible for input validation. Passing attacker
...
+
+### Vulnerability Chaining and Gadgets
+
+If a report demonstrates impact only by combining a Lodash behavior with a separate, independent vulnerability in another library or application component, the vulnerability lies in the downstream consumer, not in Lodash. Each reported vulnerability must be independently exploitable through Lodash alone.
+
+If a Lodash function produces a specific object shape that, when later passed to a different library's vulnerable function, leads to code execution or other impact, the root cause is in the downstream library, not Lodash.
```
jonchurch:
```suggestion
A Lodash vulnerability has to be exploitable through Lodash alone. If a report's impact depends on combining Lodash behavior with a separate bug in another library or application component, the bug is in the downstream code, not Lodash.

Gadget reports fall under this too. If a Lodash function produces an object shape that later causes a vulnerable function in some other library to execute code, the root cause is that library accepting the shape, not Lodash producing it.
```
brings the "main point" to the top, like the other suggested edits, and calls out the specific class w/ "Gadget reports fall under this too".

jonchurch (approved):
It looks good! I took a pass to try to frontload the most relevant information, and call out specific classes of reports that these out of scope issues usually present as.

Feel free to merge w/o these, they are suggestions.

UlisesGascon (commented):
@jonchurch thanks for these great suggestions ❤️

Issue #19, opened 2012-05-29
Attempting to load minified Lo-Dash results in Uncaught ReferenceError


Guiswa:
See this fiddle:

http://jsfiddle.net/XCsuX/

Attempting to load the minified Lo-Dash (https://raw.github.com/bestiejs/lodash/master/lodash.min.js) results in the following error in the console:

"Uncaught ReferenceError: obj is not defined"

Changing the fiddle to load the unminified version (https://raw.github.com/bestiejs/lodash/master/lodash.js) works as expected without error.

This is reproducible every time for me on:

Windows 7 x64
Google Chrome 21.0.1145.0
Mozilla Firefox 12.0

PR #6207, opened 2026-04-28
Cleanup unused code (#19)

Nikhil31102007:
Automated cleanup


package.json
```diff
@@ -47,15 +47,13 @@
     "jscs": "^3.0.7",
     "lodash": "4.17.20",
     "lodash-doc-globals": "^0.1.1",
-    "markdown-doctest": "^0.9.1",
```
ljharb, line 50: this is used in this very file https://github.com/lodash/lodash/blob/3a546c4fe6c1224c601c6b61fde12bbae1e30182/package.json#L27

package.json
```diff
@@ -47,15 +47,13 @@
...
     "optional-dev-dependency": "^2.0.0",
     "platform": "^1.3.3",
     "qunit-extras": "^3.0.0",
-    "qunitjs": "^2.1.0",
```
ljharb, line 54: and this is used in lots of test HTML files https://github.com/lodash/lodash/blob/3a546c4fe6c1224c601c6b61fde12bbae1e30182/test/fp.html#L23

ljharb (changes requested):
none of this is unused.

Nikhil31102007:
Sorry regarding that. I was just testing on our pipeline with this repo
which end point was to make a pr and our analysis part is faulty so
sorry about that.

PR #6213, opened 2026-05-18
test: cover array-wrapped prototype paths

### LINKED ISSUE #6183: Regression test request: prototype-pollution-safe omit/unset paths
Please add regression tests covering array-wrapped path segments in _.omit / _.unset so prototype pollution defenses stay intact.

Context: our dependency audit flagged lodash in a security-sensitive chain and we want a concrete test guard for the historical edge case.

JetProc:
## Summary

- Add regression coverage for array-wrapped `__proto__` and `constructor.prototype` paths in `_.omit`.
- Add a plain-object `constructor.prototype` array-wrapped path regression check for `_.unset`.
- Keep the change test-only; no implementation behavior is changed.

Closes #6183

## Testing

- `npm test`
  - `test:main`: `PASS: 6828  FAIL: 0  TOTAL: 6828`
  - `test:fp`: `PASS: 327  FAIL: 0  TOTAL: 327`
- `node --check test/test.js`
- `git diff --check`

## Notes

- `npm test` ran the documented build/test flow and left no generated file changes.
- `npm run style:test` and `./node_modules/.bin/jscs test/test.js` both hung under Node 22 after a JSCS circular dependency warning, so I terminated those processes.
- `npm ci` completed using the existing `package-lock.json`; npm reported deprecated packages and audit findings from existing old dev dependencies, but no dependency files changed.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -16578,7 +16578,7 @@
     // Prevent regression for
     // https://github.com/lodash/lodash/security/advisories/GHSA-xxjr-mmjv-4gpg
     QUnit.test('Security: _.omit should not allow modifying prototype or constructor properties', function(assert) {
-      assert.expect(3);
+      assert.expect(5);
 
       var testObj1 = {};
       assert.strictEqual(typeof testObj1.toString, 'function', 'Object.toString should work before omit');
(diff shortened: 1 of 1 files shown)


JetProc:
Hi, just checking in on this. This PR adds test-only regression coverage for the array-wrapped `_.omit` / `_.unset` path cases requested in #6183, with the full test suite passing locally. Happy to adjust the coverage if a different test shape would be preferred.

PR #6218, opened 2026-05-29
Preserve __proto__ values in invert outputs

### LINKED ISSUE #6195: _.invert and _invertBy removes key when _proto_ value

subhwastaken:
## Problem
`_.invert` and `_.invertBy` currently drop entries when an input value becomes the output key `__proto__`.

## Root cause
Both functions assign inverted entries with direct property writes. For `__proto__`, that goes through the prototype accessor instead of defining an own enumerable key, so the entry is not preserved as normal output data.

## Fix
Use Lodash's existing `baseAssignValue` helper for newly-created inverted keys. This keeps `__proto__` as a regular own property while preserving the existing prototype-pollution safeguards already used elsewhere in the codebase.

## Testing
- Added regression coverage for `_.invert({ a: '__proto__' })`.
- Added regression coverage for `_.invertBy({ a: '__proto__' })`.
- Ran `git diff --check` successfully.

Fixes #6195

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -13312,7 +13312,7 @@
         value = nativeObjectToString.call(value);
       }
 
-      result[value] = key;
+      baseAssignValue(result, value, key);
     }, constant(identity));
 
     /**
@@ -13350,7 +13350,7 @@
       if (hasOwnProperty.call(result, value)) {
         result[value].push(key);
       } else {
-        result[value] = [key];
+        baseAssignValue(result, value, [key]);
       }
     }, getIteratee);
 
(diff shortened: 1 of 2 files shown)


jdalton:
Duplicate of #6201, which I'm keeping as the consolidated fix for `_.invert` / `_.invertBy` dropping `__proto__` keys (it uses `baseAssignValue` at both assignment sites and includes tests). Closing this one to keep it in one place. Thanks for the contribution.

PR #6223, opened 2026-05-31
fix(omit): prevent mutation of input object when using deep array paths

### LINKED ISSUE #5610: _.omit from lodash/fp mutates the input object when using deep property paths with arrays

```js
import _ from "lodash/fp";

const obj = {
  foo: [{ bar: 1 }]
};

// Input object log
// { bar: 1 }

// {}
```

```js
```

xxiaoxiong:
## Problem
`_.omit` with deep paths that include array indices (e.g. `foo.0.bar`, `foo[0].bar`) mutates the input object when used with `lodash/fp`.

## Root cause
`customOmitClone` defers cloning only for plain objects (`isPlainObject`). Arrays are returned as-is, so `baseClone` skips deep cloning for array-containing objects. When `baseUnset` later modifies the cloned result, it mutates the shared array reference in the original object.

## Changes
- **`customOmitClone`**: return `undefined` for both plain objects and arrays, allowing `baseClone` to deep-clone them properly
- **`test/test.js`**: add regression test for deep array path mutation

Closes #5610

### DIFF
--- a/dist/lodash.js
+++ b/dist/lodash.js
@@ -5687,7 +5687,7 @@
      * @returns {*} Returns the uncloned value or `undefined` to defer cloning to `_.cloneDeep`.
      */
     function customOmitClone(value) {
-      return isPlainObject(value) ? undefined : value;
+      return isPlainObject(value) || isArray(value) ? undefined : value;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Exagone313:
Hello,

This user is a bot that is automatically generating pull requests on a lot of projects.

This could be used to gain access to insecure GitHub Actions setups. By default, approval on pull requests is required only for first-time contributors. If that is the case for this project, you might need to change the setting `Approval for running fork pull request workflows from contributors`.

I recommend closing this PR.

<https://github.com/search?q=author%3Axxiaoxiong+is%3Apr&s=created&type=pullrequests>

Thanks.

PR #6225, opened 2026-06-02
fix(cloneDeep): preserve Temporal immutable types

algojogacor:
This is an independent contribution and has no connection to any hackathon, competition, or sponsored event.

## Summary

`cloneDeep` currently destroys `Temporal.*` instances — the clone loses its prototype chain, internal slots, and `instanceof` identity, resulting in a plain `{}` object with no Temporal methods.

## Root Cause

In `baseClone`, Temporal objects fall through to the generic object path. Since they have no enumerable own properties and are not in the `cloneableTags` registry, they are returned as empty plain objects `{}`. This breaks `isEqual` comparison, dirty-state checks, and any pattern that relies on cloning immutable built-in types.

## Fix

Added an early return in `baseClone` for objects whose `[[Class]]` tag starts with `[object Temporal.`. These types are immutable by design (per the TC39 Temporal proposal) and should be returned by reference — no cloning is needed or possible.

Detection uses `Object.prototype.toString.call(value)` which all Temporal types set to `[object Temporal.PlainDate]`, `[object Temporal.ZonedDateTime]`, etc. This is the same pattern lodash uses for `getTag`.

## Testing

- All Temporal value types return the same reference through `cloneDeep`: PlainDate, PlainTime, PlainDateTime, PlainYearMonth, PlainMonthDay, Instant, ZonedDateTime, Duration
- `isEqual(cloned, original)` returns `true` after cloning
- Normal object cloning behavior is unaffected (only Temporal-tagged objects hit the early return)

### DIFF
--- a/dist/lodash.js
+++ b/dist/lodash.js
@@ -2679,6 +2679,10 @@
       if (!isObject(value)) {
         return value;
       }
+      // Temporal types are immutable - return reference directly.
+      if (Object.prototype.toString.call(value).slice(0, 18) === '[object Temporal.') {
+        return value;
+      }
       var isArr = isArray(value);
       if (isArr) {
         result = initCloneArray(value);
--- a/lodash.js
+++ b/lodash.js
@@ -2679,6 +2679,10 @@
       if (!isObject(value)) {
         return value;
       }
+      // Temporal types are immutable - return reference directly.
+      if (Object.prototype.toString.call(value).slice(0, 18) === '[object Temporal.') {
+        return value;
+      }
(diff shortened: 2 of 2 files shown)


jdalton:
We're enthusiastic about Temporal support in cloneDeep. Temporal is still moving, though, and v4 is solid as-is, so we'd rather not extend cloneDeep's type coverage in the 4.x line. Saving it for the next major and closing for now. Thanks for the thorough writeup — we'll revisit at the major.

PR #6227, opened 2026-06-03
fix: prevent _.invert and _.invertBy from dropping __proto__ keys

### LINKED ISSUE #6195: _.invert and _invertBy removes key when _proto_ value

godfengliang:
Fixes lodash/lodash#6195

When a value in the input object is , both  and  silently dropped the entry. This happened because the result object was created with  which has a prototype chain. Setting  modifies the prototype instead of creating a data property.

**Fix:** Use  for the result accumulator in , so  is treated as a regular property key.

Same fix applies to .

One-line change:  ->  in .

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -5298,7 +5298,7 @@
      */
     function createInverter(setter, toIteratee) {
       return function(object, iteratee) {
-        return baseInverter(object, setter, toIteratee(iteratee), {});
+        return baseInverter(object, setter, toIteratee(iteratee), Object.create(null));
       };
     }
 


Ap-0007 (commented):
Nice! Using `Object.create(null)` instead of `{}` is a classic and robust way to ensure that keys like `__proto__` or `constructor` are not accidentally dropped or misinterpreted as prototype properties during the inversion process. Great patch!

ljharb:
`{ __proto__: null }` is better.

jdalton:
Duplicate of #6201, which I'm keeping as the consolidated fix for `_.invert` / `_.invertBy` dropping `__proto__` keys (it uses `baseAssignValue` at both assignment sites and includes tests). Closing this one to keep it in one place. Thanks for the contribution.

PR #6229, opened 2026-06-04
Fix cloneDeep Map key cloning

### LINKED ISSUE #6228: _.cloneDeep does not deep clone Map keys (only values)
# `_.cloneDeep` does not deep-clone Map keys (only values) - inconsistency with Set handling

## Description

`_.cloneDeep` deep-clones Map **values** but does NOT deep-clone Map **keys**. This creates an inconsistency:

- **Set items** → deep-cloned ✓
- **Map values** → deep-cloned ✓  
- **Map keys** → shared reference ✗

In my understanding i think there is a trade-off here - if keys are cloned, the original key references can't be used for `.get()` on the clone. However:

1. This is **inconsistent** with how lodash handles Sets (Set items are cloned, despite the same identity lookup concern for `.has()`)
2. This deviates from **`structuredClone()`** behavior (which deep-clones Map keys)
3. This means `cloneDeep` doesn't produce a fully independent copy --> mutations to the original's key objects corrupt the clone

At minimum, this behaviour should be documented. Ideally, it should match Set handling for consistency.

## Reproduction

```javascript
const _ = require('lodash');

// Map keys: NOT deep-cloned
const key = { id: 1 };
const map = new Map([[key, 'value']]);
const clone = _.cloneDeep(map);
console.log(key === [...clone.keys()][0]); // true --> shared reference

key.id = 999;
console.log([...clone.keys()][0].id); // 999 --> clone corrupted

// Set items: ARE deep-cloned (inconsistency)
const item = { id: 1 };
const set = new Set([item]);
const setClone = _.cloneDeep(set);
console.log(item === [...setClone][0]); // false --> properly independent
```

## Comparison with `structuredClone`

```javascript
const key = { id: 1 };
const map = new Map([[key, 'value']]);
const clone = structuredClone(map);

key.id = 999;
console.log([...clone.keys()][0].id); // 1 --> correctly independent
```

## Root cause

`lodash.js` line ~2723 in `baseClone`:

```javascript
// Set items ARE cloned:
value.forEach(function(subValue) {
  result.add(baseClone(subValue, bitmask, customizer, subValue, value, stack));
});

// Map keys are NOT cloned (inconsistency):
value.forEach(function(subValue, key) {
  result.set(key, baseClone(subValue, bitmask, customizer, key, value, stack));
});
```

## Suggested resolution

Either:
1. **Clone Map keys** (matching Set behavior and `structuredClone`):
```javascript
result.set(
  baseClone(key, bitmask, customizer, key, value, stack),
  baseClone(subValue, bitmask, customizer, key, value, stack)
);
```
2. Or **document** that Map keys are intentionally not cloned, and explain why.

## Environment

- lodash 4.18.1
- Node.js v24.1.0

puneetdixit200:
Fixes #6228.

## Summary
- clone Map keys through the same deep-clone path used for Set entries and Map values
- keep the existing stack so shared and circular references are preserved while cloning Map entries
- add regression coverage for object Map keys with nested objects

## Checks
- direct Node regression: current `HEAD` reused the Map key/nested object (`true true`), patched build clones both and preserves the old nested value (`false false 1`)
- `npm run style:main`
- `npm run test:main` (6,828 passing, 0 failing)
- `git diff --check`

## Notes
- `npm run style:test` and direct `jscs test/test.js` were attempted, but the legacy JSCS process did not complete locally under Node v24.15.0.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -2720,7 +2720,10 @@
         });
       } else if (isMap(value)) {
         value.forEach(function(subValue, key) {
-          result.set(key, baseClone(subValue, bitmask, customizer, key, value, stack));
+          result.set(
+            baseClone(key, bitmask, customizer, key, value, stack),
+            baseClone(subValue, bitmask, customizer, key, value, stack)
+          );
         });
       }
 
--- a/test/test.js
+++ b/test/test.js
@@ -2772,6 +2772,31 @@
       assert.notStrictEqual(actual, cyclical['v' + (LARGE_ARRAY_SIZE - 1)]);
     });
 
+    QUnit.test('`_.cloneDeep` should deep clone map keys', function(assert) {
+      assert.expect(3);
+
+      if (Map) {
+        var actual,
+            actualKey,
+            key = { 'a': { 'b': 1 } },
+            map = new Map;
+
+        map.set(key, 'value');
+
+        actual = _.cloneDeep(map);
+        actualKey = actual.keys().next().value;
+
+        assert.notStrictEqual(actualKey, key);
+        assert.notStrictEqual(actualKey.a, key.a);
+
+        key.a.b = 2;
+        assert.strictEqual(actualKey.a.b, 1);
+      }
+      else {
+        skipAssert(assert, 3);
+      }
+    });
+
     QUnit.test('`_.cloneDeepWith` should provide `stack` to `customizer`', function(assert) {
       assert.expect(1);
 


ljharb (changes requested):
this is a breaking change; we can't do it by default.

Ap-0007 (commented):
Great fix! Cloning the keys in a Map is essential for a true deep clone, especially since Map keys are often complex objects themselves. The test cases clearly demonstrate the issue and the fix. Looks solid to me!

PR #6232, opened 2026-06-10
fix: avoid mutating arrays in omit deep paths

### LINKED ISSUE #5610: _.omit from lodash/fp mutates the input object when using deep property paths with arrays

pupuking723:
Fixes #5610.

`_.omit` deep-clones plain objects before unsetting omitted paths, but arrays were returned from `customOmitClone` as-is. For deep paths that traverse arrays, `baseUnset` then operated on an array reference shared with the original input object.

This lets `baseClone` clone arrays as well as plain objects for `omit`, so deep array paths can be removed from the result without mutating the source object. The change covers both `_.omit` and `lodash/fp` `omit`.

Verification:

- `node test/test` -> 6827 pass, 0 fail
- `node test/test-fp` -> 329 pass, 0 fail
- `npm test` -> 6827 main assertions + 329 fp assertions pass
- `npm run style`
- `git diff --check`


jdalton:
Duplicate of #6206, which I'm keeping as the consolidated fix for `_.omit` mutating input arrays on deep paths (the approved minimal `customOmitClone` change). Closing this one to keep it in one place. Thanks for the contribution.

PR #6240, opened 2026-07-01
chore: add homepage and bugs URLs to package.json

DakshSinghDhami:
Adds homepage and bugs.url fields


jonchurch:
Please stop opening these low value PRs across the ecosystem, you will get banned.

This branch's package.json is never even published anywhere, so your changes here do nothing.

Please take more care in your work and at least verify what your diff does.

PR #6243, opened 2026-07-01
Fix _.invert and _.invertBy dropping `__proto__` keys

### LINKED ISSUE #6195: _.invert and _invertBy removes key when _proto_ value

highoncomputers:
This fixes #6195.

## Problem

When the value being inverted is `__proto__`, the property assignment `result[value] = key` in the `createInverter` function would set the prototype of the result object instead of creating a regular property, causing the key to be silently dropped.

## Reproduction

```js
_.invert({userId: 'alice', role: '__proto__', team: 'eng'});
// Expected: {alice: 'userId', __proto__: 'role', eng: 'team'}
// Actual:   {alice: 'userId', eng: 'team'}
```

## Fix

Changed the accumulator in `createInverter` from `{}` to `Object.create(null)`. Since null-prototype objects have no `__proto__` setter, `__proto__` is treated as a regular property name.

This is consistent with how other lodash methods (like `_.assign`, `_.merge`, etc.) handle `__proto__`.

### DIFF
--- a/dist/lodash.js
+++ b/dist/lodash.js
@@ -5298,7 +5298,7 @@
      */
     function createInverter(setter, toIteratee) {
       return function(object, iteratee) {
-        return baseInverter(object, setter, toIteratee(iteratee), {});
+        return baseInverter(object, setter, toIteratee(iteratee), Object.create(null));
       };
     }
 
--- a/lodash.js
+++ b/lodash.js
@@ -5298,7 +5298,7 @@
      */
     function createInverter(setter, toIteratee) {
       return function(object, iteratee) {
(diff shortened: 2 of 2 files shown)


jdalton:
Heads up — this looks like a possible duplicate of #6201, which fixes the same `__proto__`-key drop in `_.invert`/`_.invertBy` by routing assignment through `baseAssignValue`. Closing to consolidate on #6201. If you think that's wrong, please reopen.

PR #6245, opened 2026-07-01
fix(omit): clone arrays on deep paths so input is not mutated

### LINKED ISSUE #5610: _.omit from lodash/fp mutates the input object when using deep property paths with arrays

jdalton:
`_.omit` (and `lodash/fp`'s `omit`) mutated the input object when a path traversed an array:

```js
const obj = { foo: [{ bar: 1 }] };
_.omit(obj, ['foo[0].bar']);
obj.foo[0]; // {} — the source was mutated
```

`customOmitClone` deferred to `cloneDeep` for plain objects but returned every other value — including arrays — by reference. `baseUnset` then deleted the key from the original array's element. Cloning arrays the same way plain objects are already cloned fixes both `_.omit` and the fp `omit`.

Added deep-array-path cases (`a[0].b`, `a.0.b`) to the existing "should not mutate `object`" test.

Fixes #5610.


Ziiyodullayevv (commented):
Important fix for input mutation. Cloning arrays on deep paths in omit prevents unintended side effects on the original input.

PR #6247, opened 2026-07-01
fix(clone): clone BigInt64Array, BigUint64Array, and Float16Array instead of returning {}

spokodev:
`_.clone` and `_.cloneDeep` silently return an empty object `{}` for `BigInt64Array`, `BigUint64Array`, and `Float16Array`, discarding the values, length, and type. That is a data-loss bug: the result is not a copy of the input, it is an empty plain object with no indication anything went wrong.

## Repro

```js
const _ = require('lodash');

_.cloneDeep(new BigInt64Array([1n, 2n, 3n]));   // {}   data lost
_.cloneDeep(new BigUint64Array([1n, 2n]));      // {}   data lost
_.cloneDeep(new Float16Array([1.5, 2.5]));      // {}   data lost

// contrast: the other typed arrays clone correctly
_.cloneDeep(new Uint8Array([1, 2, 3]));         // Uint8Array [1, 2, 3]

// native oracle for comparison
structuredClone(new BigInt64Array([1n, 2n, 3n])); // BigInt64Array [1n, 2n, 3n]
```

## Root cause

The tags `[object BigInt64Array]`, `[object BigUint64Array]`, and `[object Float16Array]` are missing from three tag maps: `typedArrayTags`, `cloneableTags`, and the `initCloneByTag` switch. With the tag absent from `cloneableTags`, `baseClone` hits `return object ? value : {}` and yields `{}` instead of reaching the typed-array case.

lodash is internally inconsistent about these three types: `_.isTypedArray` and `_.isEqual` already recognize them (via `util.types`), only clone drops them.

## Fix

Register the three tags in `typedArrayTags` and `cloneableTags`, and add the three cases to the existing typed-array branch of `initCloneByTag`, which routes them to `cloneTypedArray`. `cloneTypedArray` clones through the value's own constructor, so it handles the new types with no further change.

## Tests

Added `BigInt64Array`, `BigUint64Array`, and `Float16Array` (each guarded by a `typeof` check for engine support) to the shared `typedArrays` list, so they run through the existing clone, `cloneDeep`, `isEqual`, and `isTypedArray` suites. Added a focused assertion that `_.cloneDeep(new BigInt64Array([1n, 2n]))` is a `BigInt64Array` with equal values and a separate buffer. The `should merge typed arrays` test now derives element lengths dynamically and excludes the BigInt-backed types (merging bigint elements with a plain-number array mixes element types, unrelated to clone).

Full suite: 6899 passing, 0 failing (6831 to 6899 from the widened typed-array coverage), zero regressions.

### DIFF
--- a/lodash.js
+++ b/lodash.js
@@ -115,7 +115,10 @@
       weakSetTag = '[object WeakSet]';
 
   var arrayBufferTag = '[object ArrayBuffer]',
+      bigInt64Tag = '[object BigInt64Array]',
+      bigUint64Tag = '[object BigUint64Array]',
       dataViewTag = '[object DataView]',
+      float16Tag = '[object Float16Array]',
       float32Tag = '[object Float32Array]',
       float64Tag = '[object Float64Array]',
       int8Tag = '[object Int8Array]',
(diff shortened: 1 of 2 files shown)


ljharb (approved):

jdalton (approved):

jdalton:
@spokodev unit tests failing

Ziiyodullayevv (commented):
Good fix for typed array cloning. Properly cloning BigInt64Array, BigUint64Array, and Float16Array instead of returning empty objects ensures data integrity.

spokodev:
Fixed in 0b8411a9. The test used a bigint literal (`1n`), which the jscs style check cannot parse, so the style step failed. I swapped it for `BigInt(1)`. `npm test` (main and fp) and `npm run style` pass locally now: main 6899, fp 327, jscs clean.

spokodev:
Nudge on this one. The check you flagged, @jdalton, was a `1n` literal that the jscs style step cannot parse; it was swapped for `BigInt(1)` in 0b8411a9 on 3 July. The branch is still clean against master and there are no open threads.

Both approvals landed just before that fix, so if either of you would rather re-look before merging, say the word — happy to rebase or split it.

PR #6253, opened 2026-07-10
fix: block prototype pollution in set, setWith, update, updateWith, zipObjectDeep

dlavrenuek:
## Description
Fixes prototype pollution in `set`, `setWith`, `update`, `updateWith`, `zipObjectDeep`.

The root cause is path traversal logic in `setBase` which can overwrite built in methods and creates prototyped objects for nested path.

## Alternatives

I did consider changing `{}` to `Object.create(null)` in https://github.com/lodash/lodash/blob/main/lodash.js#L4050 but this would introduce breaking behavior if you rely on f.e. `toString()` being called on the sub-objects created with `_.set(...)`

## Reproduction steps

```
const _ = require('lodash');

// Vector 1 - base has a normal prototype (the common case)
_.set({}, 'toString.call', 'ok');
Object.prototype.toString.call([]); // see error

// Vector 2 - pollution through an auto-created intermediate node
_.set(Object.create(null), 'a.toString.call', 'ok');
// Object.create(null) creates a prototypeless object, but `a` is created as `{}` (object with global object prototype)
// same impact as Vector 1

// Same via the other entry points
_.update({}, 'valueOf.call', () => 'ok');
_.zipObjectDeep(['toString.call'], ['ok']);
```


codeCraft-Ritik (commented):
Nice security fix! Addressing the prototype pollution issue within `setBase `ensures all affected APIs (`set`, `setWith`, `update`, `updateWith`, and `zipObjectDeep`) benefit from the same protection without introducing duplicated logic.👍

UlisesGascon:
As discussed in the advisory ([ref](https://github.com/lodash/lodash/security/advisories/GHSA-85gm-5r6j-jwr9#advisory-comment-290278)) I am closing the PR

PR #6254, opened 2026-07-11
block nested __proto__ traversal in baseUnset

Ramya-9353:
Repro: `_.unset(JSON.parse('{"__proto__":{},"x":{}}'), 'x.__proto__.toString')` deletes `Object.prototype.toString`; afterwards `({}) + ""` throws "Cannot convert object to primitive value".

Cause: the `__proto__` guard consults `hasOwnProperty.call(object, "__proto__")` on the root object only. `JSON.parse` materialises `__proto__` as an own data property, so an attacker-supplied root with an own `__proto__` disables the guard for every segment, and a nested `x.__proto__` then resolves to the real prototype where `delete` removes a built-in method.

Fix: block `__proto__` as a non-terminal traversal key regardless of the root, mirroring the constructor/prototype rule added for GHSA-f23m-r3pf-42rh. A terminal own `__proto__` stays deletable, so removing the JSON-parsed own property still works.


Ramya-9353:
any update?

PR #6255, opened 2026-07-14
prevent defaultsDeep from mutating inherited built-in methods

Ramya-9353:
Repro: after `_.defaultsDeep({}, JSON.parse('{"toString":{"x":1}}'))`, `Object.prototype.toString.x === 1`, so a shared built-in method now carries an own property taken from the source. `valueOf`, `hasOwnProperty` and, recursively, `Function.prototype` methods are reachable the same way, and the write is global and persistent for the process.

Cause: `customDefaultsMerge` reads the destination value with `safeGet`, which returns the value `object` inherits for a key it does not own (`Object.prototype.toString` for a bare `{}`). When that inherited value and the source value are both objects, the customizer calls `baseMerge` on it in place, writing source data onto the shared object. `_.merge` is unaffected because `baseMergeDeep` clones function-valued destinations before merging; the customizer intercepts before that guard runs.

Fix: merge in place only when `key` is an own property of `object`, otherwise return `undefined` so the default merge clones `srcValue` onto `object`. The missing default is still filled as an own property, and `constructor`/`__proto__` stay blocked by `safeGet`. Regenerated `dist/` with `npm run build` and added a regression test under the `lodash.defaultsDeep` module.


Ramya-9353:
any update?

PR #6264, opened 2026-07-26
fix: check __proto__ ownership on the current node in baseUnset

### LINKED ISSUE #6263: baseUnset: __proto__ guard checks ownership on the root object, not the current node
Hi!

Whilst trying to backport the patches for some security issues in Debian, we discovered an interesting problem.

The prototype-pollution guard added to `baseUnset` in 4.18.0 (commit fe8d32e) can be bypassed for `__proto__` when the root object owns a `__proto__` key.

The guard walks the path but never advances `object`:

```
    while (++index < length) {
      var key = toKey(path[index]);
      if (key === '__proto__' && !hasOwnProperty.call(object, '__proto__')) {
        return false;
      }
      ...
    }
```

`object` is the original root the whole time, so `hasOwnProperty.call(object, '__proto__')` only ever inspects the root — not the node at the current position in the path. If the root owns a `__proto__` key, the `__proto__` check is effectively disabled for every segment, and a deeper `__proto__` reaches a nested object's real prototype.

An own `__proto__` key is easy to get from untrusted input: `JSON.parse('{"__proto__":1}')` creates a genuine own data property (it doesn't trigger the setter). Parsing attacker JSON and then running `_.unset` / `_.omit` on it with a path is a normal pattern.

lazerg:
`baseUnset` ran its `hasOwnProperty` check for `__proto__` against the root object instead of the node it had actually reached in the path. If the root owns a `__proto__` data property, which untrusted input can supply through `JSON.parse('{"__proto__":1}')`, the check passes at every segment. A deeper `__proto__` then reaches a nested object's real prototype and can delete inherited methods. The fix walks the current node along with the path and checks ownership on that node.

Fixes #6263


utkarsh2102 (approved):
That looks like the correct behavior. Thank you. 💛 

If you'd like to make it perfect - the test only covers the blocked direction. It'd be worth also asserting that a legitimate own-__proto__ deletion still returns `true` (the `JSON.parse('{"__proto__":{"x":1}}')` -> `_.unset(obj,'__proto__')` case), so a future refactor doesn't "fix" the bypass by over-blocking and silently regress that behavior. That'll be a pretty stronger test also.

lazerg:
I added that case to the same test. 

An object that owns its own `__proto__` (from `JSON.parse('{"__proto__":{"x":1}}')`) still returns `true` from `_.unset(obj, '__proto__')`, and I check the property is gone afterward too. 

So if a future change over-blocks and regresses that, the test catches it now.

utkarsh2102:
@ljharb - hi! any chance we can get a review and merge of this, please?

lazerg:
It's been sitting for a few weeks without a maintainer response, so closing for now. Reopen or merge anytime it works for you.

lazerg:
@ljharb - hi! any chance we can get a review and merge of this, please?

dist/lodash.min.js
```diff
@@ -45,8 +45,8 @@ var u=t[r];if(r==e||u!==i){var i=u;Ui(u)?Cl.call(n,u,1):du(n,u)}}return n}functi
...
 var o=n[r],f=t?t(o):o;if(!r||!Hf(f,c)){var c=f;i[u++]=0===o?0:o}}return i}function vu(n){return"number"==typeof n?n:wc(n)?Un:+n}function gu(n){if("string"==typeof n)return n;if(wh(n))return c(n,gu)+"";if(wc(n))return gs?gs.call(n):"";var t=n+"";return"0"==t&&1/n==-Wn?"-0":t}function yu(n,t,r){var e=-1,u=o,i=n.length,c=!0,a=[],l=a;if(r)c=!1,u=f;else if(i>=tn){var s=t?null:Is(n);if(s)return P(s);c=!1,u=S,l=new dr}else l=t?[]:a;n:for(;++e<i;){var h=n[e],p=t?t(h):h;if(h=r||0!==h?h:0,c&&p===p){for(var _=l.length;_--;)if(l[_]===p)continue n;
-t&&l.push(p),a.push(h)}else u(l,p,r)||(l!==a&&l.push(p),a.push(h))}return a}function du(n,t){t=Iu(t,n);var r=-1,e=t.length;if(!e)return!0;for(;++r<e;){var u=to(t[r]);if("__proto__"===u&&!wl.call(n,"__proto__"))return!1;if(("constructor"===u||"prototype"===u)&&r<e-1)return!1}var i=Hi(n,t);return null==i||delete i[to(Ao(t))]}function bu(n,t,r,e){return cu(n,t,r(ve(n,t)),e)}function wu(n,t,r,e){for(var u=n.length,i=e?u:-1;(e?i--:++i<u)&&t(n[i],i,n););return r?lu(n,e?0:i,e?i+1:u):lu(n,e?i+1:0,e?u:i)}
```
ljharb, line 48: only releases should update this file, not PRs. please revert.

dist/lodash.js
```diff
@@ -4380,7 +4380,8 @@
       // https://github.com/lodash/lodash/security/advisories/GHSA-xxjr-mmjv-4gpg
       // https://github.com/lodash/lodash/security/advisories/GHSA-f23m-r3pf-42rh
       var index = -1,
-          length = path.length;
+          length = path.length,
+          current = object;
```
ljharb: also this file.

lodash.js
```diff
@@ -4399,6 +4400,8 @@
         if ((key === 'constructor' || key === 'prototype') && index < length - 1) {
           return false;
         }
+
+        current = current == null ? current : current[key];
```
ljharb, line 4404: is it intentional that this allows `document.all`?

ljharb (changes requested):

dist/lodash.min.js
lazerg: Good catch, reverted. dist/lodash.min.js is back to the upstream/main state in f72f3af7.

dist/lodash.js
lazerg: Reverted the same way in f72f3af7. dist/lodash.js now matches upstream/main again.

lodash.js
lazerg: Yes, and it lines up with how the rest of the file already treats document.all. This line uses lodash's normal `== null` check. baseGet, hasPath, and isObject all use the same check for path traversal and type detection. document.all is loosely equal to null by spec, so all of those functions already treat it as nullish. This line does not add a special case; it follows the same convention the codebase already uses.

Here that only affects the __proto__ check. If current is document.all, `current == null` is true, so the check short-circuits and returns false, blocking the unset. The code never calls hasOwnProperty on document.all, and it never indexes into it either, since current stays put instead of advancing. So this is more restrictive than the old root-only check, not less. I do not see a path where it lets something through that the old code would have blocked. If you have a specific bypass in mind, let me know and I will test it directly.

lazerg:
Closing this since it has been inactive for a while.

PR #6270, opened 2026-08-10
docs: clarify default step of -1 when end < start in range

### LINKED ISSUE #6057: _.range(10, 0) behavior differs from documentation when step not specified
Consider the call:
```
_.range(10, 0)
```

In lodash `4.17.21` this generates the array:
```
> _.range(10, 0)
[
  10, 9, 8, 7, 6,
   5, 4, 3, 2, 1
]
```

The docs on mouseover in my IDE (so presumably from the typescript typings) specify:

> Creates an array of numbers (positive and/or negative) progressing from start up to, but not including, end. If end is not specified it’s set to start with start then set to 0. **If end is less than start a zero-length range is created unless a negative step is specified.**

The end is less than the start, but I am not specifying a step, so I would expect a zero-length array.

The docs (https://lodash.com/docs/4.17.15) specify:

> Creates an array of numbers (positive and/or negative) progressing from start up to, but not including, end. A step of -1 is used if a negative start is specified without an end or step. If end is not specified, it's set to start with start then set to 0.

The start value is not negative, the end value is not negative and is specified, so I'd expect the step to still be positive and the function to return an empty array.

Probably this is already "expected behaviour" in a large number of peoples code bases, so probably the docs should be updated rather than change what happens.

ChenfromChina123:
## What does this PR do?

Clarifies the documentation of `_.range` regarding the default step when `end` is less than `start`.

Closes #6057

## Changes

- `lodash.js`: the `range` JSDoc now states that a step of `-1` is used when `end` is less than `start` and no `step` is specified, and `1` otherwise.

## Verification

- `node --check lodash.js` passes.
- Runtime: `_.range(10, 0)` → `[10, 9, 8, 7, 6, 5, 4, 3, 2, 1]` matches the documented behavior.


lbesecker195 (approved):
Verified against `baseRange` at `lodash.js:5435`: `step = step === undefined ? (start < end ? 1 : -1) : toFinite(step)`. The new wording matches the actual behavior and resolves the confusion in #6057. LGTM.

PR #6276, opened 2026-08-11
fix: advance lastIndex past empty separator matches in truncate

### LINKED ISSUE #6274: [Bug]`_.truncate` hangs forever with a zero-length-capable regex `separator`
## Description

`_.truncate(string, { separator })` enters an infinite loop when `separator` is a regex that can match the empty string at a non-zero offset (e.g. `/(?<=,)\s*/`). The inner `exec` loop never advances `separator.lastIndex`, so a zero-length match makes `exec` return the same match forever. One CPU core pins at 100% and the call never returns.

This is not a security issue (already triaged as out of scope: the separator is application-chosen, and every documented separator is safe). It is a correctness bug in the `isRegExp(separator)` branch.

## Steps to reproduce

```js
const _ = require('lodash');
_.truncate('a,b' + 'X'.repeat(200), { length: 30, separator: /(?<=,)\s*/ });
// never returns
```

## Expected

Returns a truncated string in bounded time, as it does for string and ordinary regex separators.

## Actual

Infinite loop; the process hangs at 100% CPU.

## Root cause

`lodash.js`, `truncate`, the `isRegExp(separator)` branch:

```js
separator.lastIndex = 0;
while ((match = separator.exec(substring))) {
  var newEnd = match.index;   // lastIndex is never advanced
}
```

For a zero-length match, `match.index === separator.lastIndex` and `lastIndex` stays fixed, so `exec` returns the identical match on every iteration.

## Recommended fix

Advance `lastIndex` past zero-length matches inside the loop:

```js
separator.lastIndex = 0;
while ((match = separator.exec(substring))) {
  var newEnd = match.index;
  if (match[0].length === 0) {
    separator.lastIndex++;
  }
}
```

## Environment

- lodash 4.18.1 
- Node.js, any recent version

lazerg:
`_.truncate` never returns when `separator` is a regexp that can match an empty string, such as `/(?<=,)\s*/`. The loop that scans for the last separator match reads `match.index` but leaves `lastIndex` where it is, so a zero-length match makes `exec` hand back the same match forever and the call pins a CPU core.

Bumping `lastIndex` past an empty match lets the scan move forward and finish. Added a `truncate` test for both a plain and a global regexp of that shape.

Fixes #6274


lazerg:
Hey @jdalton, could you check this fix for lastIndex not advancing past empty separator matches in truncate? No feedback yet after a month.

